Issue #4, opened 2019-11-19
Deprecated apiVersion for Deployment and Ingress object


krysennn:
Unable to install victoria metrics with helm on kubernetes cluster v1.16.1.
```
$ kubectl version
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T16:51:36Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```
 
According to https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

tenmozes:
the new version of helm chart is released

PR #5, opened 2019-11-19
fix: apiVersion of Deployment and Ingress object

krysennn:
https://github.com/VictoriaMetrics/helm-charts/issues/4


charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if .Values.vminsert.enabled -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
```
tenmozes: don't need to add `if` please change `apiVersion` and bump version of chart in `Chart.yml`

krysennn: you are right, i changed
it's ok for you ?

tenmozes: thank you so much

Issue #6, opened 2019-11-26
Can't install chart version 0.3.0


ThePants999:
helm install my-release vm/victoria-metrics-cluster --dry-run
Error: YAML parse error on victoria-metrics-cluster/templates/clusterrole.yaml: error converting YAML to JSON: yaml: line 7: did not find expected key

This was with Helm 3, but I do get the same error with Helm 2. (Not during dry run - Helm 2 happily does a dry run but gives the error above when you attempt a real install. Helm 3 gives the error either way.)

kostyaplis:
Same error installing cluster with terraform helm provider.
```
Error: rpc error: code = Unknown desc = YAML parse error on victoria-metrics-cluster/templates/clusterrole.yaml: error converting YAML to JSON: yaml: line 7: did not find expected key
```

tenmozes:
hello thanks for a report will fix ASAP

tenmozes:
fixed, please update repo

kostyaplis:
Confirm. Works now. Thanks!

tenmozes:
@kostyaplis thank you, closing an issue, feel free to reopen

PR #11, opened 2019-12-20
fix extraLabels indentation

Meroje:
This is related to #6, using extraLabels will give the `error converting YAML to JSON` error because the first label has an extra indent.


tenmozes:
@Meroje thank for contribution, i will verify the fix, eta today :)

tenmozes:
@Meroje thanl for a fix

tenmozes:
fix is released in 0.4.1

PR #17, opened 2020-03-10
Add initial version of vmagent chart

afoninsky:


charts/victoria-metrics-agent/templates/deployment.yaml
```diff
@@ -0,0 +1,67 @@
...
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
+          args:
+            - -promscrape.config=/config/scrape.yml
+            {{- toYaml .Values.args | nindent 12 }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          # TODO: implement xhecks
```
tenmozes: could you please get checks from other charts? I prefer to use tcp probe for liveness and http for readiness

charts/victoria-metrics-agent/values.yaml
```diff
@@ -0,0 +1,67 @@
...
+  #   memory: 128Mi
+  # requests:
+  #   cpu: 100m
+  #   memory: 128Mi
+
+nodeSelector: {}
+
+tolerations: []
+
+config: |
```
tenmozes: please uncomment and leave self-scrape only

charts/victoria-metrics-agent/values.yaml
```diff
@@ -0,0 +1,67 @@
...
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name:
+
+args:
```
tenmozes: the basic idea of every chart is to work by default, do you have any idea how this can be achieved?

tenmozes:
LGTM, thank you

PR #18, opened 2020-03-14
Making it possible to set namespace on ServiceMonitor manifest

langecode:
It is often common practice to place the ServiceMonitor manifest in another namespace, e.g., `monitoring`. Thus, it would be nice to have the ability to optionally set the target namespace for the ServiceMonitor.


tenmozes:
thank you for contribution, looks good to me, will do release today evening

PR #19, opened 2020-03-16
Namespace for ServiceMonitor

langecode:
Adding possibility of setting ServiceMonitor namespace also for single container version (see also #18 for cluster version).


tenmozes:
@langecode thanks

PR #25, opened 2020-04-10
Use the same service for statefulSet and deployment modes

### LINKED ISSUE #24: Several issues in victoria-metrics-single
Found several issues using the new chart version:
- deployment mode exists (server.statefulSet.enabled = false) but there is deployment resource defined
- ClusterIP is defined to None in the service resource defined in statefulSet mode
- Most options aren't available in the service resource defined in statefulSet mode (type, ExternalIPs...)

Will provide PR soon.

NaPs:
There is no reason to use an headless service in that case. The
statefulSet resource is used here to maintain access to the same volume
even if the workload is redeployed.

As the statefulSet is always scaled to 1, the service is used to access
the workload and not to do an auto discovery of each replicas.

So:

- The headless service has been removed
- The service section in the statefulSet section (in values.yaml) has
  been removed
- The service resource is now used for both workload modes

This closes #24.


tenmozes:
please bump version of chart it brings incompatible changes

NaPs:
Done!

tenmozes:
in case of new added deployment you may setup VM using it and pre-created pvc(pv) with full features of services,

NaPs:
@tenmozes humm I'm not sure, are you expecting something from me?

tenmozes:
this is question, do we really need to remove stateless service now, when user can choose what to deploy sts or deployment? also what about backward comparability?

NaPs:
I'm not sure to understand what you mean here.

Before my patch:

- The deployment mode have its own "full-feature" service (including ClusterIP, ExternalIP etc)
- The statefulset mode have its own "headless" service (no ClusterIP, no ExternalIP etc)

After my patch:

- The deployment mode keep its service
- The statefulset mode now use the same service as deployment mode

The changes in values.yaml file involve a breaking if somebody used the settings I removed but as it stands, the chart isn't usable since the deployment itself is missing in deployment mode and the service doesn't have a ClusterIP in the STS mode...

tenmozes:
i don't want to break existing configuration of running charts, e.g. user runs some custom stateless service with own port, after the release of these changes service will be changed that may cause problems with an existing working app

tenmozes:
close due to age

PR #26, opened 2020-04-14
Add storageClass parameter in victoria-metrics-single to README

tanelso2:
I noticed this parameter was undocumented but still usable, so I added it to the README


tenmozes:
thanks for contribution

PR #29, opened 2020-04-16
Fix server deployment upgrade requiring a manual pod delete

tanelso2:
When a pvc is attached, the deployment won't be able to create a new pod
after an upgrade until the first pod dies/is deleted. Changing the deployment
strategy to 'Recreate' allows upgrades to proceed as expected.


tenmozes:
i will test PR today and merge

tenmozes:
@tanelso2 thank you for contribution

tenmozes:
released 0.3.2 version with fix

PR #32, opened 2020-04-24
added annotations and podAnnotations in helm chart vmagent

Denissa89:
added annotations and podAnnotations in helm chart vmagent


tenmozes:
thanks for contribution

tenmozes:
the new version of agent chart has been released

PR #33, opened 2020-05-06
Add missing namespace to single chart

charandas:
Current single chart is missing namespace for statefulset, and service objects.

Maybe other charts here are also missing it, but haven't had the time to inspect.


tenmozes:
@charandas thank you for contribution

charandas:
@tenmozes will you bump the chart version and republish?

tenmozes:
i'm going to add missing namespace annotations that bump version and publish

PR #36, opened 2020-05-12
Fix kubelest cadvisor forbidden scrape problem

artemsre:
GCE k8s cluster a have got error:

```
1s, error="unexpected status code returned when scraping \"https://10.138.0.121:10250/metrics\": 403; expecting 200; response body: \"Forbidden (user=system:serviceaccount:devops:vmagent-gcp-stg-victoria-metrics-agent, verb=get, resource=nodes, subresource=metrics)\""
```

add one more resource to cluster role


tenmozes:
@artemsre thank you for contribution

PR #40, opened 2020-05-22
fix: remove duplicated key affinity in sts

qtheya:
This PR is fixing yaml linter error : `error    duplication of key "affinity" in mapping  (key-duplicates)`


tenmozes:
thank you for contribution

tenmozes:
the new version of cluster version is available

PR #41, opened 2020-05-22
Automatically roll vmalert deployment

schndr:
- Restart pods when configmap updates
- bump vmalert chart version


tenmozes:
@schndr thank you :)

PR #43, opened 2020-06-11
Add alertmanager, ability to specify custom configmaps

afoninsky:


charts/victoria-metrics-agent/values.yaml
```diff
@@ -76,6 +75,11 @@ affinity: {}
 
 # vmagent scraping configuration:
 # https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/docs/vmagent.md#how-to-collect-metrics-in-prometheus-format
+
+# use existing configmap if specified
+# otherwise .config values will be used
+configMap: ""
```
tenmozes, line 81: please create a helper which can define what name should be used

charts/victoria-metrics-agent/values.yaml
```diff
@@ -27,7 +27,6 @@ serviceAccount:
   name:
 
 # WARN: need to specify at least one remote write url
-remoteWriteUrls: []
```
tenmozes, line 30: please don't delete the var

charts/victoria-metrics-alert/test.yaml
```diff
@@ -0,0 +1,33 @@
+server:
```
tenmozes, line 1: should it be here?

charts/victoria-metrics-agent/values.yaml
afoninsky: sorry, forgot to return

charts/victoria-metrics-agent/values.yaml
afoninsky: nice trick, thanks for pointing it

PR #45, opened 2020-06-25
Update vmselect-statefulset.yaml

paprickar:
Hey everyone,

I just discovered that Pod Security Policy is not working for vmselect as the Service Account was missing.

Therefore it used the default Service Account which had no permission to use the created PSP.

I would love to see this upstream - let me know if I should change something.


tenmozes:
thanks for contribution

PR #51, opened 2020-07-20
Change working dir to /tmp in vmagent chart

ulikl:
"/" may not be writable in restricted containers

### DIFF
--- a/charts/victoria-metrics-agent/templates/deployment.yaml
+++ b/charts/victoria-metrics-agent/templates/deployment.yaml
@@ -39,6 +39,7 @@ spec:
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
+          workingDir: /tmp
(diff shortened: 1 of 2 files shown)


charts/victoria-metrics-agent/templates/deployment.yaml
```diff
@@ -39,6 +39,7 @@ spec:
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
           image: "{{ .Values.image.repository }}:{{ default .Chart.AppVersion .Values.image.tag }}"
+          workingDir: /tmp
```
tenmozes, line 42: this is not a backward compatible changes, so please move `workingDir` to values.yaml (the default value should be "/") and bump chart version

ulikl: you are right. I adapted it.

tenmozes:
thank you for contribution, will do release soon

tenmozes:
the new version of helm chart for vmagent (0.4.9) contains `containerWorkingDir` variable

PR #53, opened 2020-07-29
Allow regular service to be used with StatefulSet

RichieSams:
A headless service is required for the statefulset itself, but unfortunately, a headless service doesn't support all the features of regular services. (Like LoadBalancer type, with static IP)

This change renames the headless service to "fullname-headless" and additionally creates the regular service for statefulsets.

I bumped the minor version to signal users that there are potentially breaking changes.


tenmozes:
Hi @RichieSams thatk you for contribution, sorry for the silence, i will review PR soon

maxpain:
Any news?

PR #57, opened 2020-08-27
Add optional pod labels

schndr:
This change adds optional pod labels to vmselect, vminsert and vmstorage.

I bumped the version in case of issues.

### DIFF
--- a/charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml
+++ b/charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml
@@ -8,7 +8,7 @@ metadata:
 {{- end }}
   labels:
     {{- include "victoria-metrics.vminsert.labels" . | nindent 4 }}
-{{- if .Values.vminsert.labels }}
+{{- with .Values.vminsert.extraLabels }}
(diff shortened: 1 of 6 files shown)


charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
 {{- end }}
   labels:
     {{- include "victoria-metrics.vminsert.labels" . | nindent 4 }}
+{{- if .Values.vminsert.labels }}
```
tenmozes, line 11: please use `with` instead of if

charts/victoria-metrics-cluster/values.yaml
```diff
@@ -146,6 +147,7 @@ vminsert:
   fullnameOverride: ""
   extraArgs: {}
   annotations: {}
+  labels: {}
```
tenmozes, line 150: please rename to extraLabels as far some labels are already present

tenmozes:
@schndr thank you for contribution :)

PR #62, opened 2020-09-11
Add 'extraHostPathMounts' option for vmagent

quite4work:
Additional hostPath mounts for vmagent, [same option as in Prometheus Chart][1].

[1]: https://github.com/helm/charts/blob/master/stable/prometheus/values.yaml#L660


tenmozes:
LGTM, @quite4work thanks for contribution i will merge and release new version of chart soon

Issue #65, opened 2020-09-17
Duplicate Scrape Target Error on Kube-DNS


silverbp:
We're getting this error with the default configuration in helm when installing in k8s in AWS (EKS).

error	VictoriaMetrics/lib/promscrape/scraper.go:269	skipping duplicate scrape target with identical labels; endpoint=http://10.100.30.94:9153/metrics, labels={eks_amazonaws_com_component="kube-dns", instance="10.100.30.94:9153", job="kubernetes-service-endpoints", k8s_app="kube-dns", k8s_cluster_name="k8s-devops", kubernetes_io_cluster_service="true", kubernetes_io_name="CoreDNS", kubernetes_name="kube-dns", kubernetes_namespace="kube-system", kubernetes_node="ip-10-100-30-216.us-east-2.compute.internal"}; make sure service discovery and relabeling is set up properly

I haven't figured how how to fix this one yet as the documentation has suggestions for fixing this issue when it comes up on a pod. However, it would be great if the default config worked with core k8s components out of the box without logging errors.

tenmozes:
hello @silverbp please check the https://github.com/VictoriaMetrics/VictoriaMetrics/tree/master/app/vmagent#troubleshooting 

section:
```
If you see skipping duplicate scrape target with identical labels errors when scraping Kubernetes pods, then it is likely these pods listen multiple ports or they use init container.
```

silverbp:
@tenmozes this error is coming up in the kubernetes-service-endpoints job, not the pod job.

silverbp:
we're using the default config that's in helm.

tenmozes:
@valyala could you pleas take a look?

silverbp:
for context, this is the kube-dns service definition in k8s.

```
apiVersion: v1
kind: Service
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
[...]

PR #66, opened 2020-09-17
changes keep rule for pods and services

f41gh7:
https://github.com/VictoriaMetrics/helm-charts/issues/65

if prometheus_io_port annotation is defined, it keeps only targets if prometheus_io_port matches pod_container_port

prometheus_io_scrape must present as well

target ports from init containers is dropped

https://github.com/VictoriaMetrics/VictoriaMetrics/tree/master/app/vmagent\#troubleshooting

### DIFF
--- a/charts/victoria-metrics-agent/values.yaml
+++ b/charts/victoria-metrics-agent/values.yaml
@@ -262,6 +262,11 @@ config:
       kubernetes_sd_configs:
         - role: endpoints
       relabel_configs:
+        - action: drop
+          source_labels: [__meta_kubernetes_pod_container_init]
+          regex: true
+        - action: keep_if_equal
+          source_labels: [__meta_kubernetes_pod_annotation_prometheus_io_port, __meta_kubernetes_pod_container_port_number]
         - source_labels:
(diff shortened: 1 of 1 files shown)


silverbp (commented):
I tested those changes in our setup and it got rid of the metrics it was pulling from 'kubernetes-service-endpoints'. Adding those to the pod I think makes sense, not sure it makes sense for endpoints as they may not be backed by a pod? It's really just the Kube-DNS one that causes me issues on the endpoints.

f41gh7:
> I tested those changes in our setup and it got rid of the metrics it was pulling from 'kubernetes-service-endpoints'. Adding those to the pod I think makes sense, not sure it makes sense for endpoints as they may not be backed by a pod? It's really just the Kube-DNS one that causes me issues on the endpoints.

 Well, it solves duplicate error, if services is backed by pod and prometheus.io/port annotation matches pod port. It could lead to side effects:
- you have only `prometheues.io/scrape: true` annotation, without port specification, such targets will be dropped.
- endpoints without pods dropped as well.


 I suggest to leave this PR as example, how to solve possible possible duplicates errors for now. We will discuss possible suppression for this errors.

PR #70, opened 2020-10-08
Promote rbac.authorization.k8s.io to v1

weisdd:
`rbac.authorization.k8s.io/v1beta1` was deprecated in k8s 1.17, and will no longer be served in v1.22.
In the VictoriaMetrics' charts we can see a mix of both `v1beta1` and `v1`. I think it's time to promote everything to `v1`.
image


tenmozes:

PR #71, opened 2020-10-09
Drop all required capabilities for psp. Link vmagent psp with clusterrole

iordachelm:
1. Since most of the default docker capabilities are dropped, remaining ones are not needed also. This was validated by running tests.
2. PSP for vmagent was not binded with existing clusterrole, so decided to link it in order to be able to use it.


tenmozes:

Issue #64, opened 2020-09-17
vm-agent persistent storage support


afoninsky:
VM uses local fs to keep buffer in case if remote_write endpoint is not available. Currently, there is no possibility to mount emptyDir/persistent volume as a storage.

1. Data will be lost after pod restart.
2. Service exists with error if following security option is set: "securityContext.readOnlyRootFilesystem = true"
3. It's not a good practice to write into root fs in containers.

Recommended default behaviour:
- create emptyDir with possibility to change to other storages
- set reasonable default "remoteWrite.maxDiskUsagePerURL" to avoid huge disc usage
- point "remoteWrite.tmpDataPath" to a mounted storage folder

p.s.: helm chart has pod security policy but it's not enabled in the role -> setting this flag in values.yaml will be ignored:
https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-agent/templates/clusterrole.yaml

tenmozes:
@afoninsky is it still relevant?

f41gh7:
Now vmagent supports statefulset configuration, https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-agent/values.yaml#L20 it must solve this issue.

PR #72, opened 2020-10-09
Use emptyDir volume to store the vmagent buffer

hamelg:
It fixes partially this issue :
https://github.com/VictoriaMetrics/helm-charts/issues/64


tenmozes:

PR #87, opened 2020-11-02
Bugfix for retention period smaller than one month (vmstorage/single)

weisdd:
**Description**
Starting from v1.45.0, VM allows to have retention periods smaller than one month (https://github.com/VictoriaMetrics/VictoriaMetrics/issues/173).

The issue is that the current template fails to render non-numerical values like `5d`, `1w` - they produce  `--retentionPeriod=0`

**Bugfix**
The proposed bugfix changes the respective templates to let them treat `.Values.vmstorage.retentionPeriod` and `.Values.server.retentionPeriod` as strings instead of numbers. So, values in any format become possible.

### DIFF
--- a/charts/victoria-metrics-cluster/templates/vmstorage-statefulset.yaml
+++ b/charts/victoria-metrics-cluster/templates/vmstorage-statefulset.yaml
@@ -46,7 +46,7 @@ spec:
           securityContext:
             {{- toYaml .Values.vmstorage.podSecurityContext | nindent 12 }}
           args:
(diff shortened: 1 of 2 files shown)


weisdd:
This is a bugfix for #88.

f41gh7 (approved):
LGTM

weisdd:
@valyala could you, please, take a look at the PR? :) Thanks!

f41gh7:
sorry, forgot to merge.

PR #93, opened 2020-11-17
Add example scrape config for other charts services

imcitius:
Changed victoria-metrics-agent chart's default scrape_config, added configs for services deployed by victoria-metrics-cluster chart.


tenmozes (changes requested):
This change will introduce scrape errors if cluster in not deployed, so please move comment or add flag to values (by default false) and describe this section in a docs

PR #95, opened 2020-11-18
Set proper name of ServiceAccount for Alert and AlertManager Deployments

iordachelm:
While testing psp for VM components noticed that Alert and Alertmanger pods cannot start when their psp is enforced to be used. Noticed that in their deployments **default** ServiceAccount instead of dedicated SA for these two components.
Enhanced the Alert chart to have proper name of SA set in deployment


tenmozes:
Thank you for contribution, LGTM

PR #99, opened 2020-12-22
addded deploymet stratagy for vm-agent

yuriydzobak:
The victoriam metrics agent doesn't have the deployment strategy
I faced with the issue when you use PVC and re-create the deployment, it can't re-create POD because the previous POD is using PVC volume, to resolve such issue, deployment strategy should be defined `recreate`
I've just added this parameter `deploymentStratagy`


yuriydzobak:
duplicate https://github.com/VictoriaMetrics/helm-charts/pull/100

PR #100, opened 2020-12-22
VictoriaMetrics Agent Stafullset implementation

yuriydzobak:
The helm chart can't work with more one replicate because can't use ReadWriteOnce PV
So, I made a statefulset of VM agent and it helps to anybody to switch between deployment or statefulset
Also, I keep the old relation of deployment and PVC. 
Update from old version of chart to this version won't break anything but in my point of view, PVC should be removed from Deployment and disable the ability to turn on
I've bumped the version to `0.7.0`. I don't know your release strategy


f41gh7 (approved):
Thanks for contribution. Looks good for me.

charts/victoria-metrics-agent/templates/statefulset.yaml
```diff
@@ -0,0 +1,145 @@
...
+  labels:
+    {{- include "chart.labels" . | nindent 4 }}
+  {{- with .Values.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  replicas: {{ .Values.replicaCount }}
+  serviceName: {{ include "chart.fullname" . }}
+  updateStrategy:
```
tenmozes, line 16: please don't specify the strategy if it's not defined

charts/victoria-metrics-agent/templates/statefulset.yaml
```diff
@@ -0,0 +1,145 @@
...
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+            - name: http
+              containerPort: 8429
+          {{- with .Values.env }}
+          env: {{ toYaml . | nindent 10 }}
+          {{- end }}
+          readinessProbe:
+            httpGet:
+              path: /health
```
tenmozes, line 62: @f41gh7 is it correct probe

charts/victoria-metrics-agent/values.yaml
```diff
@@ -4,6 +4,22 @@
 
 replicaCount: 1
 
+## Deployment or StatefulSet
+# Deployment ref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/
+# StatefulSet ref:  https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/
+kind: Deployment
+
+## WARN: If kind is Deployment that uses
+deploymentStrategy: {}
```
tenmozes, line 13: don't  need to add comment if keys will already contain type of strategy eg.
```
updateStrategy
 statefulset:
 deployemnt:
 ```

**Review by tenmozes — CHANGES_REQUESTED** [review]

File: charts/victoria-metrics-agent/templates/deployment.yaml [hunk 4]
```diff
@@ -11,6 +12,8 @@ metadata:
   {{- end }}
 spec:
   replicas: {{ .Values.replicaCount }}
+  strategy:
```
tenmozes, line 15: please specify strategy if value is not empty

tenmozes:
@yuriydzobak Thank you for contribution :)

charts/victoria-metrics-agent/templates/statefulset.yaml
yuriydzobak: should be https://github.com/VictoriaMetrics/helm-charts/blob/25c519aa178d8b7770619a7900541ec27e11576d/charts/victoria-metrics-agent/templates/deployment.yaml#L58

charts/victoria-metrics-agent/templates/deployment.yaml
yuriydzobak: i've added condition

charts/victoria-metrics-agent/values.yaml
yuriydzobak: replaced adn added ref to documentation

charts/victoria-metrics-agent/templates/statefulset.yaml
yuriydzobak: fixed, please, check it again

charts/victoria-metrics-agent/values.yaml
```diff
@@ -4,6 +4,24 @@
 
 replicaCount: 1
 
+## Deployment or StatefulSet
+# Deployment ref: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/
+# StatefulSet ref:  https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/
+kind: Deployment
```
tenmozes, line 10: actually, what do you think about  defining the top level properties
statefulset:
 enabled: bool
 fileds related to sts
deployemnt:
 enabled: bool
  ....
?

yuriydzobak: it could be done like this, and any issue I don't see. I'll make the commit

Only, I don't like the PVC https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-agent/templates/pvc.yaml because it can confuse people, between PVC template in statefulset and PVC in deployment but if it removed, the users faced the issue, they should switch from deployment to statefulset,and remove old PVC who is using chart now. Maybe do you have an idea ?

tenmozes: thank you, as sts a new entity we can define  own pvc inside an object

yuriydzobak: > actually, what do you think about defining the top level properties
> 
> ```yaml
> statefulset:
>  enabled: bool
>  fileds related to sts
> deployemnt:
>  enabled: bool
>   ....
> ```
> 
> ?

Are we able to enabled statefulset and deployment at once?
or if statefulset is enabled then deployment should be disabled automaticly?

yuriydzobak: So, it could be enabled deployment and statefulset at one moment =) I guess users will see the issue if they  are enabled this two option at once

tenmozes:
@yuriydzobak thank you for contribution

tenmozes:
i will change default behavior to deployment for backward compatibility sake

tenmozes:
the new version of agent have been released, thanks

Issue #89, opened 2020-11-09
Bug: vmselect deployment should use selectNode argument to support metrics deletion


fsperling:
Hi,

deleting a metric via:
https://fqdn/delete/0/prometheus/api/v1/admin/tsdb/delete_series?match[]=metric-name

results in the following error in the vmselect log:

`2020-11-09T08:23:49.299Z        panic   VictoriaMetrics/app/vmselect/prometheus/prometheus.go:304       BUG: missing -selectNode flag
2020-11-09T08:23:49.300Z        error   net/http/server.go:3093 http: panic serving 10.28.1.3:36772: BUG: missing -selectNode flag
goroutine 20556321 [running]:
net/http.(*conn).serve.func1(0xc0252c83c0)
        net/http/server.go:1801 +0x147
panic(0x9d2860, 0xc025739c80)
        runtime/panic.go:975 +0x3e9
github.com/VictoriaMetrics/VictoriaMetrics/lib/logger.logMessage(0xa4d4fa, 0x5, 0xc02574b480, 0x1d, 0x4)
        github.com/VictoriaMetrics/VictoriaMetrics/lib/logger/logger.go:191 +0xa0c
github.com/VictoriaMetrics/VictoriaMetrics/lib/logger.logLevelSkipframes(0x1, 0xa4d4fa, 0x5, 0xa5c44a, 0x1d, 0x0, 0x0, 0x0)
        github.com/VictoriaMetrics/VictoriaMetrics/lib/logger/logger.go:124 +0xd1
github.com/VictoriaMetrics/VictoriaMetrics/lib/logger.logLevel(...)
        github.com/VictoriaMetrics/VictoriaMetrics/lib/logger/logger.go:116
github.com/VictoriaMetrics/VictoriaMetrics/lib/logger.Panicf(...)
        github.com/VictoriaMetrics/VictoriaMetrics/lib/logger/logger.go:112
github.com/VictoriaMetrics/VictoriaMetrics/app/vmselect/prometheus.resetRollupResultCaches()
        github.com/VictoriaMetrics/VictoriaMetrics/app/vmselect/prometheus/prometheus.go:304 +0x3ef
github.com/VictoriaMetrics/VictoriaMetrics/app/vmselect/prometheus.DeleteHandler(0xbfe25d054d2d8760, 0x174d1191d6a4e, 0xda5540, 0xc02c041638, 0xc01fa24a00, 0x2a, 0xc025739ba0)
[...]

PR #101, opened 2020-12-22
adds selectNode param for vmselect,

f41gh7:
fixes issue https://github.com/VictoriaMetrics/helm-charts/issues/89

It possible to use --selectNode flag only with statefulset version of vmselect.

### DIFF
--- a/charts/victoria-metrics-cluster/templates/_helpers.tpl
+++ b/charts/victoria-metrics-cluster/templates/_helpers.tpl
@@ -156,6 +156,16 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 {{- end -}}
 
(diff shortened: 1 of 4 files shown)


charts/victoria-metrics-cluster/templates/_helpers.tpl
```diff
@@ -156,6 +156,16 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- end -}}
 {{- end -}}
 
+{{- define "victoria-metrics.vmselect.vmselect-pod-fqdn" -}}
```
tenmozes, line 159: need to bump version of chart

f41gh7: done, thanks!

tenmozes (approved):
thank  you :) LGTM

PR #106, opened 2021-02-01
When alertmanager enabled is true then check alertmanager.url

TinySong:


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #109, opened 2021-02-17
There are no custom labels for pods


auroaj:
```
{{/*
Create unified labels for victoria-metrics components
*/}}
{{- define "victoria-metrics.common.matchLabels" -}}
app.kubernetes.io/name: {{ include "victoria-metrics.name" . }}
app.kubernetes.io/instance: {{ .Release.Name }}
{{- end -}}

{{- define "victoria-metrics.common.metaLabels" -}}
helm.sh/chart: {{ include "victoria-metrics.chart" . }}
app.kubernetes.io/managed-by: {{ .Release.Service }}
{{- end -}}

{{- define "victoria-metrics.server.labels" -}}
{{ include "victoria-metrics.server.matchLabels" . }}
{{ include "victoria-metrics.common.metaLabels" . }}
{{- end -}}

{{- define "victoria-metrics.server.matchLabels" -}}
app: {{ .Values.server.name }}
{{ include "victoria-metrics.common.matchLabels" . }}
{{- end -}}
```
Sometimes you need to configure additional labels for pods.
Example from grafana:
```
  template:
    metadata:
      labels:
        {{- include "grafana.selectorLabels" . | nindent 8 }}
{{- with .Values.podLabels }}
{{ toYaml . | indent 8 }}
{{- end }}
```
https://github.com/grafana/helm-charts/blob/86c43461578f0d53d9f3edc49a2a36e4a984e7a7/charts/grafana/templates/deployment.yaml#L30

PR #108, opened 2021-02-08
add podLabels and priorityClassName params to victoria-metrics-alert

t3mi:

PR #114, opened 2021-03-01
#109 custom labels for victoria-metrics-single

auroaj:
Did the same as https://github.com/VictoriaMetrics/helm-charts/pull/108
@f41gh7 

What do you think, would it be better to add custom labels to `{{- define "victoria-metrics.server.labels" -}}`?

https://github.com/VictoriaMetrics/helm-charts/issues/109


f41gh7 (approved):
LGTM

f41gh7:
I guess, its possible to add label declaration into helpers template later. For now, its good. 

Thanks for contribution!

PR #115, opened 2021-03-03
[vmalert] allow empty config if rule file specified

Scott8440:
When deploying the vmalert chart, templating will fail if neither a config nor an existing configmap are provided in the values. There is another way to provide rules to vmalert though, through the `rule` argument. This PR allows a deployment with no config or configmap as long as an additional rule parameter is provided.


Scott8440:
It might be a good idea to put `rule` as a `.server` level value, but I'm leaving it as-is (in extraArgs) in order to make fewer changes. I'd be happy to make that change if you think it would be better though.

f41gh7:
Thanks for contribution!
 
 Lets leave it as-is for now.

PR #119, opened 2021-03-18
Added go_template for readiness & liveness probes (initialDelaySeconds)

kvitals:
we have had an issue with an 'initialDelaySeconds' on readiness probe - and predefined 5 seconds wasn't enough. So, we need this change to increase the value for this option. Kindly hope you'll find it useful ;) thanks

### DIFF
--- a/charts/victoria-metrics-cluster/values.yaml
+++ b/charts/victoria-metrics-cluster/values.yaml
@@ -49,6 +49,19 @@ vmselect:
   extraLabels: {}
   env: []
 
+  # Readiness & Liveness probes
+  probe:
+    readiness:
+      initialDelaySeconds: 5
+      periodSeconds: 15
+      timeoutSeconds: 5
+      failureThreshold: 3
+    liveness:
+      initialDelaySeconds: 5
+      periodSeconds: 15
+      timeoutSeconds: 5
+      failureThreshold: 3
+
   # Additional hostPath mounts
   extraHostPathMounts:
    []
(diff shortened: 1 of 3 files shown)


charts/victoria-metrics-cluster/values.yaml
```diff
@@ -49,6 +49,19 @@ vmselect:
   extraLabels: {}
   env: []
 
+  # Readiness & Liveness probes
+  probe:
+    readiness:
+      initialDelaySeconds: 5
+      periodSeconds: 15
+      timeoutSeconds: 5
```
tenmozes, line 57: this param can confuse as it is not been used in template but defined

charts/victoria-metrics-cluster/values.yaml
```diff
@@ -49,6 +49,19 @@ vmselect:
   extraLabels: {}
   env: []
 
+  # Readiness & Liveness probes
+  probe:
```
tenmozes, line 53: could you please use all params from values file in template or remove them

btw what about vminsert and storage
do you think, can this be helpful for other components?

tenmozes (changes requested):

charts/victoria-metrics-cluster/values.yaml
kvitals: understood, already fixed it)

kvitals:
hi @tenmozes 
Could you please check and approve?)

tenmozes:
Thank you, i will test and approve, eta 1d

PR #120, opened 2021-03-26
Us aecio 6

aeciopires:
**what was done?**

Adjustments in the main README to inform the new versions of the chart and application.

**Why was it done?**

To keep the content up to date.


tenmozes:
thank you for contribution

PR #122, opened 2021-04-03
added extrasecret and initcontainers

shnurok672:
This changes are useful for backup/restore processes. Extra secret are used to provide s3 credentials for vmrestore(for ex.). Init containers are also may be used to restore VM from backup.

### DIFF
--- a/charts/victoria-metrics-cluster/templates/vmstorage-statefulset.yaml
+++ b/charts/victoria-metrics-cluster/templates/vmstorage-statefulset.yaml
@@ -39,6 +39,12 @@ spec:
       schedulerName: "{{ .Values.vmstorage.schedulerName }}"
 {{- end }}
       automountServiceAccountToken: {{ .Values.serviceAccount.automountToken }}
+      {{- if .Values.vmstorage.initContainers }}
+      initContainers:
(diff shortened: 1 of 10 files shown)


charts/victoria-metrics-cluster/templates/vmstorage-statefulset.yaml
```diff
@@ -39,6 +39,12 @@ spec:
       schedulerName: "{{ .Values.vmstorage.schedulerName }}"
 {{- end }}
       automountServiceAccountToken: {{ .Values.serviceAccount.automountToken }}
+      {{- if .Values.vmstorage.initContainers }}
```
f41gh7, line 42: it can be simplified to 

```yaml
     {{- with .Values.vmstorage.initContainers }}
      initContainers:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

 also, can you add it to other cluster components and single version?

shnurok672: Ok, done

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #124, opened 2021-04-06
Add ports for all ingestion protocols to vmagent

sti-jans:
Added the ability to add ports for all ingestion protocols in vmagent

### DIFF
--- a/charts/victoria-metrics-agent/Chart.yaml
+++ b/charts/victoria-metrics-agent/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
 name: victoria-metrics-agent
 description: Victoria Metrics Agent - collects metrics from various sources and stores them to VictoriaMetrics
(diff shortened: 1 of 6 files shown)


f41gh7:
It looks good to me and should work.

 But probably its better to introduce additional config option:

smthg like:
```yaml
insertPorts:
 graphite: 
   enabled: false
   port: 1112
 influx:
   enabled: false
   port: 1113
 openTSDB:
  enabled: false
  port: 1114
```
 And add it explicitly? 
 It will be documented and more obvious.

 What do you think?

sti-jans:
I agree.
The current implementation was made by analogy with other charts.
In the args section, it is better to set these values before `.Values.extraArgs` ?

f41gh7:
Can you add example of usage to the chart readme and  lets merge it as is.

f41gh7:
Thanks for contribution!

PR #136, opened 2021-05-05
bump dependendies to actual version

kevinvirs:
bump dependendies to actual version
change depracated kube-state-metrics to actual from prometheus-community

PR #138, opened 2021-05-05
Remove unused dependency from README

lexfrei:
Since #136 this repo is not used


f41gh7:
Thanks for contribution!

PR #139, opened 2021-05-07
Update kube-controller-manager.yaml

kevinvirs:
incorrect job label generate alert "kubecontroller manager is down" into vmalert


f41gh7:
Thanks!

PR #145, opened 2021-05-13
comment external labels in default values

kevinvirs:
by default external label "cluster: cluster-name" merged with overwrite by custom values
as result - excess label:
```
spec:
   externalLabels:
     cluster: cluster-name
     k8s_cluster: k8s-test
```


k1rk:
this label is mandatory as all dashboards depends on it

kevinvirs:
label "cluster" conflict with some application for me :)
i'm rewrite dashboard tag from "cluster" to "k8s_cluster"

PR #146, opened 2021-05-19
Fix scrape config for kubernetes-service-endpoints

hendrikmaus:
I noticed that the example config was referring to a pod label in the service configs, hence all the service metrics were dropped.


tgruenert (commented):
After running into the same issue i can confirm that this solution works.

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #161, opened 2021-06-16
in case single is disabled

vladimir4862:
just disable single datasource. in case single is disabled.


k1rk (approved):
LGTM. thanks @vladimir4862

PR #162, opened 2021-06-16
moved templates outside alertmanager.yaml

bon3o:
Templates should be stored in a separate files outside alertmanager.yaml


k1rk (approved):

f41gh7:
Thanks for contribution!

PR #172, opened 2021-07-08
Add HPA Support to vmselect component

arnoldyahad:
Configuration to allow the vmselect component to be scaled based on its CPU Utilization. 
We are already using this in our production vm-clusters.

PR #175, opened 2021-07-13
Add HPA Support to vminsert component

arnoldyahad:
same as https://github.com/VictoriaMetrics/helm-charts/pull/172 just for the vminsert component.


tenmozes:
thank you for contribution the HPA will be included to the next release of cluster helm chart

PR #181, opened 2021-08-03
Add startup probe support

james64:
Startup of application is often more demanding on cpu/memory then rest of the runtime. Sometimes it is hard to tune readiness/liveness probes so that they allow application to start but are responsive enough for issues during normal operation. Startup probe can save the day here.

VictoriaMetrics is no difference. Full disk for example can cause compactification to be halted. After volume is enlarged startup of Victoria takes longer since it needs to catch up in maintenance backlog.

Thus I propose adding support for startup probes into official Helm chart.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #185, opened 2021-08-06
Codeowner rule fix

k1rk:
so now out of the box we can create alerts with label `code_owner_channel` and they will be routed to appropriate channel.
of course most users have their own routing configuration, but this is nice to have out of the box, and also can be an inspiration for users =)

also updated dependencies =)

### DIFF
--- a/charts/victoria-metrics-k8s-stack/Chart.lock
+++ b/charts/victoria-metrics-k8s-stack/Chart.lock
@@ -1,15 +1,15 @@
 dependencies:
 - name: victoria-metrics-operator
   repository: https://victoriametrics.github.io/helm-charts
-  version: 0.1.17
+  version: 0.1.19
 - name: kube-state-metrics
   repository: https://prometheus-community.github.io/helm-charts
-  version: 3.4.1
+  version: 3.4.2
 - name: prometheus-node-exporter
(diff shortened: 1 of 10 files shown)


charts/victoria-metrics-k8s-stack/Chart.lock
```diff
@@ -1,15 +1,15 @@
 dependencies:
 - name: victoria-metrics-operator
   repository: https://victoriametrics.github.io/helm-charts
-  version: 0.1.17
+  version: 0.1.19
```
tenmozes, line 4: Looks good but could you please bump chart version

tenmozes (changes requested):

charts/victoria-metrics-k8s-stack/Chart.lock
k1rk: what do you mean? 
`victoria-metrics-operator` chart used last available version. 

for this chart i bumped in [a3a3688](https://github.com/VictoriaMetrics/helm-charts/pull/185/commits/a3a3688e96126af65374ec2d7ee714f233386d7a)

f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #191, opened 2021-08-17
Fix grafana configmap

k1rk:
we had this dashboard for etcd 3.1 (probably by mistake)
but the problem is when chart was deployed with default name dashboard configmap ended up with name `victoria-metrics-k8s-stack-grafana` which overlaps with grafana's internal configmap for configuration (the one that contains grafana.ini)

i dropped this dashboard since we have already dashboard for etcd in `charts/victoria-metrics-k8s-stack/templates/grafana/dashboards/etcd.yaml`

also vminsert endpoint for cluster was missing `prometheus/`. now fixed

also like always updated all dependencies and bumped chart version


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #192, opened 2021-08-18
Fix cluster dashboard name

k1rk:
since cluster version and single version uses same name for dashboard file `victoriametrics.json` resulting dashboards were overlapping. fixed in `charts/victoria-metrics-k8s-stack/hack/sync_grafana_dashboards.py`

also updated upstream url for kube-prometheus dashboard and rules to use main branch


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #193, opened 2021-08-19
Extra alertmanagerTemplates and argocd releaseName issue

turbotankist:
1. allow set custom alertmanager templates as in prometheus-operator
```
alertmanager:
  templateFiles:
    template_1.tmpl: |-
      {{ define "hello" -}}
      hello, Victoria!
      {{- end }}
    template_2.tmpl:  |-
      {{ define "bye" -}}
      arrivederci!
      {{- end }}
```
2. We use it with argocd, and it uses label app.kubernetes.io/instance to define release and always overrides it.
And if set releaseName in argocd, then VMServiceScrape select incorrect labels.
I've set the "argocdReleaseOverride" value and can manage with it.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #197, opened 2021-08-31
Nodexporter servicescrape argo fix

k1rk:
in #193 `argocdReleaseOverride` was introduced
but for nodeexporter service scrape we initally were overriding `release` while argo actually doesn't touch release label and only overrides `app.kubernetes.io/instance`

also added vmalert dashboard, thanks @hagen1778


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #210, opened 2021-09-24
add alertmanager service, fixed var in ingress

krakazyabra:
for using in ingress added alertmanager's service. also fixed wrong variable path in alertmanager-ingress


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #213, opened 2021-09-25
[k8s-stack] Update dependent components and dashboards

k1rk:
- updated all dependency charts
- updated dashboards and rules from upstream
- made all dashboards readonly (since they managed by configmap anyway grafana doesn't allow to change them, but now we don't see annoying question from grafana)
- add common tag for all imported dashboards, so that it's easy to see in list which dashboard came with chart
- added nice and finally working dashboard for coredns 
CleanShot 2021-09-25 at 18 51 01


charts/victoria-metrics-k8s-stack/templates/grafana/dashboards/clusterbytenant.yaml
```diff
@@ -1,695 +0,0 @@
-{{- /*
```
f41gh7, line 1: I dont get why it was removed?

f41gh7 (approved):
LGTM

PR #216, opened 2021-09-26
update labeler rules

k1rk:


k1rk:
just noticed that labeler in our case actually removing labels instead of adding 🤷

f41gh7 (approved):
LGTM

PR #217, opened 2021-09-26
Fix drilldown grafana 8

k1rk:
Drill-down links doesn't work properly in grafana 8
https://github.com/kubernetes-monitoring/kubernetes-mixin/issues/659

so i added ugly fix in sync_grafana script

also updated VM components to 1.66.2


f41gh7 (approved):
LGTM

PR #220, opened 2021-09-27
Specify custom service account

afoninsky:
This MR fixes a bug when it's not possible to specify already existing service account for single/cluster VM versions


f41gh7:
Thanks for contribution!

PR #222, opened 2021-09-27
add topologySpreadConstraints to select deployment to make it possible to spread them between az

vladislavPV:
This Pr will allow to use `topologySpreadConstraints` to spread pods between zones in case you have a cluster per zone architecture.


vladislavPV:
@valyala  it would be nice if you could review this . Thank you!

dima-vm (approved):
LGTM

PR #225, opened 2021-09-28
Fix scrape config for kubernetes-service-enpoints (victoria-metrics-single)

shusugmt:
Identical to #146 but for victoria-metrics-single chart.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #230, opened 2021-10-03
Need info how to set flags via env


paragor:
There is no info info about how to set flags via env in readme.md in charts. Especially about how to set repeating flags.

tenmozes:
thank you for contribution, closing

PR #231, opened 2021-10-03
Add info how to set flags via env in readme.md (#230)

paragor:
related issue https://github.com/VictoriaMetrics/helm-charts/issues/230


tenmozes:
Thank you for contribution, merged

PR #235, opened 2021-10-07
add extraLabels for Pods

Vladikamira:
This PR is adding extra Labels to Pods, Deployments and StatefulSet spec


charts/victoria-metrics-agent/values.yaml
```diff
@@ -90,6 +90,9 @@ env:
   #       name: auth_secret
   #       key: password
 
+# extra Labels for Pods, Deployment and Statefulset
+extraLabels: {}
```
tenmozes, line 94: Hi we need to separate labels for two
- extraLabels - these are labels for sts, deploy
- podLabels - these are labels for pod under sts, deploy

Vladikamira: sure, I can do that 🆗 I was inspired from [vmcluster](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-cluster/templates/vmselect-statefulset.yaml#L11) chart which has `extraLabels` as common value for both `Pods` and `Statefulset`

PR #243, opened 2021-10-14
[victoria-metrics-single] Fix ClusterRole to include networking.k8s.io

shusugmt:
Fix victoria-metrics-single ClusteRole to include missing `networking.k8s.io` causing permission error when using `role: ingress` in `kubernetes_sd_configs`.

In victoria-metrics-agent chart the same fix is applied at [this commit](https://github.com/VictoriaMetrics/helm-charts/commit/5f41ebda18caab18d6a88ad43502da91411d0852).


tenmozes (approved):
thank you for contribution

PR #254, opened 2021-11-05
Add minReadySeconds and strategy fields

arramos84:
This change will help when ensuring new vmalert pods can be spun up within a deployment
and be running  their rules before the old pods spin down.


tenmozes:
@arramos84 thank you for contribution, merged

tenmozes (approved):

tenmozes:
`0.4.13` version of vmalert helm chart contains your improvements

PR #259, opened 2021-11-25
Include datasource.yaml for any datasource

tumd:
The datasource.yaml-boilerplate needs to be present if *any* datasource is enabled; including `.Values.grafana.additionalDataSources`.

Before this change the chart would fail if both `.Values.vmsingle.enabled` & `.Values.vmcluster.enabled` was set to `false`, but `.Values.grafana.additionalDataSources` was containing dashboard configuration.

I've only tested this with the above mentioned circumstances.


k1rk (approved):
@tumd thanks for contribution

PR #262, opened 2021-12-03
Allow setting jsonData for Grafana datasource(s)

hwuethrich:
We are using a polling interval of 1m (for snmp-exporter) and would like to change the default interval on the provisioned VM datasources in Grafana (see [docs](https://grafana.com/docs/grafana/latest/administration/provisioning/#json-data)). With this PR, we can change it as follows:

```yaml
grafana:
  sidecar:
    datasources:
      jsonData:
        timeInterval: "1m"
```


k1rk (approved):
LGTM. thanks @hwuethrich

Issue #265, opened 2021-12-19
[victoria-metrics-k8s-stack]  please add ability to use external victoria metrics cluster


cristicalin:
I would like to use an external VM cluster in which to publish metrics with `victoria-metrics-k8s-stack` by disabling the grafana, alertmanager and vmsingle or vmcluster objects but keep just the vmagent and vmalert agents to publish to a remote VM cluster.

I see that the [VMAgent CRD](https://github.com/VictoriaMetrics/operator/blob/master/docs/quick-start.MD#vmagent) and [VMAlert CRD](https://github.com/VictoriaMetrics/operator/blob/master/docs/quick-start.MD#vmalert) already have the capability to target a remote VM cluster but these are not supported by the `victoria-metrics-k8s-stack` chart.

k1rk:
hi @cristicalin. to write to external VM cluster you can just specify remote write in spec of vmagent
example:
`values.yaml`
```
vmagent:
  spec:
    remoteWrite:
      - url: "https://insert.vmcluster.domain.com/insert/0/prometheus/api/v1/write"
```

cristicalin:
Thank you @k1rk for the help, indeed I missed that you can add anything in the spec.

PR #268, opened 2021-12-24
feat: allow to set custom vmInsertEndpoint for own remoteWrite URL

Wouter0100:
We needed to deploy a custom vmInsertEndpoint. Right now we use the following to override it manually:
```
vmagent:
  spec:
    containers:
    - name: vmagent
      args:
      - -httpListenAddr=:8429
      - -promscrape.config=/etc/vmagent/config_out/vmagent.env.yaml
      - -promscrape.streamParse=true
      - -remoteWrite.maxDiskUsagePerURL=1073741824
      - -remoteWrite.tmpDataPath=/tmp/vmagent-remotewrite-data
      - -remoteWrite.url=x/api/v1/write
```

But this required us to manually override all args of the vmagent. We'd love to keep this automated and this PR makes that possible.

In that case we just have to define it as follows:
```
vmInsertEndpoint: x
```


k1rk:
@Wouter0100 you can actually specify it as spec for vmagent object
```
vmagent:
  spec:
    remoteWrite:
      - url: "https://insert.vmcluster.domain.com/insert/0/prometheus/api/v1/write"
```

see https://github.com/VictoriaMetrics/helm-charts/issues/265

PR #269, opened 2021-12-29
quote var value to avoid string-to-bool conversion

acondrat:
Kubernetes's Pod specification requires environment variable values to be strings. I run into an issue trying to set environment variables to "false" or "true" as those get converted to boolean.

The example bellow results in a templating error.
```
env:
  - name: TEST123
    value: "false"
```

```
v1.EnvVar.Value: ReadString: expects " or n, but found f, error found in #10 byte of ...|,"value":false},{"na|..., bigger context ...|nd":["manager"],"env":[{"name":"TEST123","value":false},{"name":"WATCH_NAMESPACE","value":""},{"name|...
```


f41gh7:
Thanks for contribution!

PR #274, opened 2022-01-28
[victoria-metrics-operator] Do not specify default resources

acondrat:
Hi,

This is a tiny change to prevent setting default resources for the Operator. Helm recommends not to specify default resources and to leave this as a conscious choice for the user.

We currently have two issues with default resources:
1. Using CPU limits leads to unnecessary throttling - https://github.com/kubernetes/kubernetes/issues/67577
2. We can not unset the cpu limit key in a sub-chart due to a bug in Helm - https://github.com/helm/helm/issues/9027

Most community helm charts do not set resources but provide guidance with some commented values instead. 

Thanks!

### DIFF
--- a/charts/victoria-metrics-operator/values.yaml
+++ b/charts/victoria-metrics-operator/values.yaml
@@ -65,13 +65,13 @@ serviceAccount:
   name: ""
 
 # -- Resource object
-resources: 
-  limits:
-    cpu: 120m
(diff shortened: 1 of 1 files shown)


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #279, opened 2022-02-07
Mixed helm chart values

megakid:
`.Values.kubelet.cadvisor` / `.Values.kubelet.probes` wrong way around.

Noticed this when reading around how to use the helm chart.  Pretty sure this is just a copy/paste error.


k1rk (approved):
thanks for contribution @megakid

PR #283, opened 2022-02-10
added remote overrides and vmagent.additionalRemoteWrite

bon3o:
Add an option to write to multiple remote storages alongside with vmsingle and use remote storage to store alert states while vmsingle beeing a  datasource.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
+++ b/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
@@ -107,9 +107,9 @@ VMAlert remotes
 */}}
 {{- define "victoria-metrics-k8s-stack.vmAlertRemotes" -}}
 remoteWrite:
-     url: {{ .Values.vmalert.remoteWriteOverride | default (include "victoria-metrics-k8s-stack.vmInsertEndpoint" .) }}
+     url: {{ include "victoria-metrics-k8s-stack.vmInsertEndpoint" . }}
 remoteRead:
-     url: {{ .Values.vmalert.remoteReadOverride  | default (include "victoria-metrics-k8s-stack.vmSelectEndpoint" .) }}
+     url: {{ include "victoria-metrics-k8s-stack.vmSelectEndpoint" . }}
 datasource:
      url: {{ include "victoria-metrics-k8s-stack.vmSelectEndpoint" . }}
 notifiers:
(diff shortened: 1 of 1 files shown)


charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
```diff
@@ -128,8 +128,12 @@ VMAlert spec
 VM Agent remoteWrite
 */}}
 {{- define "victoria-metrics-k8s-stack.vmAgentRemoteWrite" -}}
-remoteWrite:
-    - url: {{ include "victoria-metrics-k8s-stack.vmInsertEndpoint" . }}/api/v1/write
+{{- $addressList := list (dict "url" (printf "%s/%s" (include "victoria-metrics-k8s-stack.vmInsertEndpoint" . ) "api/v1/write")) -}}
```
tenmozes, line 131: as far i understand remoteWrite is an array of https://docs.victoriametrics.com/operator/api.html#vmagentremotewritespec 
so we need to add properties from spec to values, we can just merge object

k1rk:
@bon3o thanks for contribution

you can override vmalert remotes already by specifying them in 
```
vmalert:
  spec:
    remoteWrite:
      url: blabla
    remoteRead:
      url: blabla
```

this chart is configure a lot of stuff but idea that you cal always specify your own through spec. 

same goes for VMAgent
however if you need to add an aditional target for VMAgent you will need to specify both of them in spec

bon3o:
Ok, I understand how to add remotes to vmalert. But adding both remotes to vmagent manually is a bad idea. If external url is always the same, the internal is generated from the chart.

k1rk:
@bon3o 
can you explain a bit more your use case? or why do you think its a bad idea?

when you are filling values for your helm installation you should already know what will be the URL for your insert, or vmsingle 
it will always be `vmsingle-<chart-name>` or `vmsingle-<chart-name>.<chart-namespace>.svc.cluster.local` (`vminsert-<chart-name>` in case of cluster)
and second url as you mentioned is external, so also should be known 

so if for example you are installing this chart in different clusters, you will still be able to use same values (at least this portion)

```
vmagent:
  spec:
    remoteWrite:
      - url: vmsingle-<chart-name>
      - url: your-external-url
```

bon3o:
OK, I tested the solution with vmaler's spec and it works fine. I am still sure, that it would be better to add only external url for vmagent. Also added check if there are extra templates added. If we turn off monzo`s templates and have not defined any configmaps, then our secret with alertmanager.templateFiles is not mounted in the alertmanager pod.

f41gh7:
Thanks for pull request! Your changes was adopted and added to the k8s-stack chart.

 It can be used via following param:
https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-k8s-stack/values.yaml#L485


 More over, if both vmcluster and vmsingle enabled, chart by default configures each service as remoteWrite target for vmagent.

PR #286, opened 2022-02-17
allows to add additional remoteWrite urls for k8s-stack charts

f41gh7:
allows to use vmsingle and vmcluster deployment inside single chart


k1rk:
Looks good thanks. however one question. when both vmsingle and vmcluster will be enabled, vmagent will write to both. but vmalert will be only looking into vmsinge. is this expected?

f41gh7:
I think, that it's the best behavior for given setup. Otherwise, user always can override this option with vmalert.spec.

 More over, vmagent spec can be also override by vmagent.spec.

PR #291, opened 2022-03-02
vmagent: add possibility to configure pod priority class

ebensom:
On congested clusters the vmagent pod can be preempted in favor of higher priority pods, which might not be desirable in all cases.
This PR adds the possibility to configure `priorityClassName` for the `Deployment` and `StatefulSet` pods (by default it's unset).


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #303, opened 2022-04-02
add vmgateway helm chart

tenmozes:


charts/victoria-metrics-gateway/templates/ingress.yaml
```diff
@@ -0,0 +1,51 @@
+{{- if .Values.ingress.enabled }}
+{{- $ingressApiIsStable := eq (include "vmauth.ingress.isStable" .) "true" -}}
```
f41gh7, line 2: Should it be vmauth -> vmwagateway?

f41gh7 (approved):
LGTM

charts/victoria-metrics-gateway/templates/ingress.yaml
```diff
@@ -0,0 +1,51 @@
...
+{{- $ingressSupportsIngressClassName := eq (include "vmauth.ingress.supportsIngressClassName" .) "true" -}}
+{{- $ingressSupportsPathType := eq (include "vmauth.ingress.supportsPathType" .) "true" -}}
```
k1rk, line 4: this templates are missing from `_helpers.tpl`

also as @f41gh7 mentioned we should also use unique name like `vmgtw.ingres..`

charts/victoria-metrics-gateway/templates/service-monitor.yaml
```diff
@@ -0,0 +1,40 @@
+{{- if .Values.serviceMonitor.enabled -}}
+apiVersion: monitoring.coreos.com/v1
```
k1rk, line 2: should we use native VM resource here? i mean service scrape?

tenmozes: just to be compatible with other charts

f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #306, opened 2022-04-15
Add optional relabelings for serviceMonitors

acondrat:
Hi,

This is a small change to allow defining relabelings for serviceMonitors. Looks like nobody run `helm-docs` for a while and it updated a few readme files not related to this PR.

Thanks,
Arcadie

### DIFF
--- a/charts/victoria-metrics-gateway/templates/service-monitor.yaml
+++ b/charts/victoria-metrics-gateway/templates/service-monitor.yaml
@@ -37,4 +37,8 @@ spec:
       tlsConfig:
           {{- toYaml . | nindent 8 }}
       {{- end }}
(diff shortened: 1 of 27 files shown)


charts/victoria-metrics-gateway/templates/service-monitor.yaml
```diff
@@ -37,4 +37,8 @@ spec:
       tlsConfig:
           {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.serviceMonitor.relabelings }}
```
tenmozes, line 40: NIT: could you please use `with` instead of `if`

tenmozes (approved):
Thank you for contribution

PR #313, opened 2022-04-26
BugFix: change int to toString for the retentionPeriod in VMSingle deployment

weibo-zhao:
Hello VictoriaMetrics team,

we are currently use the victoria-metrics-single helm chart and found that our settings of retentionPeriod of vmsingle deployment is set to zero because of the int function. I notice that there is similar [pull request](https://github.com/VictoriaMetrics/helm-charts/pull/87) for the statefulset but the deployment is not changed. 

Best regards,
Weibo


dreamblack86:
I have also seen this bug. It would be great if this comes quickly in the main branch. Thanks

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #319, opened 2022-04-29
Can't use one port for TCP/UDP with LoadBalancer service type (OpenTSDB input)


ipeacocks:
I use Helm for installation of clustered VM version and OpenTSDB input. That's why in `values.yaml` I am using `opentsdbListenAddr` option:
```
...
  extraArgs:
    envflag.enable: "true"
    envflag.prefix: VM_
    loggerFormat: json
    opentsdbListenAddr: ":4242"
...
```
And service `type: LoadBalancer`
As results it produces such resulted yaml:
```
---
# Source: victoria-metrics-cluster/templates/vminsert-service.yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app: vminsert
    app.kubernetes.io/name: victoria-metrics-cluster
    app.kubernetes.io/instance: release-name
    helm.sh/chart: victoria-metrics-cluster-0.9.23
    app.kubernetes.io/managed-by: Helm
  name: release-name-victoria-metrics-cluster-vminsert
  namespace: default
spec:
  ports:
    - name: http
      port: 8480
      protocol: TCP
      targetPort: http
    - name: opentsdb-udp
      protocol: UDP
      port: 4242
      targetPort: opentsdb-udp
    - name: opentsdb-tcp
      protocol: TCP
      port: 4242
      targetPort: opentsdb-tcp

  selector:
    app: vminsert
    app.kubernetes.io/name: victoria-metrics-cluster
    app.kubernetes.io/instance: release-name
  type: "LoadBalancer"
```
So `opentsdbListenAddr` gotta open two OpenTSDB ports TCP and UDP (can OpenTSDB even work on UDP?). As a result of such description K8s opens only one UDP port:
```
NAME                                                  TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                         AGE
victoria-metrics-victoria-metrics-cluster-vminsert    LoadBalancer   10.11.82.14     10.15.12.10   8480:31219/TCP,4242:31862/UDP   42h
```
Of course it's possible to use also
```
  extraArgs:
...
    opentsdbListenAddr: ":4242"
    opentsdbHTTPListenAddr: ":4243"
```
But according to this https://github.com/VictoriaMetrics/VictoriaMetrics/issues/266#issuecomment-566759632  it's not needed now and that's produce error:
```
[...]

PR #322, opened 2022-05-03
[cluster] disable udp port for service type LoadBalancer by default

k1rk:
with this change, if service type `LoadBalancer` and `extraArgs.opentsdbListenAddr` specified for VMinsert, by default UDP port will not be created in service spec. Because out of the box LoadBalancer service type doesn't support TCP and UDP in one service. 

However if needed, `udp: true` may be specified. this is to cover cases when feature gate `MixedProtocolLBService` enabled

addresses #319


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #332, opened 2022-05-25
chore: update sources in the charts

nvtkaszpir:
This way it adds clickable URLs in sources section per chart, thus
it is easier to find chart source.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #333, opened 2022-06-13
charts/victoria-metrics-agent: replaces endpoints role

f41gh7:
Replaces `role: endpoints` with `role: endpointslices` for kubernetes service discovery

 It should fix error `the number of targets for "role: endpoints" exceeds 1000 and will be truncated in the next k8s releases`


tenmozes (approved):
LGTM, thank you

Issue #335, opened 2022-06-21
MIssing permission to query endpointslices


AlexShemeshWix:
Hi
I can see that victoria-metrics-agent chart includes [endpointslices](https://github.com/VictoriaMetrics/helm-charts/blob/65c53ff80baa254f894aea8ec3e3de01267e1c75/charts/victoria-metrics-agent/values.yaml#L434) as role for kubernetes_sd_config sections in scrape configuration.
But ClusterRole definition in [clusterrole.yaml](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-agent/templates/clusterrole.yaml) is lacking permission for endpoinslices resource. 
When i deploy this chart it wont scrape from endpointslices because of lacking permissions. 
Please add endpointslices resource to ClusterRole definition.

AlexShemeshWix:
This is PR that suppose to fix it: [#93](https://github.com/VictoriaMetrics/helm-charts/compare/master...AlexShemeshWix:patch-1)

hagen1778:
Supposed to be fixed at https://github.com/VictoriaMetrics/helm-charts/pull/336

PR #336, opened 2022-06-21
Add endpointslices permissions to ClusterRole manifest

AlexShemeshWix:
Need to add endpointslices to list of allowed resources to ClusterRole definition in victoria-metrics-agent chart.


AlexShemeshWix:
mentioned here https://github.com/VictoriaMetrics/helm-charts/issues/335

PR #337, opened 2022-07-03
Fix victoria-metrics-agent clusterrole

AlexShemeshWix:
Endpointslices are not in base api groups. They are from discovery.k8s.io api group.
Need to be defined explicitly in clusterrole manifest.


tenmozes:
@f41gh7 @k1rk could you please take a look?

k1rk (approved):

f41gh7 (approved):
LGTM

Issue #348, opened 2022-07-22
Missing `{{- if }}` condition for vminsert-deployment helm template


jsanant:
**Issue**

When deploying the latest helm template for victoria-metrics-clustered version, there is an issue with the `vminsert-deployment.yaml` where it renders an empty `volume:` block even when vminsert is disabled. 

And this is the error we got when we deployed the clustered version using ArgoCD:

<img width="1080" alt="Screenshot 2022-07-22 at 6 37 40 PM" src="https://user-images.githubusercontent.com/10994557/180445713-54e59767-cfe3-4e5d-9053-aeb631535f86.png">

**Steps to reproduce the issue**

Disable vminsert in the `values.yaml`

```bash
$ cd charts/victoria-metrics-cluster

$ helm template .
```

Output:
```
---
# Source: victoria-metrics-cluster/templates/vminsert-deployment.yaml
volumes:
```

**Fix**

Add an if condition to check if `extraVolumes` are defined in the `values.yml` for vminsert

jsanant:
Fixed in PR #356

PR #349, opened 2022-07-22
Added check for extraVolumes

jsanant:
Fixes issue #348 

Since `vminsert` is stateless, there is no need to add any default volumes unlike other components like `vmagent`, `vmselect` or `vmstorage`.

Only add `volumes` and `volumeMounts` when they are defined in `values.yml`

### DIFF
--- a/charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml
+++ b/charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml
@@ -150,8 +150,10 @@ spec:
       affinity:
 {{ toYaml .Values.vminsert.affinity | indent 8 }}
     {{- end }}
-{{- end }}
+    {{- if .Values.vminsert.extraVolumes }}
       volumes:
         {{- with .Values.vminsert.extraVolumes }}
           {{- toYaml . | nindent 8 }}
         {{- end }}
+        {{- end }}
+{{- end }}


charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml
```diff
@@ -150,8 +150,10 @@ spec:
...
 {{ toYaml .Values.vminsert.affinity | indent 8 }}
     {{- end }}
-{{- end }}
+    {{- if .Values.vminsert.extraVolumes }}
       volumes:
         {{- with .Values.vminsert.extraVolumes }}
           {{- toYaml . | nindent 8 }}
         {{- end }}
+        {{- end }}
+{{- end }}
```
tenmozes, line 159: you add one `if` but two `end` is it correct?

tenmozes (changes requested):

charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml
jsanant: Yes, there was an `end` missing in the previous config so added that at line 159 and also the `end` at 158 is for the `if `condition.

yuriydzobak, line 159: i recommend to use like that
```suggestion
      {{- with .Values.vminsert.extraVolumes }}
      volumes:
          {{- toYaml . | nindent 8 }}
      {{- end }}
{{- end }}
```
in this case less logic and `with` is checking if var isn't empty and also fixed end of manifest [PR](https://github.com/VictoriaMetrics/helm-charts/pull/356)

jsanant: This also works, @tenmozes shall I make this change?

jsanant:
Fixed in PR #356

PR #362, opened 2022-08-18
Expose podSecurityContext also for StatefulSet

beniwtv:
When deploying with a StatefulSet, podSecurityContext was not exposed as it was with the Deployment template, making it impossible to use the Helm chart with our secured cluster.

This exposes it.


tenmozes (approved):
Thank you for contribution

PR #365, opened 2022-08-26
Correct comment on persistentVolume size: option

ivankovnatsky:
victoria-metrics-single.

Pardon me if I'm missing something, but it looks like the volume size here does not correspond to memory itself, rather a storage for the metrics persisting.


tenmozes (approved):
Thank you for contribbution

PR #369, opened 2022-09-05
Add AlertManager External URL Prefix

syepes:
This adds an additional support to the existing `alertmanager.baseURL` that allows the customization of the prefix `alertmanager.baseURLPrefix`


tenmozes (approved):
Thank you for contrinution

PR #373, opened 2022-09-06
Add clusternativeListenAddr port support in vmselect chart

okzheng:
vmselect already supported multi-level cluster setup by setting clusternativeListenAddr.


tenmozes (approved):
Thank you so much, approved

PR #374, opened 2022-09-07
Schedule operator cleanup hook onto operator nodes/affinity/tolerations

megakid:
This fixes a bug whereby running clusters with tainted nodes (e.g. Spot nodes in AKS are tainted out the box) means you cannot run the cleanup helm hook and you cannot remove the helm release from your cluster without first removing the finalizer


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #381, opened 2022-09-15
victoria-metrics-k8s-stack: add feature for select nonzero default tenant 

lutskevich:
When working with an external storage, I would like to specify a non-zero default tenant so that the correct tenant is immediately specified for all endpoints (for vmagent (remoteWrite), vmalert (remoteWrite,remoteRead,datasource) and grafana(default datasource)


tenmozes (approved):
thank you for contribution

PR #386, opened 2022-09-22
charts/victoria-metrics-gateway/README.md: fix wrong command

denisgolius:
Fix wrong grep pattern in README.md for vmgateway

**Get the pods lists by running this commands** section contain:
```console
kubectl get pods -A | grep 'vminsert\|vmselect\|vmstorage'
```

when it should be:
```console
kubectl get pods -A | grep 'vmgateway'
```


denisgolius:
> Could you please fix it here as well https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-gateway/README.md.gotmpl#L64

sure, fixed in the last commit

tenmozes (approved):
Thank you

PR #394, opened 2022-10-19
fix: add missing clusterrole rule for serviceendpoint discovery

furkansb:
VictoriaMetrics Single Server is unable to discover Kubernetes Service Endpoints with its current Cluster role rules. This PR addresses this problem by adding the necessary rule to the Cluster role.


f41gh7:
Thanks!

PR #400, opened 2022-11-05
Add missing alert rules for vmcluster

b-a-t:
As was discussed in Telegram here is the MR with missing `vmcluster` alert rules. In addition:

* removed stale `kube-apiserver.rules.yaml` file as it was superseded by newer `kube-apiserver-*.rules.yaml`
* fixed references to the alert rules in the deployment/docker, as the monolith file was split into chunks there.
* updated rules to match current ones in the deployment/docker


tenmozes (commented):
LGTM, thank you @f41gh7 please take a look as well

f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #405, opened 2022-11-16
vmcluster name override used when mapping services to ingresses

ericholt:
The ingresses created for vmcluster components (vmstorage, vmselect, vminsert) were not mapping to the correct service names when `vmcluster.name` was specified.
Also, vmagent does not remote write to the correct service name for vminsert.


tenmozes:
@k1rk could you please take a look?

tenmozes:
@k1rk could oyu please check

k1rk (approved):
@ericholt 
thanks for contribution, and sorry that i completely overlooked this for so long

PR #406, opened 2022-11-17
removed external label

bon3o:
Remove external label 'cluster' as it breaks cassandra jobs as it also has this label in its metrics. Anyway this label`s default value 'cluster-name' is useless and should be set only if needed.


tenmozes (approved):
Thank you for contribution

tenmozes:
I had to revert commit because it breaks Grafana dashboards

tenmozes:
for you case please filter our Cluster label by using metric name filter

PR #410, opened 2022-11-22
Fix configMap value in helpers

SnelsSM:
Should use .Values.server.scrape.configMap instead of .Values.configMap


tenmozes (approved):
Thank you for contribution, i will prepare release soon

PR #421, opened 2022-12-02
vmanomaly: fix config formatting, make naming consistent

zekker6:
Makes `monitoring` top-level config option as required for `vmanomaly`

Use `enabled` for consistency with another parameters.

### DIFF
--- a/charts/victoria-metrics-anomaly/values.yaml
+++ b/charts/victoria-metrics-anomaly/values.yaml
@@ -174,7 +174,7 @@ monitoring:
   # /metrics server.
   pull:
     # -- if true expose metrics on /metric endpoint
-    enable: true
+    enabled: true
     port: 8440
   push:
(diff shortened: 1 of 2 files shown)


charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -174,10 +174,10 @@ monitoring:
   # /metrics server.
   pull:
     # -- if true expose metrics on /metric endpoint
-    enable: true
+    enabled: true
```
tenmozes, line 177: this is a part of vmanomaly config :) we can do mapping and I prefer enabled

zekker6: added a mapping to leave `enable` to vmanomaly config

tenmozes (approved):

PR #422, opened 2022-12-06
vmanomaly: add annotation for pods to trigger recreating pod after config change

zekker6:
Currently, pod won't be recreated after CM change, so it will require either waiting until next restart or manual restart.
This PR adds annotation which will be changed after CM change, this will force pod recreation.


f41gh7 (approved):
LGTM

PR #423, opened 2022-12-06
PDB is still created when vmstorage.enabled is set to false

monteiro-renato:
While switching the `vmstorage.enabled` flag to false, the **Service** and **StatefulSet** are removed from the rendered manifest but not the **PDB**. Looking at all the templates, it looks like this `if` statement is missing from the PDB template.

### DIFF
--- a/charts/victoria-metrics-cluster/templates/vmstorage-pdb.yaml
+++ b/charts/victoria-metrics-cluster/templates/vmstorage-pdb.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.vmstorage.enabled -}}
 {{- if .Values.vmstorage.podDisruptionBudget.enabled }}
 {{- if .Capabilities.APIVersions.Has "policy/v1beta1" }}
 apiVersion: policy/v1beta1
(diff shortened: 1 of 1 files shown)


charts/victoria-metrics-cluster/templates/vmstorage-pdb.yaml
```diff
@@ -1,3 +1,4 @@
+{{- if .Values.vmstorage.enabled -}}
```
tenmozes, line 1: Hi there is no `{{ end }}` statement

tenmozes (changes requested):

charts/victoria-metrics-cluster/templates/vmstorage-pdb.yaml
monteiro-renato: Refactored it to look more like > https://github.com/VictoriaMetrics/helm-charts/blob/d1e0060d64cc17789a9c08b6dde98314c9cd0c42/charts/victoria-metrics-cluster/templates/vmstorage-service-monitor.yaml#L1

PR #430, opened 2022-12-09
ci: configure chart installation for testing

zekker6:
This PR adds a workflow which will run install command and verify that deployment will be successful for the given chart.
Workflow is using dynamically built matrix to validate only changed charts.
Here are several run examples:
- https://github.com/VictoriaMetrics/helm-charts/actions/runs/3658586426 - single changed chart
- https://github.com/VictoriaMetrics/helm-charts/actions/runs/3658372967 - all charts "fake" changed

### DIFF
--- a/.github/ci/chart_schema.yaml
+++ b/.github/ci/chart_schema.yaml
@@ -1,0 +1,37 @@
+name: str()
+home: str(required=False)
+version: str()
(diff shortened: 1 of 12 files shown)


f41gh7 (approved):
LGTM

PR #440, opened 2023-01-10
Fix sidecar support

hafe:
The extraContainer support was partly broken, partly missing. With these changes it is possible to inject a sidecar container into vminsert & vmselect PODs. Such extra container can terminate incoming traffic, perform authentication and proxy traffic to the "real" container.


f41gh7:
Thanks for contribution!

Issue #354, opened 2022-08-01
Cluster-scoped resources have namespace fields


hoboroten:
Some cluster-level resources in charts have namespace field in them. For example, ClusterRole and ClusterRoleBinding in victoria-metrics-agent and victoria-metrics-cluster charts. Helm and kubectl usually ignores this field, but when i try to apply these charts with terraform kustomization provider (it has built-in schema checks) - it gives in error.

PR #442, opened 2023-01-11
Remove namespaces from cluster-scoped objects

hoboroten:
Fixes [#354](https://github.com/VictoriaMetrics/helm-charts/issues/354) 
Cluster-scoped resources need no namespace field. Some tools, like terraform-kustomization-provider, will throw validation errors if these fields exist


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #443, opened 2023-01-12
Bug in vmagent ConfigurationReloadFailure alert causes false alerting with vmsingle

brokenjacobs:
vm_promscrape_config_last_reload_successful metric is emitted by vmagent and the victoriametrics pod. This update causes the alert to only fire if vmagent fails to reload.


f41gh7:
I believe it's an issue at VictoriaMetrics side. It must return vm_promscrape_config_last_reload_successful = 1 if there is no config file at all.

brokenjacobs:
> I believe it's an issue at VictoriaMetrics side. It must return vm_promscrape_config_last_reload_successful = 1 if there is no config file at all.

It could be but older releases also have the metric.

brokenjacobs:
I'd also point out that this alert is new, and has been alerting off vmsingle since it was added to the chart.

notz:
After installing the latest helm chart  (0.14.10) the alert is gone on my installation

brokenjacobs:
It was hard to tell because my alerting is broken at the moment writing to vmagent in the new release, but you are correct this is no longer firing. Closing pr.

PR #453, opened 2023-01-24
fix: Default scraping for pod and node metrics using K8s-VM-Stack

tejavegesna:
Default Chart should be able to work out of box. Instead of looking how to inject metrics from node exporter and kube-state-metrics. Without specifying these vmagent doesn't scrape those metrics


k1rk:
hi @tejavegesna. 
This annotations are not used by VictoriaMetrics stack by default. And VMAgent correctly scrapes metrics from node exporter and kube-state-metrics because VMNodeScrape (ref: https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-k8s-stack/templates/servicemonitors/nodeexporter.yaml) and VMServiceScrape (ref: https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-k8s-stack/templates/servicemonitors/kube-state-metrics.yaml) objects are created in cluster

can you tell a bit more what is not working for you?

PR #456, opened 2023-01-26
fix vmReadEndpoint and vmWriteEndpoint

bon3o:
When vmcluster.spec.vmselect.extraArgs.http.pathPrefix and vmcluster.spec.vminsert.extraArgs.http.pathPrefix is not defined(as it is by default) address becomes invalid as it adds extra slash after port number.  
i.e: http://vmselect-k8s-stack.default.svc:8481//select/0/prometheus


k1rk (approved):
@bon3o great catch, thank you

PR #463, opened 2023-02-08
update default retentionPeriod to 14d

jgillich:
If the time unit is not specified, it defaults to months. So the current default retention period is 14 months, which *probably* wasn't intended.


k1rk:
@jgillich thank you for contribution. 
however this was intended. and also changing this default may be very damaging to lots of installations out there

closing this PR

PR #465, opened 2023-02-17
Add namespaced flag for vmagent

jdmaguire:
For a certain use case, I need to run victoria-metrics-agent in a single namespace w/o clusterroles/clusterrolebinding.

The chart as is does give me the option of installing without the ClusterRole/ClusterRoleBinding via the .Values.rbac.create. It would be useful for me if I could tell it to created a namespaced RBAC.

I am aware this is different than some of the other helm charts here. For example, I do see that the `victoria-metrics-alert` chart used the same flag and behaves as following (if .Values.rbac.create is true):
- Create the Role/RoleBinding
- If not namespaced, also create the ClusterRole/ClusterRoleBinding

I went with the approach in this PR as it is the least code changes and the most maintainable (as it doesn't add an additional two files/CRs).


tenmozes (approved):
Thank you for your contribution, LGTM

PR #469, opened 2023-02-21
alert: add support of auth settings for notifier

hagen1778:


tenmozes (approved):
LGTM, thank you for contribution

charts/victoria-metrics-alert/values.yaml
```diff
@@ -107,6 +107,16 @@ server:
   notifier:
     alertmanager:
       url: ""
+      # -- Basic auth for alertmanager
+      basicAuth:
```
f41gh7, line 111: I don't think, that we should expose sensitive data with plain text format.

 I'd recommend to create kubernetes Secret:

```yaml
metadata:
 name: notifier-access
kind: Secret
stringData:
   username: u
   password: p
```

 and use it as reference for extra args:

```yaml
# values.yaml
  env:
     - name: VM_notifier.basicAuth.password
       valueFrom:
         secretKeyRef:
           name: notifier-access
           key: password
     - name: VM_notifier.basicAuth.username
       valueFrom:
         secretKeyRef:
           name: notifier-access
           username: password
```

tenmozes: we do that for every other chart - so if we want to change we need to change this everywhere

PR #477, opened 2023-02-28
[victoria-metrics-alert] Bumped default alertmanager version to v0.25.0

shanoor:
Bumped alertmanager version to v0.25.0 because the default version is more than obsolete (v0.20.0 from 2019).


tenmozes (approved):
LGTM, thank you

Issue #475, opened 2023-02-27
Charts appear to lack any TLS cert-related config options


XANi:
`vmalert` for example have a whole slew of `remoteWrite.tls*` options yet there doesn't seem to be a way to pass it via helm config to fill the k8s config. In our particular case I have internal CA which all internal services use and want to pass it a key

AndrewChubatiuk:
Hey @XANi. We're usually adding new properties to a spec which simplify setup a lot, like `streamAggrConfig`, which generates config map with a data from parameter, provides validation of this config and adds respective volumemount configs and cmd flags. We do not add to properties flags, which are just straightforwardly converted into cmd args only, like in case of `remoteWrite.tls*`. You can add these flags using `extraArgs`:
```yaml
vmalert:
  spec:
     extraArgs:
       remoteWrite.tlsCAFile: "ca1.pem,ca2.pem"  # means use ca1.pem for a first remote write and ca2.pem for a second one
```

PR #478, opened 2023-02-28
Add option to specify -remoteWrite.tlsInsecureSkipVerify per URL

XANi:
Currently there is no option to add a custom CA nor to skip TLS verification for the remoteWrite targets, which might be needed for internal CAs (or just test instances). This adds option to set -remoteWrite.tlsInsecureSkipVerify per URL (not sure about variable naming here tbh.), related to #475  

Ideally I'd like to just pass a CA cert in the config YAML with everything else (or alternatively have option to point to existing secret) but, uh, that's first time I use Helm and I'm not exactly sure how to even do that in it.


f41gh7:
Hello, thanks for PR.

But it's better to use `extraArgs` for this case:

```yaml
extraArgs:
  remoteWrite.tlsInsecureSkipVerify: "true,false"
```

 Only viable flag settings should be added to the values file directly.

XANi:
@f41gh7 Why it is "better" ? Every other TLS option gets its own option

PR #486, opened 2023-03-09
fix label selector for argocd

bon3o:
When deployed via argocd servicescrape is selecting wrong labels


tenmozes:
@zekker6 please take a look

zekker6 (approved):
LGTM
@bon3o Thanks for contribution!

PR #488, opened 2023-03-11
fix custom vmstorage node address render in vmselect and vminsert args

okzheng:
When specified vmstorage node address in extraArgs, multiple vmstorage nodes would be deduplicated because of same key `storageNode`. 
If adding these address in a list, it could resolve the problem.


f41gh7:
When using extraArgs - provide comma-separated string  of storage nodes for vmstorageNode flag.

 E.g.

```
extraArgs:
  storangeNode: vmstorage-1,vmstorage-2,vmstorage-3

PR #489, opened 2023-03-13
[victoria-metrics-k8s-stack] add extraArgs options for operator

z-anshun:
When I'm using `victoria-metrics-k8s-stack`, I want to set the args of the operator, but I can't find it in the chart. Then I found that `operator` supports extraArgs option on v0.30.4(https://github.com/VictoriaMetrics/helm-charts/commit/26450278f9853cbd5ff48a81d8880c10ea420e36), so probably `victoria-metrics-k8s-stack` should support it too.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/values.yaml
+++ b/charts/victoria-metrics-k8s-stack/values.yaml
@@ -22,6 +22,7 @@ victoria-metrics-operator:
   operator:
     # -- By default, operator converts prometheus-operator objects.
     disable_prometheus_converter: true
+  extraArgs: {}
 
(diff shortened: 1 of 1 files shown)


k1rk:
hi @z-anshun . you don't need to set it in parent chart if it is supported in dependent chart. 

you can just provide it in your values value and it will work out of the box

z-anshun:
> hi @z-anshun . you don't need to set it in parent chart if it is supported in dependent chart.
> 
> you can just provide it in your values value and it will work out of the box

thanks, i was wrong.

PR #490, opened 2023-03-13
[victoria-metrics-operator]:fix extraArgs options

z-anshun:
I found some bugs when I used `extraArgs` of operators. The chart resolves `extraArgs` as shorthand and `operator` via `pflag` so that `--` and `-` have different meanings.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #493, opened 2023-03-15
[victoria-metrics-k8s-stack]:adding tpl upstream for external labels …

z-anshun:
hi,
When I use `prom` to add external labels, it supports tpl upstream. But the chart doesn't support tpl upstream to set `-externalLabels` parameter  of `vmalert`.


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #498, opened 2023-03-21
[victoria-metrics-cluster] add 'appProtocol' and 'enableHttp2' options

AlphaStyle:
When Istio is configured with PeerAuthentication-[policy](https://istio.io/latest/docs/concepts/security/#authentication-policies) mode set to STRICT, mutual TLS is enforced by default. Istio is relying on [protocol selection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) to proxy traffic. This pull-request will add the optional `appProtocol`-filed to the service file.

This pull-request will also add the optional `enableHttp2` ([prometheus-operator](https://prometheus-operator.dev/docs/operator/api/#monitoring.coreos.com/v1.ServiceMonitorSpec)) filed to the serviceMonitor-file. To make the `enableHttp2` field optional the Helm [kindIs helper function](https://helm.sh/docs/chart_template_guide/function_list/#kind-functions) is used.

```yaml
{{- if (kindIs "bool" .Values.vmselect.serviceMonitor.enableHttp2) }}
enableHttp2: {{ .Values.vmselect.serviceMonitor.enableHttp2 }}
{{- end }}
```

This snippet will only set the `enableHttp2` filed if `serviceMonitor.vmselect.enableHttp2` is either `true` or `false`. If it is empty or set to something else, it will be ignored.

Feedback and improvements are welcome.

### DIFF
--- a/charts/victoria-metrics-cluster/Chart.yaml
+++ b/charts/victoria-metrics-cluster/Chart.yaml
@@ -2,6 +2,6 @@ apiVersion: v1
 appVersion: 1.90.0
 description: Victoria Metrics Cluster version - high-performance, cost-effective and scalable TSDB, long-term remote storage for Prometheus
 name: victoria-metrics-cluster
-version: 0.9.61
+version: 0.9.62
 sources:
 - https://github.com/VictoriaMetrics/helm-charts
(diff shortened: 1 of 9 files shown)


AlphaStyle:
Did a rebase to resolve conflicting files.

Issue #505, opened 2023-03-28
[vm-operator] PSP isn't automagically skipped with ArgoCD on k8s >= 1.25


b-a-t:
We deploy the `victoria-metrics-k8s-stack` Helm chart via ArgoCD and finally tried newer Kubernetes 1.25.

Unfortunately, deployment fails with the error:

> victoria-metrics-operator: the server could not find the requested resource: policy/v1betav1/PodSecurityPolicy

That isn't a surprise, as K8s 1.25 [deprecated](https://kubernetes.io/blog/2022/08/04/upcoming-changes-in-kubernetes-1-25/#podsecuritypolicy-removal) in favor of [PodSecurityAdmission](https://kubernetes.io/docs/tasks/configure-pod-container/migrate-from-psp/)

This came as a surprise, as PSP resource in the `vm-operator` is surrounded by the guardian code, which should handle the situation properly:
```yaml
{{- if and .Values.rbac.pspEnabled (.Capabilities.APIVersions.Has "policy/v1beta1") }}
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
...
{{- end }}
```
Except that it doesn't work with ArgoCD. 

This is again not a surprise, as ArgoCD uses `helm template` to generate manifests and then applies them by itself, rather than doing `helm install/upgrade`(I'm speculating here, but I think that's the case).

The `.Capabilities.APIVersions.Has` doesn't work properly with such usage and even despite supplying `--api-versions` parameters to the `helm` call the result of the above evaluation is `true` somehow:

```
[...]

PR #506, opened 2023-03-29
charts: fix k8s v1.25 compatibility for PSP

zekker6:
- adds k8s v1.25 into testing matrix
- replaces `(.Capabilities.APIVersions.Has "policy/v1beta1")` calls with `(.Capabilities.APIVersions.Has "policy/v1beta1/PodSecurityPolicy")` to make it compatible with ArgoCD

See #505 for more details.


k1rk (approved):

tenmozes (approved):

pcperera:
How can we remove the following psp as they are not created by the operator ?

image

pcperera:
How can we remove the following psp as they are not created by the operator ?

image

PR #507, opened 2023-03-30
charts/vmalert: note that chart does not allow to setup HA Alertmanager

zekker6:
- remove `alertmanger.replicaCount`
- add a note with workarounds for HA setup

Existing `alertmanager.replicaCount` would only set multiple instance of pod to be running, this does not provide a real HA configuration as users assume when using this. Also, since it uses `Deployment` using multiple replicas will fail to deploy as single PVC cannot be attached to multiple pods for `ReadWriteOnce`. For `ReadWriteMany` it might lead to data corruption since multiple Alertmanager instance might modify the same data.


f41gh7 (approved):
LGTM

PR #508, opened 2023-03-30
Fix double "rules" on no-PodSecurityPolicy path

jackie-ob:
Discovered when trying to use this chart on EKS k8s v1.25 (PSP API removed).  In our case, we are using FluxCD HelmRelease x helm-controller to perform the install.  Possibly helm CLI will ignore the dupe (not sure).


tenmozes (approved):
Thank you for contribution

PR #510, opened 2023-04-06
charts/vmauth: use `.Values.configMap` value for config volume spec

zekker6:
Currently, this setting is only disabling creating secret which makes vmauth config volume reference invalid and failing deployment.

### DIFF
--- a/charts/victoria-metrics-auth/values.yaml
+++ b/charts/victoria-metrics-auth/values.yaml
@@ -184,8 +184,8 @@ serviceMonitor:
 #    tlsConfig:
 #      insecureSkipVerify: true
 
-# -- Use existing configmap if specified otherwise .config values will be used. Ref: https://victoriametrics.github.io/vmauth.html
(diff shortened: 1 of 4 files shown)


charts/victoria-metrics-auth/values.yaml
```diff
@@ -184,8 +184,8 @@ serviceMonitor:
 #    tlsConfig:
 #      insecureSkipVerify: true
 
-# -- Use existing configmap if specified otherwise .config values will be used. Ref: https://victoriametrics.github.io/vmauth.html
-configMap: ""
+# -- Use existing secret if specified otherwise .config values will be used. Ref: https://victoriametrics.github.io/vmauth.html
```
f41gh7, line 187: Could you mention, that `Secret` must contain key `auth.yml` with valid vmauth configuration?

f41gh7 (approved):
LGTM

PR #512, opened 2023-04-10
Use non-hardcoded `app.kubernetes.io/name` match label in victoria-metrics-k8s-stack VMServiceScrapes

sapslaj:
The Grafana, kube-state-metrics, and node_exporter subcharts derive the `app.kubernetes.io/name` label in their Service manifests from a helper, usually `{chartname}.name`. If `nameOverride` is given (for whatever reason) then the output of that helper will not match the label value for the VMServiceScrape label selector, which causes it to not work.

Here's a minimum `values.yaml` to reproduce the issue:

```
prometheus-node-exporter:
  nameOverride: vm-node-exporter
```

With the current version of the victoria-metrics-k8s-stack the above will cause node_exporter metrics to not work. This PR will fix that. It executes the helper in the context of the subchart so all of the `.Values` and such lookups it does will work as intended.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/templates/servicemonitors/grafana.yaml
+++ b/charts/victoria-metrics-k8s-stack/templates/servicemonitors/grafana.yaml
@@ -8,7 +8,7 @@ metadata:
 spec:
   selector:
     matchLabels:
-      app.kubernetes.io/name: grafana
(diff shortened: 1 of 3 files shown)


f41gh7 (approved):
LGTM

PR #516, opened 2023-04-19
[victoria-metrics-k8s-stack]: add tpl to vmagent, vmalert, vmcluster and vmsingle spec

nvanheuverzwijn:
This is adding the `tpl` function to all the spec of the operator. It basically allows us to use helm template language inside the `spec` section of `vmagent`, `vmalert`, `vmcluster` and `vmsingle`. For example, I would prefer to set one global variable for the cluster name (`global.clusterName` for example) and use that as an external labels for both vmagent and vmalert `vmagent.spec.externalLabels.cluster: {{ .Values.global.clusterName }}`.

This is used extensively in kube-prometheus-stack chart, although, it is used in more precise places than what I have here.

Let me know your thoughts.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
+++ b/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
@@ -132,15 +132,6 @@ configMaps:
 {{- end -}}
 
 {{/*
-VMAlert externalLabels
-*/}}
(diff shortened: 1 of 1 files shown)


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

PR #518, opened 2023-04-21
Anomaly: change defaults

dima-vm:
Made them more realistic:

- Bump version to 0.3.2
- Change defaults
- Add samplingPeriod param


dima-vm:
ping @tenmozes

charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -173,17 +173,15 @@ queries: {}
...
 monitoring:
   # /metrics server.
   pull:
     # -- if true expose metrics on /metric endpoint
-    enabled: true
     port: 8440
   push:
     # -- push metrics on prometheus format if defined
-    endpoint_url: ""
+    url: ""
```
f41gh7, line 185: NOTE, it looks like breaking change. You have to mention it and provide migration path.

charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -173,17 +173,15 @@ queries: {}
...
     extra_labels:
-      job: "vmanomaly"
-  
+      job: "vmanomaly-push"
```
f41gh7, line 187: It may break filters for grafana dashboards, FYI.

PR #521, opened 2023-04-24
victoria-metrics-operator chart: split securityContext

timp87:
Currently victoria-metrics operator chart (0.21.0) has securityContext option in values.yaml
It sets securityContext for the pod only. But there is no way to set securityContext for the container.
This matter since those contexts have different configuration options (some of them interfere though):
- [container security context](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#securitycontext-v1-core)
- [pod security context](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#podsecuritycontext-v1-core)

For example, there is a really crucial readOnlyRootFilesystem option. It can be set only on container security context level. No way to set it on pod security context level.

Divide securityContext into two parts: pod and container. This breaks values.yaml backward compatibility with previous chart versions. So if merged then a proper note should be put into corresponding release notes.

### DIFF
--- a/charts/victoria-metrics-operator/templates/deployment.yaml
+++ b/charts/victoria-metrics-operator/templates/deployment.yaml
@@ -38,9 +38,9 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
     {{- end }}
-{{- with .Values.securityContext }}
+{{- with .Values.podSecurityContext }}
       securityContext:
-{{ toYaml . | nindent 8 }}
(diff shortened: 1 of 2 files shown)


timp87:
I started this PR to be able to set readOnlyRootFilesystem to true for the operator container.
I think this option may be hardcoded into the spec as there is no sense to make fs writable.

So another way to achieve it would be just hardcode readOnlyRootFilesystem to the spec and leave values.yaml untouched.

What do you think?

PR #522, opened 2023-04-27
Fix typo in values.yaml: VMSingle --> VMCluster

imroc:


k1rk (approved):
@imroc thanks for contribution

PR #524, opened 2023-04-27
[victoria-metrics-k8s-stack] Bump chart dependencies

D3luxee:
Hi,
I've recently switched from kube-prometheus-stack to victoria-metrics-k8s-stack and noticed that the sub charts are older.

In this MR i would propose to bump the sub charts to the same versions as used in kube-prometheus-stack.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/Chart.yaml
+++ b/charts/victoria-metrics-k8s-stack/Chart.yaml
@@ -13,14 +13,14 @@ dependencies:
     repository: https://victoriametrics.github.io/helm-charts
     condition: victoria-metrics-operator.enabled
   - name: kube-state-metrics
-    version: "4.24.*"
+    version: "5.5.*"
     repository: https://prometheus-community.github.io/helm-charts
     condition: kube-state-metrics.enabled
   - name: prometheus-node-exporter
-    version: "4.7.*"
+    version: "4.16.*"
     repository: https://prometheus-community.github.io/helm-charts
     condition: prometheus-node-exporter.enabled
   - name: grafana
-    version: "6.44.*"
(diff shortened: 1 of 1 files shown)


f41gh7:
It must be tested for backward compatibility. As far as I remember, node-exporter breaks at regular basics.

D3luxee:
Hi,
do you mean backward compatibility in the chart or the applications, or both?
We were using kube-prometheus-stack a long time and i do not remember breaking changes in any of the components.
But i totally understand that this must be validated.

onedr0p:
Any update on this?

Haleygo:
@D3luxee @onedr0p Hello! I think there are several things need to be done:
1. Test if the full chart can be installed successfully. So does upgrade, there could be breaking changes for crds or other resources, those changes need to be clarified in [upgrade guide](https://github.com/VictoriaMetrics/helm-charts/blob/cb7245813752f980540d773eea0c9be031a8e104/charts/victoria-metrics-k8s-stack/README.md?plain=1#L174).
2. There are some deprecated metrics in new version of ksm or node-exporter, listed in their changelogs. We need to check the possible usage of those deprecated in dashboards or rules, change or update them if there is.

kube-prometheus-stack chart's dependencies upgrade looks [casual](https://github.com/prometheus-community/helm-charts/pull/3473), but we are facing big version bump here)

sydorovdmytro:
Do you know if there is any progress? Grafana Helm chart is using outdated (removed) API versions for HPA.

Haleygo:
Should implemented by https://github.com/VictoriaMetrics/helm-charts/pull/657, close this as dup, thanks)

PR #527, opened 2023-05-02
add security context to vmbackupmanger

tenmozes:


f41gh7 (approved):
LGTM

zekker6 (approved):

ptimofee:
Sorry, but this change is not quite right.
I have added podSecurityContext into container's securityContext. And as far as I can see it's already been there, you've just completed it.
Those are two different things. Let me prepare a proper fix.

PR #530, opened 2023-05-03
Update link to kubernetes.io

nemobis:
The page was redirected in October 2022 and now returns an HTTP 404 error.


f41gh7 (approved):
LGTM

PR #532, opened 2023-05-04
added chart releaser workflow to CI/CD

Amper:


.github/workflows/releaser.yaml
```diff
@@ -0,0 +1,42 @@
...
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+
+      - name: Prepare chart
+        run: |
+          find charts/*/Chart.yaml -exec sh -c 'sed "s/^version: .*$/version: ${{ github.ref_name }}/g" $1 1<> $1' -- {} \;
+          make package
+          make index
```
Haleygo, line 37: I see index stage being commented out, do we need it here?
https://github.com/VictoriaMetrics/helm-charts/blob/82b09f658fbd6fe03b067fb28a6836b4cfd8927b/Makefile#L107-L109

PR #535, opened 2023-05-11
charts/vmauth: add a note about secret format

zekker6:
- added `notifiers` section which will be merged with `notifier` to create a list of notifiers
- update version of helm used for CI and linting


k1rk (approved):
good job 👍 

nit: should we maybe deprecate eventually `notifier` in values in favor of `notifiers` ?

PR #540, opened 2023-05-19
change path if 'http.pathPrefix' exists in extraArgs

krakazyabra:


zekker6 (approved):

k1rk (approved):
@krakazyabra thanks for contribution

PR #544, opened 2023-05-24
do not render empty pod security context

tenmozes:


charts/victoria-metrics-cluster/templates/vmstorage-statefulset.yaml
```diff
@@ -50,8 +50,10 @@ spec:
           {{- if .Values.vmstorage.containerWorkingDir }}
           workingDir: {{ .Values.vmstorage.containerWorkingDir }}
           {{- end }}
+          {{- with .Values.vmstorage.podSecurityContext }}
           securityContext:
-            {{- toYaml .Values.vmstorage.podSecurityContext | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
+          {{-end }}
```
f41gh7, line 56: Typo -> {{-end }} -> {{- end }}

f41gh7:
Fixed by https://github.com/VictoriaMetrics/helm-charts/pull/545

PR #545, opened 2023-05-24
charts/cluster: do not render empty securityContext

zekker6:
Rendering empty securityContext will force deployemtns and STS to be re-created.


f41gh7 (approved):
LGTM

PR #556, opened 2023-06-06
Update to new HPA apiVersion, and allow specifying in values

### LINKED ISSUE #514: HorizontalPodAutoscaler is using autoscaling/v2beta when autoscaling/v2 is out and in use
As a person that is running on Kubernetes 1.26.3 I am having issues when enabling the helm-charts horizontalPodAutoscaler.

Namely it tells me I don't have appropriate CRD, this is because `autoscaling/v2beta2` is deprecated in k8s `1.23` and removed presumingly in 1.26? (not 100% sure, maybe it was 1.25). If backward compatibility is important here I guess adding option to override the version in the values.yaml override file might be an option? Either way using `autoscaling/v2beta2` will only become problematic to more people as time goes on and needs replacing sooner or later.

ellie:
1. Upgrade to new apiVersion. autoscaling/v2beta2 is no longer supported as of 1.26, and has been deprecated for a few versions now
2. Allow specifying the apiVersion in values. This is a fairly common pattern and will allow users to workaround this change.

Resolves #514

### DIFF
--- a/charts/victoria-metrics-cluster/templates/vminsert-hpa.yaml
+++ b/charts/victoria-metrics-cluster/templates/vminsert-hpa.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.vminsert.horizontalPodAutoscaler.enabled }}
+{{- if .Capabilities.APIVersions.Has "autoscaling/v2" -}}
+apiVersion: autoscaling/v2
+{{- else}}
 apiVersion: autoscaling/v2beta2
+{{- end}}
 kind: HorizontalPodAutoscaler
 metadata:
   labels:
--- a/charts/victoria-metrics-cluster/templates/vmselect-hpa.yaml
+++ b/charts/victoria-metrics-cluster/templates/vmselect-hpa.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.vmselect.horizontalPodAutoscaler.enabled }}
+{{- if .Capabilities.APIVersions.Has "autoscaling/v2" -}}
+apiVersion: autoscaling/v2
+{{- else}}
 apiVersion: autoscaling/v2beta2
+{{- end}}
 kind: HorizontalPodAutoscaler
 metadata:
   labels:


k1rk (changes requested):
thanks for contribution @ellie 

instead of implementing this as parameter could you please use `.Capabilities.APIVersions.Has` check in helm template? this will be simpler for users of chart

ellie:
Sure! Just to point out though, if using `helm template`, the version will be fetched from just the local client, vs helm install/upgrade using the remote cluster version. Probably fine, but could cause confusion for some users

k1rk (approved):

tenmozes:
@ellie thanks for contribution

tenmozes:
the feature/fix should be available in 0.9.72 version

atropos112:
> thanks for contribution @ellie 
> 
> instead of implementing this as parameter could you please use `.Capabilities.APIVersions.Has` check in helm template? this will be simpler for users of chart 

This is going to bite with ArgoCD, i would argue have the capabilities api version unless overwritten by variable not just that.

b-a-t:
> > thanks for contribution @ellie
> > instead of implementing this as parameter could you please use `.Capabilities.APIVersions.Has` check in helm template? this will be simpler for users of chart
> 
> This is going to bite with ArgoCD, i would argue have the capabilities api version unless overwritten by variable not just that.

@ellie righteously pointed out the issues with `helm template` usage, which is, basically what ArgoCD is using. I remember we addressed a similar problem with it in PSP detection in the chart with ArgoCD.

From other side I see(on older cluster) that ArgoCD supplies:

```
helm template . --name-template vm-cluster --namespace monitoring --kube-version 1.24 ... --api-versions autoscaling/v1 --api-versions autoscaling/v1/HorizontalPodAutoscaler --api-versions autoscaling/v2 --api-versions autoscaling/v2/HorizontalPodAutoscaler --api-versions autoscaling/v2beta1 --api-versions autoscaling/v2beta1/HorizontalPodAutoscaler --api-versions autoscaling/v2beta2 --api-versions autoscaling/v2beta2/HorizontalPodAutoscaler ...
```

And
```
helm template . --name-template dev-vm --namespace monitoring --kube-version 1.25 ... --api-versions autoscaling/v1 --api-versions autoscaling/v1/HorizontalPodAutoscaler --api-versions autoscaling/v2 --api-versions autoscaling/v2/HorizontalPodAutoscaler --api-versions autoscaling/v2beta2 --api-versions autoscaling/v2beta2/HorizontalPodAutoscaler ...
```

So, that conditioning may work if matching is done on exact capability paths...

k1rk:
Argocd supplies capabilities to helm template based on data gathered from your cluster, so it should be fine with argocd

PR #560, opened 2023-06-09
charts/victoria-logs-single: add new chart

zekker6:


charts/victoria-logs-single/values.yaml
```diff
@@ -0,0 +1,265 @@
+# Default values for victoria-logs.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+rbac:
+  # Note: The PSP will only be deployed, if Kubernetes (<1.25) supports the resource.
+  pspEnabled: true
```
tenmozes, line 7: Let's remove that, i think we don't need to add PSP support for the new chart

charts/victoria-logs-single/values.yaml
```diff
@@ -0,0 +1,265 @@
...
+  startupProbe: {}
+    #tcpSocket:
+    #  port: http
+    #failureThreshold: 30
+    #periodSeconds: 15
+    #successThreshold: 1
+    #timeoutSeconds: 5
+
+  # -- Security context to be added to server pods
+  securityContext: {}
```
tenmozes, line 176: what do you adding the default security context and specify it as NON-ROOT?

hack/helm/victoria-logs-single.yaml
```diff
@@ -0,0 +1,2 @@
+podDisruptionBudget:
```
tenmozes, line 1: should we remove helm folder?

tenmozes (changes requested):

charts/victoria-logs-single/values.yaml
zekker6: Added sane dafaults here.

hack/helm/victoria-logs-single.yaml
zekker6: No, files in helm folder are used in CI for running testing.  Files in root of `hack/` are used only to run linter.

PR #562, opened 2023-06-20
charts/victoria-metrics-anomaly: make monitoring config more configurable

zekker6:
Update configuration to template `pull` and `push` sections only in case configuration for these is defined. Fixes compatibility with latest vmanomaly config version.


zekker6:
@k1rk @f41gh7 PTAL

f41gh7 (approved):
LGTM

PR #568, opened 2023-06-23
Fix wrong condition on fluent-bit dependency

AlexisDucastel:
The condition of the fluent-bit dependy contains a typo that will deploy fluent-bit even if disabled in values.yaml.
As the values is correct according to the dependency name, I've only fixed the dependency condition for minimal impact


zekker6 (approved):
LGTM
Thanks for the fix!

PR #573, opened 2023-07-06
Add params configuration option for default rules

ntaylor1781:
Added a default empty value for defaultRules.params. Updated hack/sync_rules.py to include the params options when set. This will allow users to configure options like extra_label.


Haleygo:
@ntaylor1781 Thanks for the implement.
But I think this sync script is meant for adding obligatory and standard fields for vmrules when release, not sure about adding optional parameters here since it's not a chart for users. You can modify the script locally and generate your rules if you want.
wdyt @Amper @f41gh7

charts/victoria-metrics-k8s-stack/templates/rules/vm-health.yaml
```diff
@@ -58,8 +64,8 @@ spec:
     - alert: TooHighMemoryUsage
       annotations:
         description: Too high memory usage may result into multiple issues such as OOMs or degraded performance. Consider to either increase available memory or decrease the load on the process.
-        summary: It is more than 90% of memory used by "{{`{{`}} $labels.job {{`}}`}}"("{{`{{`}} $labels.instance {{`}}`}}") during the last 5m
-      expr: (process_resident_memory_anon_bytes / vm_available_memory_bytes) > 0.9
+        summary: It is more than 80% of memory used by "{{`{{`}} $labels.job {{`}}`}}"("{{`{{`}} $labels.instance {{`}}`}}") during the last 5m
```
Haleygo, line 67: Is the original 90% on purpose? Below [TooHighCPUUsage](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-k8s-stack/templates/rules/vm-health.yaml#L69) also use it as threshold. @f41gh7 @k1rk

ntaylor1781: The updated rules files were generated by running the sync_rules.py after my changes. This change was pulled from https://github.com/VictoriaMetrics/VictoriaMetrics/blob/master/deployment/docker/alerts-health.yml

ntaylor1781:
If that is the desire for the script that is fine. I feel like it would be easier to write a fresh script, than modify this one (or the upstream Prometheus version), than try and convert this one. There is a lot of helm templating that this inserts, which would not be helpful if I am generating and trying to apply separately from helm. 

As a final comment about the script, it does have logic to add additional labels to rules that can be specified in the values file. That's one of the reasons I believed that it made sense to add the extra customization there.

Issue #580, opened 2023-07-11
`suppresStorageFQDNsRender` is misspelled in Helm Charts.


perfectra1n:
Hey VictoriaMetrics,

Thanks for the great product. Just wanted to reach out and let you guys know that `suppresStorageFQDNsRender` is misspelled and it should be `suppressStorageFQDNsRender ` unless I'm losing my sanity lol. This cost me a decent amount of time since I didn't copy and paste the key into my Helm values lol.

[It's in a few places](https://github.com/search?q=repo%3AVictoriaMetrics%2Fhelm-charts+suppresStorageFQDNsRender+&type=code)

https://github.com/VictoriaMetrics/helm-charts/blob/1b20843c5c287e7811bd8062b26e2826eb73aa0a/charts/victoria-metrics-cluster/templates/vmselect-statefulset.yaml#L54

https://github.com/VictoriaMetrics/helm-charts/blob/1b20843c5c287e7811bd8062b26e2826eb73aa0a/charts/victoria-metrics-cluster/values.yaml#L55

tenmozes:
hi @perfectra1n should be fixed in https://github.com/VictoriaMetrics/helm-charts/pull/581

perfectra1n:
oooooooh, that's gonna break some deployments for sure, just FYI if they're relying on that value

PR #581, opened 2023-07-11
fix typo in suppressStorageFQDNsRender, address #580

tenmozes:


zekker6 (approved):

f41gh7:
It breaks backward compatibility,  I suggest to define this variable at _helpers and set it from both variants.

tenmozes:
@f41gh7 it doesn't break backward compatibility, please check the "check" in if statement

f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #583, opened 2023-07-12
feat(vmagent): Add support for topologySpreadConstraints field

mkilchhofer:
The field `topologySpreadConstraints` has a wide adoption in the community helm charts.
We at @swisspost use ensure that pods are spread across all our AZ's.

While this can also be achieved via affinity/podAntiAffinity, `topologySpreadConstraints` offers more options.


k1rk (approved):
thanks for contribution @mkilchhofer

Issue #455, opened 2023-01-25
podSecurityContext and securityContext in values.yaml used interchangeably across helm charts


srihas:
The standard convention of using podSecurityContext and securityContext from values.yaml in deployment/statefulset spec seems to be as follows:

      securityContext:
        {{- toYaml .Values.podSecurityContext | nindent 8 }}
      containers:
        - name: {{ .Chart.Name }}
          securityContext:
            {{- toYaml .Values.securityContext | nindent 12 }}

While, the same is not followed in many cases - vmalert, vmcluster, alertmanager etc.
for e.g. in file:
https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-alert/templates/alertmanager-deployment.yaml

      {{- with .Values.alertmanager.securityContext }}
      securityContext:
        {{- toYaml .| nindent 8 }}
      {{- end }}
      containers:
        - name: {{ template "vmalert.name" . }}-alertmanager
          securityContext:
            {{- toYaml .Values.alertmanager.podSecurityContext | nindent 12 }}

While in the case the vmagent, it is rightly used:
https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-agent/templates/deployment.yaml

      securityContext:
        {{- toYaml .Values.podSecurityContext | nindent 8 }}
      containers:
        - name: {{ .Chart.Name }}
          securityContext:
            {{- toYaml .Values.securityContext | nindent 12 }}

Haleygo:
@srihas Thanks for the report!
Feel free to pull a request if you like)

PR #592, opened 2023-07-21
fix misused securityContext and podSecurityContext

Haleygo:
fix https://github.com/VictoriaMetrics/helm-charts/issues/455
`securityContext` should be used for containers, `podSecurityContext` should be used for pod


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

Issue #548, opened 2023-05-31
vm operator not delete completely


Tri0L:
## Install:
```
helm install vmoperator vm/victoria-metrics-operator -n victoria-metrics
```

## kubectl:
```
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMCluster
metadata:
  name: main-cluster
spec:
  retentionPeriod: "12"
  vmstorage:
    replicaCount: 1
  vmselect:
    replicaCount: 1
  vminsert:
    replicaCount: 1
---
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAgent
metadata:
  name: vmagent
spec:
  serviceScrapeNamespaceSelector: {}
  podScrapeNamespaceSelector: {}
  podScrapeSelector: {}
  serviceScrapeSelector: {}
  nodeScrapeSelector: {}
  nodeScrapeNamespaceSelector: {}
  staticScrapeSelector: {}
  staticScrapeNamespaceSelector: {}
  replicaCount: 1
  remoteWrite:
    - url: "http://vminsert-main-cluster.victoria-metrics.svc.cluster.local:8480/insert/0/prometheus/api/v1/write"
```

## Uninstall:
```
 % helm uninstall vmoperator -n victoria-metrics               
W0601 00:38:12.060009   37573 warnings.go:70] policy/v1beta1 PodSecurityPolicy is deprecated in v1.21+, unavailable in v1.25+
release "vmoperator" uninstalled
```

## Check.
A lot of resources still exists.
```
% kubectl get all -A | grep -i vm
victoria-metrics         pod/vmagent-vmagent-7458478449-km26q                                  2/2     Running       0              41m
victoria-metrics         pod/vminsert-main-cluster-b86978644-4gkpg                             1/1     Running       0              44m
victoria-metrics         pod/vmselect-main-cluster-0                                           1/1     Running       0              44m
victoria-metrics         pod/vmstorage-main-cluster-0                                          1/1     Running       0              44m
victoria-metrics         service/vmagent-vmagent                                            ClusterIP      10.96.232.54    <none>          8429/TCP                                         41m
victoria-metrics         service/vminsert-main-cluster                                      ClusterIP      10.96.204.234   <none>          8480/TCP                                         44m
victoria-metrics         service/vmselect-main-cluster                                      ClusterIP      None            <none>          8481/TCP                                         44m
[...]

PR #593, opened 2023-07-21
move clean up cr job to operator chart

Haleygo:
I think the right way to fix https://github.com/VictoriaMetrics/helm-charts/issues/548 is moving the cleanup job to operator, instead of have a second job in operator.
But there will be version dependency for victoria-metrics-k8s-stack, it will only have cleanup job when having new operator version.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/values.yaml
+++ b/charts/victoria-metrics-k8s-stack/values.yaml
@@ -11,6 +11,12 @@ argocdReleaseOverride: ""
 # -- also checkout here possible ENV variables to configure operator behaviour https://docs.victoriametrics.com/operator/vars.html
 victoria-metrics-operator:
   enabled: true
+  # -- Tells helm to clean up vm cr resources when uninstalling 
+  cleanupCRD: true
+  cleanupImage:
+    repository: gcr.io/google_containers/hyperkube
+    tag: v1.18.0
+    pullPolicy: IfNotPresent
 
   createCRD: false # we disable crd creation by operator chart as we create them in this chart
   operator:
--- a/charts/victoria-metrics-k8s-stack/values.yaml
+++ b/charts/victoria-metrics-k8s-stack/values.yaml
@@ -11,12 +11,6 @@ argocdReleaseOverride: ""
 # -- also checkout here possible ENV variables to configure operator behaviour https://docs.victoriametrics.com/operator/vars.html
 victoria-metrics-operator:
   enabled: true
-  # -- Tells helm to remove CRD after chart remove
-  cleanupCRD: true
-  cleanupImage:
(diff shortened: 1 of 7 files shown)


Haleygo:
PTAL @Amper @f41gh7 
I can reverse the clean up job in operator if that feels wrong to you)

Haleygo:
@k1rk Any suggestion on this one)

charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -11,7 +11,7 @@ argocdReleaseOverride: ""
 # -- also checkout here possible ENV variables to configure operator behaviour https://docs.victoriametrics.com/operator/vars.html
 victoria-metrics-operator:
   enabled: true
-  # -- Tells helm to remove CRD after chart remove
+  # -- Tells helm to clean up vm cr resources when uninstalling 
   cleanupCRD: true
```
f41gh7, line 15: It looks, that cleanup CRD related variables must be moved into `victoria-metrics-operator.operator` section. 

For backward compatibility, we should support both section, templating it with _helpts.tpl.

charts/victoria-metrics-operator/values.yaml
```diff
@@ -13,6 +13,13 @@ image:
 # -- with this option, if you remove this chart, all crd resources will be deleted with it.
 createCRD: true
 
+# -- Tells helm to clean up vm cr resources when uninstalling
+cleanupCRD: true
```
f41gh7, line 17: I think, it may break some installations. So I suggest to assign default value for `False` at operator chart and `True` for k8s-stack chart.

Since It may break some installations, if CRD managed by another chart.

Probably, we should set `cleanupCRD` for `True`, if `createCRD` was set to true?

charts/victoria-metrics-k8s-stack/values.yaml
Haleygo: I don't see why move to `victoria-metrics-operator.operator` section, cleanCRD managed by victoria-metrics-operator chart, not by operator itself.

charts/victoria-metrics-operator/values.yaml
Haleygo: Fixed, changed the default `cleanupCRD` for operator chart.
>Probably, we should set cleanupCRD for True, if createCRD was set to true?

We also have this logic in k8s-stack chart.
https://github.com/VictoriaMetrics/helm-charts/blob/6e17ba3aa8995cac1c09a4b1d09f4d6551fba51f/charts/victoria-metrics-k8s-stack/values.yaml#L21

So `cleanupCRD` and `createCRD` are not that related, I think we can keep the simplest and let user to config.

Issue #590, opened 2023-07-18
helm chart size optimization


hagen1778:
So far, helm charts for k8s stack could exceed allowed size limit. This prevents us from adding default dashboards that users might need. Would be great if we find a workaround for this. For example, to create a sub-charts for dashboards. 

This issue requires a proper research to check if proposed solution is efficient.

Haleygo:
Helm v3 stores chart state in a kubernetes secret by default and secret are limited to 1M based on etcd’s configuration. Helm has this issue from the very beginning[like [8281](https://github.com/helm/helm/issues/8281), [1413](https://github.com/helm/helm/issues/1413)] and there is some proposal like [split release data into multiple secrets](https://github.com/helm/helm/issues/10986), but never got merged or solved.
The data in release secret is compressed and decoded, means that chart itself can be bigger then 1M as long as the secret of deployed release is less than 1M. 
I deployed the latest 0.17.3 using default values.yaml. The vm-k8s-stack-chart size is 4.4M, and the secret created by helm is 966K[vm-stack-release.json]. [Helm will compress the data by gzip and base64-encode once](https://github.com/helm/helm/blob/main/pkg/storage/driver/util.go#L35-L51) and secrets API will base64-encode again, so we can get the real data stored in etcd by `cat vm-stack-release.json｜ jq .data.release | tr -d '"' | base64 -d`, which size is 724k[vm-stack-raw-data-in-etcd.json], means we still got some buffer under default installation)
![image 2 3 4](https://github.com/VictoriaMetrics/helm-charts/assets/39937150/1960a5d0-148d-48ca-a632-63313c73f4fa)
To gain the full raw data of the helm release, decode the vm-stack-release.json by `cat vm-stack-release.json｜ jq .data.release | tr -d '"' | base64 -d | base64 -d | gzip -d`, using some json analyser and get below data:
![image 2 2](https://github.com/VictoriaMetrics/helm-charts/assets/39937150/84a28572-abcd-4267-905f-7e24dd952573)
![image 2 3 2](https://github.com/VictoriaMetrics/helm-charts/assets/39937150/3d098951-e431-4f14-a950-bcc92c88aed6)
The top two size eater are “templates” and “manifest”:
1. “templates” stores all the resources under `charts/victoria-metrics-k8s-stack/templates`, field `data` is the content of each resource yaml after base64
[...]

PR #603, opened 2023-08-06
reduce k8s-stack chart size

Haleygo:
address https://github.com/VictoriaMetrics/helm-charts/issues/590
There are two commits, first one is to move `/crds` to crd subchart, second is to move `/dashboards` to grafana-dashboard subchart.
After first move, secret data size drops from 741KB->447KB[294KB ⬇️], full data size drops from 5.1MB->3.8MB[1.3MB ⬇️].
All that comes from crd.yaml. Although crd.yaml raw file size is 973KB, after base64 encoding, it’s over 1.26MB. Before saving to secret, got gzip and another base64 encoding, size becomes about 290KB.
image
Second move drops secret data size from 447KB->223KB[224KB ⬇️], full data size drops from 3.63MB->1.74MB[1.89MB ⬇️].
All that comes from /dashboards, the gzip compression ratio on dashboard.yaml is better than crd.yaml apparently.
image
image
But second move brings some side-effects, see those changes under `charts/victoria-metrics-k8s-stack/values.yaml`. Dashboards resources used to got values from the original `grafana` subchart and reuse some value like `.Values.kubelet.enabled`. Now those values need to be redefined in subchart, it could break the old values setting for old users. I tried to use anchor in yaml to sync subchart values from the original ones, but it failed because anchors are evaluated at the time the values file is loaded into memory, but fields updated via the set flag or other occur after the file was loaded.[more [detail](https://github.com/helm/helm/issues/9558#issuecomment-811915980)]. 
So now it seems no way to reuse the original values completely and it will be a breaking change to all the fields defined in `grafana-dashboards` subchart’s values.

So the first move itself already has pretty good effect, we are now have more than 500KB under the default value installation, which looks promising to me. I think we can stop there for a while, when we got more dashboards, we can try the next move. 
If that's ok to you, I will revert the second commit. @f41gh7 @hagen1778 @Amper


hagen1778:
I don't have enoguh understanding to review this. I'd rely on @Amper @f41gh7 reviews

k1rk:
@Haleygo 
thanks. this is great

i love the idea to go with CRDs only for now, to not introduce breaking change just yet. also anchors hurting readability of values.

so lets do just CRDs for now

and regarding dashboards i was going to explore https://github.com/grafana/thema
or maybe https://github.com/K-Phoen/dark

if you have some time and energy you could explore this options too

Haleygo:
Leave only CRDs now, PTAL) @k1rk @f41gh7 @Amper

k1rk:
@Haleygo lets bump minor version of k8s-stack chart also for this change

Haleygo:
> @Haleygo lets bump minor version of k8s-stack chart also for this change

Sure, updated!

hagen1778:
@Amper @f41gh7 @k1rk did we come up with a resolution? I see this change as a big improvement for helm charts!

k1rk (approved):

k1rk:
approved. @Haleygo thanks for a great work

f41gh7 (approved):
LGTM

f41gh7:
Thanks!

PR #616, opened 2023-08-23
Reduce default scrape interval

kvaps:
Many dashboards (eg. Node Exporter Full), are expecting lower scrape interval.

Thus i think it's better to set 20s by default. See the screenshot of grafana before and after this change:
изображение


f41gh7 (approved):
LGTM

f41gh7:
Thanks for contribution!

Issue #612, opened 2023-08-21
cadvisor: set `honor_timestamps` to false by default


hagen1778:
The issue is a follow-up after https://github.com/VictoriaMetrics/VictoriaMetrics/issues/4697#issuecomment-1656550287

The motivation to set `honor_timestamps` to false by default is explained [here](https://github.com/VictoriaMetrics/VictoriaMetrics/issues/4697#issuecomment-1656540535).

hagen1778:
cc @f41gh7

BongoEADGC6:
honorTimestamps throws an error with victoria-metrics-agent chart version 0.9.5. I believe this expects honor_timestamps in its place.

Haleygo:
> disable honorTimestamps for cadvisor by default #617

@BongoEADGC6 Thanks, should be fixed in https://github.com/VictoriaMetrics/helm-charts/pull/626.

hagen1778:
Thanks @Haleygo !
The change is available starting from https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-agent/CHANGELOG.md#096

PR #617, opened 2023-08-23
disable honorTimestamps for cadvisor by default

Haleygo:
address https://github.com/VictoriaMetrics/helm-charts/issues/612


f41gh7 (approved):
LGTM

PR #619, opened 2023-08-23
[victoria-metrics-k8s-stack]:vmalert notifiers and replicas of alertmanager

identw:
If instead of replicas in notifiers add only a service, then alerts are sent to only one alertmanager and this is a problem since alerts in the alertmanager cluster are not synchronized, unlike silence.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
+++ b/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
@@ -119,8 +119,13 @@ remoteWrite:
 {{- end }}
 remoteRead: {{ include "victoria-metrics-k8s-stack.vmReadEndpoint" . | nindent 2 }}
 datasource: {{ include "victoria-metrics-k8s-stack.vmReadEndpoint" . | nindent 2 }}
+{{- if .Values.alertmanager.enabled }}
+{{- $alertManagerReplicas := .Values.alertmanager.spec.replicaCount | default 1 }}
 notifiers:
-    - url: {{ printf "http://%s-%s.%s.svc:9093" "vmalertmanager" (include "victoria-metrics-k8s-stack.fullname" .) .Release.Namespace }}
+    {{- range $n := until (int $alertManagerReplicas) }}
+    - url: {{ printf "http://%s-%s-%d.%s-%s.%s.svc:9093" "vmalertmanager" (include "victoria-metrics-k8s-stack.fullname" $) $n "vmalertmanager" (include "victoria-metrics-k8s-stack.fullname" $) $.Release.Namespace }}
+    {{- end }}
+{{- end }}
 {{- end }}
 
 {{/*


charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
```diff
@@ -119,8 +119,11 @@ remoteWrite:
 {{- end }}
 remoteRead: {{ include "victoria-metrics-k8s-stack.vmReadEndpoint" . | nindent 2 }}
 datasource: {{ include "victoria-metrics-k8s-stack.vmReadEndpoint" . | nindent 2 }}
+{{- $alertManagerReplicas := .Values.alertmanager.spec.replicaCount | default 1 }}
```
Haleygo, line 122: I think we are missing the [check](https://github.com/VictoriaMetrics/helm-charts/blob/ab04561891f13ece0c98d822d3e64d927364476d/charts/victoria-metrics-k8s-stack/templates/victoria-metrics-operator/alertmanager/vmalertmanager.yaml#L1) of vmalertmanager creation here, since vmalertmanager won't be created if `.Values.alertmanager.enabled=false`

identw: Hello @Haleygo.
I didn't add this check because it wasn't there before. Old behaviour was:

```
$ helm template vm vm/victoria-metrics-k8s-stack --set alertmanager.enabled=false | grep notifiers -A1
  notifiers:
  - url: http://vmalertmanager-vm-victoria-metrics-k8s-stack.monitoring.svc:9093:
```

And after my change it would be:
```
$ helm template vm . --set alertmanager.enabled=false  | grep notifiers -A1
  notifiers:
  - url: http://vmalertmanager-vm-victoria-metrics-k8s-stack-0.vmalertmanager-vm-victoria-metrics-k8s-stack.monitoring.svc:9093
```

VM operator can't work with empty notifiers:
```
cannot create or update vmalert: vm-1, cannot find any notifiers. At cr.spec.Notifiers, cr.spec.Notifier, cr.spec.notifierConfigRef and discovered alertmanager are empty",
  "stacktrace": "sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\tsigs.k8s.io/controller-runtime@v0.13.1/pkg/internal/controller/controller.go:326\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\tsigs.k8s.io/controller-runtime@v0.13.1/pkg/internal/controller/controller.go:273\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2\n\tsigs.k8s.io/controller-runtime@v0.13.1/pkg/internal/controller/controller.go:234
```

I want to say that helm chart and VM operator are not provided for combination vmalert enabled and alertmanager disabled. But I can do as you say, what behavior would you like if alertmanager.enabled=false?

Haleygo: >Perhaps developer of VM operator need to use option -notifier.blackhole=true for such case.

`-notifier.blackhole=true` is used only when people knowing they want blackhole notification, not as default value.

>Should I remove notifiers if alertmanager.enabled=false or disable completely vmalert for this case?

Should only move `notifiers`  field, there still [other fields](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl#L151) for vmAlertSpec here.

identw: I added condition
```
{{- if .Values.alertmanager.enabled }}
...
{{- end }}
```

Haleygo (approved):
LGTM!
Thanks!

PR #620, opened 2023-08-24
update index with new versions

Amper:


f41gh7 (approved):
LGTM

k1rk:
you can ignore date change for all old releases and just add new one

PR #636, opened 2023-08-31
fix grafana dashboards so ephemeral_storage does not mess with memory query

elee1766:
the currently, the dashboards use kube_pod_container_resource_requests metric with a constraint on unit=byte to calculate memory maximums. 

this becomes misleading when also setting the "ephemeral storage" resource, because the unit is also bytes, but the resource is ephemeral_storage, not memory, resulting in incorrect values


Haleygo (approved):
LGTM!
Thanks for the fix!

PR #651, opened 2023-09-06
Updated CRDs in operator chart

Amper:


charts/victoria-metrics-operator/templates/crd.yaml
```diff
@@ -11679,7 +11684,7 @@ spec:
                       acceptEULA:
                         description: AcceptEULA accepts enterprise feature usage,
                           must be set to true. otherwise backupmanager cannot be added
-                          to single/cluster version. https://victoriametrics.com/legal/esa/
+                          to single/cluster version. https://victoriametrics.com/legal/eula/
```
hagen1778, line 11687: Why this change? I believe @tenmozes deliberately changed eula to esa

Amper: I just updated the CRD from the release bundle, and the @tenmozes changes aren't released yet. But i can leave them, because it doesn't affect anything. I’ll fix it.

Amper: Fixed

PR #666, opened 2023-09-11
add changelog for #659

Haleygo:
make up changelog for https://github.com/VictoriaMetrics/helm-charts/pull/659


Amper:
Too late, I already released with this change without CHANGELOG :)
It is better to make a  review comment to pour PR consistency already with the entry in CHANGELOG

Haleygo:
> Too late, I already released with this change without CHANGELOG :) It is better to make a review comment to pour PR consistency already with the entry in CHANGELOG

Ops, sorry, got it.

PR #684, opened 2023-09-21
Fixed behavior of `vmalert.remoteWriteVMAgent`

Amper:
Now `remoteWrite.url` for VMAlert is correctly generated considering endpoint, name, port and http.pathPrefix of VMAgent if `vmalert.remoteWriteVMAgent=true`


hagen1778 (approved):
LGTM

PR #687, opened 2023-09-25
Fix: the statefulset pod name length is more than 63

barnettZQG:
If the statefulset name length is 63, the pod name length of the statefulset is more than 63. The Pods' names take the form `<statefulset name>-<ordinal index>`.

So this PR limits the statefulset name max length to 60.

### DIFF
--- a/charts/victoria-metrics-cluster/templates/_helpers.tpl
+++ b/charts/victoria-metrics-cluster/templates/_helpers.tpl
@@ -91,13 +91,13 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 */}}
 {{- define "victoria-metrics.vmstorage.fullname" -}}
 {{- if .Values.vmstorage.fullnameOverride -}}
-{{- .Values.vmstorage.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- .Values.vmstorage.fullnameOverride | trunc 60 | trimSuffix "-" -}}
 {{- else -}}
 {{- $name := default .Chart.Name .Values.nameOverride -}}
 {{- if contains $name .Release.Name -}}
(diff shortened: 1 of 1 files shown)


barnettZQG:
@f41gh7  @k1rk  @tenmozes  PTAL

Haleygo:
Thanks for pointing out!

Haleygo (commented):
But actually statefulSet names can't exceed 52 since there will be a metadata label `controller-revision-hash` for pod which value is also limited by 63, see https://github.com/kubernetes/kubernetes/issues/64023 for details.

f41gh7:
I'm not really happy with trimming generated values. It may lead to objects name clashing and various bugs. 

 I think, it's better to enforce user to modify chart name or provide own shorter value for name generation.

barnettZQG:
> I'm not really happy with trimming generated values. It may lead to objects name clashing and various bugs.
> 
> I think, it's better to enforce user to modify chart name or provide own shorter value for name generation.

Maybe we could provide a default name. such as `victoria-metrics-cluster-vmstorage`, and then support overwriting it with the value of the key `values.vmstorage.fullnameOverride`.

Haleygo:
> Maybe we could provide a default name. such as `victoria-metrics-cluster-vmstorage`, and then support overwriting it with the value of the key `values.vmstorage.fullnameOverride`.

I don't think having a default name is a good choice, it will cause duplicted resources name when chart be installed multiple times, then need extra config for multiple releases.

>I'm not really happy with trimming generated values. It may lead to objects name clashing and various bugs.

Yeah, I also think trunc is not the best approach, but it can help user successfully install the chart in most cases.

f41gh7 (approved):
LGTM

Issue #695, opened 2023-10-03
remove chart version related labels from pod labels


f41gh7:
Currently there are chart version related labels at the spec.podMetadata.Labels section.

 It causes pod restarts if chart version was changed. But it's no necessary.

 At first, fix victoria-metrics-cluster helm chart. 

 Also, make sure, that we do not include any dymanic labels into podMetadata.

Amper:
There's one problem - removing those labels will still result in a restart once :)

PR #697, opened 2023-10-04
From pod labels removed dynamic label `helm.sh/chart`

Amper:
From pod labels removed dynamic label `helm.sh/chart` to avoid restarting every time the chart is updated without changing the pods parameters. 

**Note that this time it will lead the pods restart** 

issue: #695


f41gh7 (approved):
LGTM

PR #733, opened 2023-10-19
vmselect-statefulset breaks when resources are set

marrau:
Fixing position of license mount.

When "vmselect.statefulSet.enabled=true" and "vmselect.resources.requests != {}" then the Template fails with 

``` 
error calling include: template: victoria-metrics-cluster/templates/_helpers.tpl:309:18: executing "chart.license.mount" at <.Values.license.secret.name>: nil pointer evaluating interface {}.license
```


Haleygo:
@marrau Thank you for the pull request!
But there are other places need to be fix too, should be addressed in https://github.com/VictoriaMetrics/helm-charts/pull/734, please take a look)

Close this in favor of https://github.com/VictoriaMetrics/helm-charts/pull/734.

PR #734, opened 2023-10-20
fix deployment `volumeMounts` when providing enterprise license key f…

Haleygo:
…or VictoriaMetrics enterprise

### DIFF
--- a/charts/victoria-metrics-gateway/Chart.yaml
+++ b/charts/victoria-metrics-gateway/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 description: Victoria Metrics Gateway - Auth & Rate-Limitting proxy for Victoria Metrics
 name: victoria-metrics-gateway
 version: 0.1.49
-appVersion: v1.94.0
+appVersion: v1.94.1
 sources:
   - https://github.com/VictoriaMetrics/helm-charts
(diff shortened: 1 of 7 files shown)


tenmozes (approved):

charts/victoria-metrics-gateway/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 description: Victoria Metrics Gateway - Auth & Rate-Limitting proxy for Victoria Metrics
 name: victoria-metrics-gateway
 version: 0.1.49
-appVersion: v1.94.0
+appVersion: v1.94.1
```
zekker6, line 5: Seems like `appVersion` was accidentally increased instead of version, could you update the `version`?

zekker6 (changes requested):
LGTM overall, please take a look at 1 comment

charts/victoria-metrics-gateway/Chart.yaml
Haleygo: Sorry, following up...

Haleygo: Please see https://github.com/VictoriaMetrics/helm-charts/pull/735

PR #740, opened 2023-10-30
vmanomaly: new version, reflect config updates

zekker6:
- add "bearer_token" parameter for read/write configs
- add "verify_tls" parameter for read/write configs
- add "extra_filters" parameter for read config

### DIFF
--- a/charts/victoria-metrics-anomaly/values.yaml
+++ b/charts/victoria-metrics-anomaly/values.yaml
@@ -121,6 +121,15 @@ remote:
     basicAuth:
       username: ""
       password: ""
+
+    # -- Bearer token to use for authentication
+    bearer_token: ""
+
+    # -- Whether to verify TLS certificates when connecting to read endpoint
+    verify_tls: true
+
+    # -- List of extra filters to apply to all queries. See: https://docs.victoriametrics.com/#prometheus-querying-api-enhancements
+    extra_filters: [ ]
   write:
     # -- URL for data writing. Required for example "http://single-victoria-metrics-single-server.default.svc.cluster.local:8428" or "http://cluster-victoria-metrics-cluster-vminsert.default.svc.cluster.local:8480/insert/"
     url: ""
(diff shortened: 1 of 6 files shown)


hack/vmanomaly-lint-hack.yaml
```diff
@@ -2,8 +2,17 @@ eula: true
...
+    basicAuth:
+      username: asdf
+      password: asdf
+    extra_filters:
+      - '{cluster!=""}'
+      - '{job!=""}'
   write:
     url: http://cluster-victoria-metrics-cluster-vminsert.default.svc.cluster.local:8480/insert/
+    verify_tls: false
+    bearer_token: asdf
```
tenmozes, line 15: should we support file, secret?

zekker6: good point, secret should be an option

zekker6: hm, but atm there is no way to pass either an env variable or a path to the token to vmanomaly
wdyt about adding this in a next version? (after adding this to vmanomaly first)

tenmozes: sure, it works for me

tenmozes (approved):

charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -121,6 +121,15 @@ remote:
...
       password: ""
+
+    # -- Bearer token to use for authentication
+    bearer_token: ""
+
+    # -- Whether to verify TLS certificates when connecting to read endpoint
+    verify_tls: true
+
+    # -- List of extra filters to apply to all queries. See: https://docs.victoriametrics.com/#prometheus-querying-api-enhancements
+    extra_filters: [ ]
```
tenmozes, line 132: what about extra labels? for both - reading and writing

zekker6: For reading `extra_filters` can already supersede `extra_label` parameter and we haven't seen a reason to add both atm as it wasn't requested.

For writing vmanomaly appends all fields passed in `metric_format`, e.g.:
```
metric_format:
  name: "node_exporter_$VAR"
  query_key: "for"
  model: prophet
```
Here `name` and `query_key` are reserved names, but `model` label will be passed as-is and will result in appending `model=prophet` to all emitted metrics.

PR #751, opened 2023-11-07
vm-operator: add `extraObjects` to allow deploying additional resourc…

Haleygo:
…es with the chart release
And victoria-metrics-k8s-stack [uses](https://github.com/VictoriaMetrics/helm-charts/blob/add-extra-objects-for-operator-chart/charts/victoria-metrics-k8s-stack/Chart.yaml#L35) victoria-metrics-operator as dependency, it will get `extraObjects` as well.


zekker6 (approved):

Amper (approved):
LGTM

PR #755, opened 2023-11-13
kube state metrics helm chart is maintained by prometheus community

AlonGluz:
kube state metrics helm chart is maintained by prometheus team


zekker6 (approved):
@AlonGluz Thank you for the PR!

PR #757, opened 2023-11-15
allow custom readiness and liveness probes for vmselect and vminsert

jifwin:


charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml
```diff
@@ -110,20 +110,28 @@ spec:
               containerPort: {{ include "split-host-port" .Values.vminsert.extraArgs.opentsdbListenAddr }}
           {{- end }}
           readinessProbe:
+            {{- if not .Values.vminsert.probe.readiness.extra }}
```
Haleygo, line 113: 1. I think If those are extra readiness/liveness, they shouldn't override the default one;
2. Also need to add corresponding values in https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-cluster/values.yaml#L72

Haleygo:
I think we should modify the vminsert/vmselect probes like [vmstorage](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-cluster/values.yaml#L623) does now, it will be more flexible to add extra probes or remove default one.
https://github.com/VictoriaMetrics/helm-charts/blob/9e7356a142202a679cf84cb92bd9f48dd5cf1ef8/charts/victoria-metrics-cluster/templates/vmstorage-statefulset.yaml#L79-L82


And can you help updating the [changelog](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-cluster/CHANGELOG.md#L7) with some description as well?

jifwin:
Good idea, but how to move this default probe logic to values.yml?
`          
          readinessProbe:
            httpGet:
            {{- if index  .Values.vmselect.extraArgs "http.pathPrefix" }}
              path: {{ trimSuffix "/" (index .Values.vmselect.extraArgs "http.pathPrefix") }}/health
            {{- else }}
              path: /health
            {{- end }}
`

Haleygo:
> Good idea, but how to move this default probe logic to values.yml?
> 
> ```
>           readinessProbe:
>             httpGet:
>             {{- if index  .Values.vmselect.extraArgs "http.pathPrefix" }}
>               path: {{ trimSuffix "/" (index .Values.vmselect.extraArgs "http.pathPrefix") }}/health
>             {{- else }}
>               path: /health
>             {{- end }}
> ```

Yeah, thanks for pointing out. Seems like there is no easy way to unify vminsert&vmstorage's probes without possibility to break some users' installation now.
Ok, that's keep the way you proposed, please see reveiew comments [here](https://github.com/VictoriaMetrics/helm-charts/pull/757#discussion_r1396854004).

jifwin:
I renamed 'extra' to 'custom' to clarify that it overrides the default probe.
Regarding values.yaml, do you want me to add an example of a custom probe, e.g. like this?
```
  probe:
    readiness:
      initialDelaySeconds: 5
      periodSeconds: 15
      timeoutSeconds: 5
      failureThreshold: 3
      #custom: # -- override for the default probe
      #  tcpSocket:
      #    port: http
```

Haleygo:
>it overrides the default probe.

Is there any specific reason to override the default probe here? I thought you want to add an extra probe.

And those new fields need to be added to [values file](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-cluster/values.yaml) too.

jifwin:
I'm binding vmselect and vminsert to localhost for security reasons and using TLS sidecar to access these services on a diferent through a different port. Thus I need to change the definitions of probes to access that new port with HTTPS scheme. Also I want to replace tcpSocket probe with httpGet probe, since opening tcp to a sidecar proxy does not give much information, I need to actually call the service (vmselect/vminsert) behind the proxy.

charts/victoria-metrics-cluster/templates/vmselect-deployment.yaml
```diff
@@ -76,20 +76,28 @@ spec:
...
             httpGet:
             {{- if index  .Values.vmselect.extraArgs "http.pathPrefix" }}
               path: {{ trimSuffix "/" (index .Values.vmselect.extraArgs "http.pathPrefix") }}/health
             {{- else }}
               path: /health
             {{- end }}
               port: http
+            {{- else }}
+              {{- toYaml .Values.vmselect.probe.readiness.custom | nindent 12 }}
+            {{- end }}
```
zekker6, line 89: Can we move values for probe into values and just use `{{ .Values.vmselect.probe.readiness }}` then?
Combining this with [`tpl` filter](https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-tpl-function) should allow to keep logic with automatic handling of `http.pathPrefix` parameter.

This allows overriding any parameters to customize probe and avoid if/else nesting.

WDYT @Haleygo @jifwin ?

jifwin: Is that what you mean?
values.yaml
`
  probe:
    readiness: |
      httpGet:
        {{- if index  .Values.vminsert.extraArgs "http.pathPrefix" }}
        path: {{ trimSuffix "/" (index .Values.vminsert.extraArgs "http.pathPrefix") }}/health
        {{- else }}
        path: /health
        {{- end }}
        port: http
      initialDelaySeconds: 5
      periodSeconds: 15
      timeoutSeconds: 5
      failureThreshold: 3
`

vminsert-deployment.yaml
`
          readinessProbe: {{ tpl .Values.vminsert.probe.readiness . | nindent 12}}
`

zekker6: Yes, exactly. 
I'd like to ask @Haleygo for the feedback regarding which option seems to be suite this case better.

jifwin: It's worth nothing that it would break existing overrides of vm*.probe.readiness, as it would no longer be an object, but a string instead.

zekker6: You're right, somehow I've missed this completely, in this case just ignore this suggestion :)

Haleygo:
@jifwin 
I think this is the way to add this new feature without breaking old installation. Could you please add new `vminsert.probe.readiness.custom` field to the [values file](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-cluster/values.yaml) with example?
And help updating the [changelog](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-cluster/CHANGELOG.md#L7) with some description as well?

AndrewChubatiuk:
closing in a favour of #1212

Issue #770, opened 2023-11-27
Install CRDs using the helm crd directory


bh-tt:
Today I moved our vm operator from ansible to argocd-managed. Imagine my surprise when the `helm uninstall` command also deleted all the crd's, and our metric system was down until I got the new operator running, which then promptly deleted alle CRs related to victoriaMetrics.

Helm provides special handling for CRDs using the `crds` directory: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/. One major usecase is that helm does not delete the CRDs if the release is uninstalled.

Haleygo:
Thanks for the report!
But I don't think we should change victoria-metrics-operator chart to use /crds now:
1. moving the crds from /templates to /crds is a breaking change and will delete all the CRDs during upgrade.
2. helm hasn't provided a way to update CRDs if they're under /crds, see https://github.com/helm/helm/issues/7735. So having crd.yaml under /templates can help users to keep update with the latest CRDs.
So I think we should fix this by adding annotation `helm.sh/resource-policy: keep` for CRD resources.

Haleygo:
Should be fixed since [victoria-metrics-operator-0.27.7](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-operator-0.27.7), close as completed.

PR #771, opened 2023-11-28
victoria-metrics-operator: skip deleting victoriametrics CRD resources when…

Haleygo:
… uninstall release

address https://github.com/VictoriaMetrics/helm-charts/issues/770,
users might installed multiple victoriametrics releases and deleting CRDs will break other releases,
and normally there is no need to clean up CRDs.


zekker6:
Wdyt about adding some kind of automation to add these labels? Something like make target with `yq`.

I'm pretty sure that some day after CRDs update those annotations will be missed, so automation will make it much easier to avoid this.

Haleygo:
> Wdyt about adding some kind of automation to add these labels? Something like make target with `yq`.
> 
> I'm pretty sure that some day after CRDs update those annotations will be missed, so automation will make it much easier to avoid this.

I don't want to add extra operation to our release script, so I add a [crd.yaml file](https://github.com/VictoriaMetrics/helm-charts/blob/reserve-crds-when-uninstall-operator/charts/victoria-metrics-operator/crd.yam) outside of /templates and read&process it in inside [crd.yaml](https://github.com/VictoriaMetrics/helm-charts/blob/reserve-crds-when-uninstall-operator/charts/victoria-metrics-operator/templates/crd.yaml).
Please take a look @zekker6 @Amper

zekker6 (approved):

PR #776, opened 2023-12-07
adding important dashboards from kube-prometheus

k4kratik:


k4kratik:
Hi @Haleygo 
Thanks for your feedback. 
I understand that but I was thinking that I want to install these dashboards when I am running `helm install`.
If I fork the chart and modify it, any suggestions, How could I keep up with the latest chart releases. 
Any options to include these in `values.yaml` file ? that would be best! 
Thanks.

Haleygo:
> If I fork the chart and modify it, any suggestions, How could I keep up with the latest chart releases.

I think you can create a new chart, using victoria-metrics-k8s-stack and chart with dashboards as it's dependencies, then you only need to bump victoria-metrics-k8s-stack version from time to time.

k4kratik:
> Any options to include these in values.yaml file ? that would be best!
Thanks! @Haleygo, I was wondering If I can do something like this. I understand if not possible but still :)

PR #777, opened 2023-12-08
victoria-metrics-cluster: fix vminsert template for extra volumes

zekker6:
Previously, using `extraVolumes` would trigger an error:
```
Error: UPGRADE FAILED: template: templates/vminsert-deployment.yaml:168:12: executing "templates/vminsert-deployment.yaml" at <include "chart.license.volume" .>: error calling include: template: templates/_helpers.tpl:297:18: executing "chart.license.volume" at <.Values.license.secret.name>: can't evaluate field Values in type []interface {}
```
Caused by using `.` from unexpected context from `with` clause.

### DIFF
--- a/charts/victoria-metrics-cluster/CHANGELOG.md
+++ b/charts/victoria-metrics-cluster/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
-- TODO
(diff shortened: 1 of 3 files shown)


Haleygo (approved):
LGTM

PR #788, opened 2023-12-09
victoria-metrics-k8s-stack: optional allowCrossNamespaceImport in GrafanaDashboard(s)

plevart:
In *victoria-metrics-k8s-stack* Helm Chart, when values define `grafanaOperatorDashboardsFormat.enabled=true`, special `GrafanaDashboard` CRs get rendered that provide dashboards to Grafana Operator (https://grafana-operator.github.io/grafana-operator/). But those CRs don't get imported into Grafana instance when the Grafana Operator is not deployed to the same k8s namespace as the victoria-metrics-k8s-stack chart is, since `GrafanaDashboard` CRs are created in this namespace.
The provided patch allows overriding an additional boolean attribute in the chart values: `grafanaOperatorDashboardsFormat.allowCrossNamespaceImport` which is false by default. When set to true, the created `GrafanaDashboard` CRs will also contain this attribute and Grafana Operator will import them from a namespace that is different than the Grafana Operator's namespace.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/CHANGELOG.md
+++ b/charts/victoria-metrics-k8s-stack/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
-- TODO
+- Add optional allowCrossNamespaceImport to GrafanaDashboard(s) (#788)
 
 ## 0.18.9
 
--- a/charts/victoria-metrics-k8s-stack/README.md
+++ b/charts/victoria-metrics-k8s-stack/README.md
@@ -419,6 +419,7 @@ Change the values according to the need of the environment in ``victoria-metrics
 | grafana.sidecar.datasources.jsonData | object | `{}` |  |
 | grafana.vmServiceScrape.enabled | bool | `true` |  |
 | grafana.vmServiceScrape.spec | object | `{}` |  |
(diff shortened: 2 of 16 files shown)


Haleygo:
Thanks for the pull request!
You also need to add this to the dashboard sync script as a part of [standard header](https://github.com/VictoriaMetrics/helm-charts/blob/4a24092457c02d2c6b4f440b8109bf2fa1505420/charts/victoria-metrics-k8s-stack/hack/sync_grafana_dashboards.py#L97C5-L97C5).
And could you help adding some changelog [here](https://github.com/VictoriaMetrics/helm-charts/blob/4a24092457c02d2c6b4f440b8109bf2fa1505420/charts/victoria-metrics-k8s-stack/CHANGELOG.md?plain=1#L5)?

plevart:
@Haleygo I see. So the GrafanaDashboard resource template embedded in the python script is rendered with the same "values.yaml" as the dashboard templates in the Helm chart? I can just add the same modifications using same references to .Values dictionary?

plevart:
So, here it is, @Haleygo ...

Haleygo (approved):
LGTM!
Thanks for contribution!

PR #790, opened 2023-12-11
Allow using rootless podman instead of docker from Makefile

plevart:
Since `podman` is getting popular these days (I use it instead of docker), I propose the following changes that would allow using rootles podman containers instead of docker when podman is installed on PATH. The changes fall-back to using docker when podman is not found.

What do you say? Is it acceptable?


zekker6 (changes requested):
This is acceptable, but could you please implement it by using same approach as this one - https://github.com/VictoriaMetrics/VictoriaMetrics/commit/2474281f1b19b6dac68f796fb7b81153c90a9758 ?

Using this approach does not require adding script and allows replacing `docker` commands with `podman` easily.

plevart:
The problem using that approach is that one would have to manually set 3 independent variables then to get the Makefile to use podman:

DOCKER=podman
DOCKER_USER_OPTION=""
DOCKER_VOLUME_OPTION_SUFFIX=":z"

The defaults defined in Makefile would be:

DOCKER ?= docker
DOCKER_USER_OPTION ?= "--user $(shell id -u):$(shell id -g)"
DOCKER_VOLUME_OPTION_SUFFIX ?= ""

Is this still acceptable? How to give hints what values to use? Add them as comments into the Makefile? Or use another Makefile.podman that would define the variables and then include Makefile?

plevart:
So here it is. Is this acceptable?

zekker6:
In order to make it simpler to switch container runtime there could be a single variable which would control which container runtime to use, something similar to:
```
CONTAINER_RUNTIME?=docker

DOCKER ?= docker
DOCKER_USER_OPTION ?= "--user=$(shell id -u):$(shell id -g)"
DOCKER_VOLUME_OPTION_SUFFIX ?= ""

ifeq ($(CONTAINER_RUNTIME),podman)
	DOCKER=podman
	DOCKER_USER_OPTION=""
	DOCKER_VOLUME_OPTION_SUFFIX=":z"
endif
```

> Is this still acceptable? How to give hints what values to use? Add them as comments into the Makefile? Or use another Makefile.podman that would define the variables and then include Makefile?

Adding a comment around these variables should be enough for now. If you're interested, you can add information about make targets in readme(create a development section)/contribution guides to make it easier to discover.

plevart:
So what about this last variant? Single Makefile with ifeq blocks inside...

zekker6 (approved):
LGTM

@plevart Thank you for the contribution!

PR #797, opened 2023-12-13
license configuration: fix passing license via k8s secret

zekker6:
- fix reference of license by correct secret key
- fix flag name after rename in VictoriaMetrics


tenmozes (approved):

zekker6:
@Haleygo Thanks for noticing!
`secret.key` indeed does not exist, but in case `secret.items` is not provided all secret keys will be mapped as files so it will work.
I'll send a follow-up PR to remove obsolete `secret.key` configurations shortly.

PR #800, opened 2023-12-13
license configuration: remove `secret.key` references for license volumes

zekker6:
Follow-up for https://github.com/VictoriaMetrics/helm-charts/pull/797

Remove `secret.key` reference as it does not actually exist in k8s resource.


Haleygo (approved):
lgtm

PR #801, opened 2023-12-15
Add missed variable 'vmcluster' in rules and default value.

MemberIT:


zekker6 (approved):
Hello @MemberIT 
Thank you for the fix!

Issue #787, opened 2023-12-08
VMInsert service has no endpoints when TLS enabled


lucker7777:
Hello,

When I use _charts/victoria-metrics-cluster_ and set VMInsert to use TLS this way:
_extraArgs:
    tls: "true"_
    
the victoria-metrics-cluster-vminsert service has no endpoints.

I think the reason is failed VMInsert's proble, which is HTTP by default and when TLS is used, it fails.
After I do this:
_$ k edit deployments.apps victoria-metrics-cluster-vminsert_
_readinessProbe:
          failureThreshold: 3
          httpGet:
            path: /health
            port: http
            scheme: HTTP**S**_
 (Note the added "S"), the service has an active endpoint.
 
 So, there should be a solution for that, so I do not need to edit it in the deployment.
 
 Thanks.

zekker6:
Hello @lucker7777, victoria-metrics-cluster chart release [v0.11.11](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-cluster-0.11.11) included a fix for this.
I'll close the issue as completed, feel free to reopen if that doesn't work for you.

PR #826, opened 2024-01-10
victoria-metrics-cluster: switch probes scheme to `HTTPS` if vminsert…

Haleygo:
… and vmselect enabled tls in extraArgs
address https://github.com/VictoriaMetrics/helm-charts/issues/787


Amper (approved):
LGTM

PR #835, opened 2024-01-31
vmanomaly new config 1.9.2

Cambaza:


Cambaza:
@zekker6 can you, please, take a look?

charts/victoria-metrics-anomaly/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@
...
 - TODO
 
+## 1.0.0
+
+**Release date:** 2024-02-01
+
+![AppVersion: v1.9.2](https://img.shields.io/static/v1?label=AppVersion&message=v1.9.2&color=success&logo=)
+![Helm: v3](https://img.shields.io/static/v1?label=Helm&message=v3&color=informational&logo=helm)
+
+- Add ability to render vmanomaly config as is via `vmanomalyConfig`.
```
Haleygo, line 14: Hey, change notes should be under `Next release` and new version title will be generated automatically.
As the [release CI](https://github.com/VictoriaMetrics/helm-charts/blob/master/.github/workflows/release.yaml#L56) reads content under `Next release` and determine whether to release new version.
```
## Next release

- Add ability to render vmanomaly config as is via `vmanomalyConfig`.
- xxx
```

charts/victoria-metrics-anomaly/CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Next release
 
-- TODO
+- Add ability to render full vmanomaly config via `config` parameter.
```
zekker6, line 5: It is better to update this item to clearly state that this is a breaking change and `config` updates will be required for upgrade.
Currently, it seems like this only adds an additional field/option to provide full configuration without breaking compatibility.

zekker6 (approved):

dima-vm (approved):
Great, thanks!

PR #850, opened 2024-02-08
charts/victoria-metrics-k8s-stack: add `promscrape.dropOriginalLabels` by default

zekker6:
This is useful to lower vmagent's memory usage by default by sacrificing built-in relabeling debug UI.


Amper (approved):
LGTM

PR #852, opened 2024-02-08
Validate manifests with `kubectl` in CI

MisguidedEmails:
Helm only outputs a warning if there are schema issues, but will continue anyway. This adds a step before install that performs validation against the k8s server before proceeding with the helm chart install.

Would be nice to have it in the linting CI, however this is the only place in CI that we have a cluster to work with. In order to validate the manifests properly, `--dry-run=server` is required.

~Not sure if you would want it as a linting step, so I've left it as a draft PR.~ actually, since the testing runs on PR's and on Push anyway, that seems fine.

I've updated so that it verifies with `kubectl` post-install since we will rely on CRDs for some charts. I'm not able to test the CI directly, but it seems we setup a k3d cluster for each chart, so it's safe to use `--skip-clean-up` and the CI will terminate everything. Though if this isn't the case then we could remove it manually post-validation


zekker6:
Hi @MisguidedEmails, thank you for the PR!

Just in order to confirm that it works could you make some change in charts to trigger the updated CI job?
After CI will finish, and we'll see that it works you can drop this commit to keep history clean.

zekker6 (approved):
@MisguidedEmails Thank you for your contribution!

PR #863, opened 2024-02-15
Add HPA support for VMAgent

rakesh-von:


charts/victoria-metrics-agent/values.yaml
```diff
@@ -253,6 +253,13 @@ persistence:
   # -- Bind Persistent Volume by labels. Must match all labels of targeted PV.
   matchLabels: {}
 
+# Horizontal Pod Autoscaling
+horizontalPodAutoscaler:
```
zekker6, line 257: Could you rename it to be `autoscaling`?
It will make it shorter and less error-prone to use.

charts/victoria-metrics-agent/README.md
```diff
@@ -293,6 +293,10 @@ Change the values according to the need of the environment in ``victoria-metrics
 | extraVolumeMounts | list | `[]` |  |
 | extraVolumes | list | `[]` |  |
 | fullnameOverride | string | `""` |  |
+| horizontalPodAutoscaler.enabled | bool | `false` | Use HPA for vmagent |
+| horizontalPodAutoscaler.maxReplicas | int | `10` | Maximum replicas for HPA to use to to scale vmagent |
+| horizontalPodAutoscaler.minReplicas | int | `1` | Minimum replicas for HPA to use to scale vmagent |
+| horizontalPodAutoscaler.metrics | object | `` | Metric for HPA to use to scale vmagent |
```
zekker6, line 299: Based on `values.yaml` `metrics` field should be a list. Could you sync docs to make sure it includes the latest updates from values?

zekker6 (commented):
Hello @rakesh-von! Thank you for the PR, please check a few comments.

rakesh-von:
Hi @zekker6 ,

Made the proposed changes, can you please review it now?

charts/victoria-metrics-agent/README.md
rakesh-von: done

charts/victoria-metrics-agent/values.yaml
rakesh-von: done

dg-nvm:
@antoinedeschenes it is usefull in multi-layer architecture, using remote-write protocol. Therefore vmagent scraping is not the one that is processing data later in the chain, and we would want to scale these.

charts/victoria-metrics-agent/values.yaml
rakesh-von: @zekker6 in every other chart it is `horizontalPodAutoscaler`, if i change it to `horizontalPodAutoscaling` or just `autoscaling` then vmagent would be the outlier. Are you sure you want this to be renamed to `autoscaling`?

zekker6: @rakesh-von Good point, let's keep it as it is now then. Thanks!

zekker6 (approved):

PR #864, opened 2024-02-16
Vmanomaly v1.10 changes

Cambaza:
- changes according to `vmanomaly` v1.10 multimodel config
- docs fixes

### DIFF
--- a/charts/victoria-metrics-anomaly/CHANGELOG.md
+++ b/charts/victoria-metrics-anomaly/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
-- apply [v1.10](https://docs.victoriametrics.com/anomaly-detection/changelog/#v1100) change in [models section](https://docs.victoriametrics.com/anomaly-detection/components/models/)(does not break previous deployments). 
+- apply [v1.10](https://docs.victoriametrics.com/anomaly-detection/changelog/#v1100) change in [models section](https://docs.victoriametrics.com/anomaly-detection/components/models/): add configuration for using multiple models at once via `models`. Old `model` field is deprecated and will be automatically converted to `models` definition starting from [v1.10](https://docs.victoriametrics.com/anomaly-detection/changelog/#v1100).
 - docs fixes
 
 ## 1.0.0
--- a/charts/victoria-metrics-anomaly/CHANGELOG.md
+++ b/charts/victoria-metrics-anomaly/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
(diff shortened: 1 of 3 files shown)


fred-navruzov (approved):
Thanks, LGTM!

charts/victoria-metrics-anomaly/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: victoria-metrics-anomaly
 description: Victoria Metrics Anomaly Detection - a service that continuously scans Victoria Metrics time series and detects unexpected changes within data patterns in real-time.
-version: 1.0.0
-appVersion: v1.9.2
+version: 1.0.1
```
zekker6, line 4: This change breaks compatibility with previous one in terms of configuration `model -> models` so incrementing patch version is misleading here. Could you increment feature version here instead?

fred-navruzov: I'd also vote for it to be `1.1.0`.
@zekker6 , technically, `models` *it's a backward compatible change* (so the old-style `model` is implicitly converted to new-style `models` under the hood, but it will definitely be deprecated in future releases) - please see [header of models page for an example](https://docs.victoriametrics.com/anomaly-detection/components/models/), so what's the best action we can take for helm charts given this context?

Cambaza: added to changelog

charts/victoria-metrics-anomaly/CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Next release
 
-- TODO
+- apply [v1.10](https://docs.victoriametrics.com/anomaly-detection/changelog/#v1100) change in [models section](https://docs.victoriametrics.com/anomaly-detection/components/models/)(does not break previous deploymet).
```
fred-navruzov, line 5: a typo: deploymet -> deployment(s).

charts/victoria-metrics-anomaly/CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Next release
 
-- TODO
+- apply [v1.10](https://docs.victoriametrics.com/anomaly-detection/changelog/#v1100) change in [models section](https://docs.victoriametrics.com/anomaly-detection/components/models/)(does not break previous deployments).
```
zekker6, line 5:
```suggestion
- apply [v1.10](https://docs.victoriametrics.com/anomaly-detection/changelog/#v1100) change in [models section](https://docs.victoriametrics.com/anomaly-detection/components/models/): add configuration for using multiple models at once via `models`. Old `model` field is deprecated and will be automatically converted to `models` definition starting from [v1.10](https://docs.victoriametrics.com/anomaly-detection/changelog/#v1100).
```

zekker6 (approved):

PR #871, opened 2024-02-21
Make alertmanager deployment more extensible

jenrik:
Add the ability to add extra arguments, containers, volumes, and volume mounts to the alertmanager deployment.

I implemented this change to able to add a sidecar to the alertmanager for automatic config reloading, and decided to round it out with the features from the server deployment.

### DIFF
--- a/charts/victoria-metrics-alert/README.md
+++ b/charts/victoria-metrics-alert/README.md
@@ -115,6 +115,10 @@ Change the values according to the need of the environment in ``victoria-metrics
 | alertmanager.configMap | string | `""` |  |
 | alertmanager.enabled | bool | `false` |  |
 | alertmanager.extraArgs | object | `{}` |  |
+| alertmanager.extraContainers | list | `[]` |  |
+| alertmanager.extraHostPathMounts | list | `[]` |  |
+| alertmanager.extraVolumeMounts | list | `[]` |  |
+| alertmanager.extraVolumes | list | `[]` |  |
 | alertmanager.image | string | `"prom/alertmanager"` |  |
 | alertmanager.imagePullSecrets | list | `[]` |  |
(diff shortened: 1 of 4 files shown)


charts/victoria-metrics-alert/README.md
```diff
@@ -115,6 +115,11 @@ Change the values according to the need of the environment in ``victoria-metrics
 | alertmanager.configMap | string | `""` |  |
 | alertmanager.enabled | bool | `false` |  |
 | alertmanager.extraArgs | object | `{}` |  |
+| alertmanager.extraArgs | object | `{}` |  |
```
zekker6, line 118: It seems like `extraArgs` is now duplicated, could you please remove the duplicate?

zekker6 (commented):
Hello @jenrik, thank you for the PR!

Please, could you check the comment?

Also, could you add an entry in a changelog about this change?
It is located here: https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-alert/CHANGELOG.md
You need to replace `- TODO` entry with description of this change so that it will be included into change list of the release.

charts/victoria-metrics-alert/README.md
jenrik: fix in fda37843f64e8b2179470067910ba4fdfadd34c5

jenrik:
@zekker6 I fixed the duplicate value in the docs, and add an entry to the change log as requested

jenrik:
I think the failed test might be flaky, I can't figure out a good reason for why it failed.

zekker6 (approved):
LGTM, @jenrik thank you for the contribution!

zekker6:
FYI: @jenrik this PR was released in https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-alert-0.9.0

Issue #865, opened 2024-02-19
[victoria-metrics-k8s-stack] Kind VMAgent "null" value in remoteWrite cause failed to helm install


tantm3:
Hi everyone!

I try to install `victoria-metrics-k8s-stack` with a custom remote write config.
When I installing the helm chart it showed an error:
```
Error: INSTALLATION FAILED: 1 error occurred:
	* VMAgent.operator.victoriametrics.com "vm-k8s-victoria-metrics-k8s-stack" is invalid: spec.remoteWrite[0]: Invalid value: "null": spec.remoteWrite[0] in body must be of type object: "null"

```
I use the helm template command and saw that there is an extra `null value` from the VMAgent remoteWrite spec 
```
# Source: victoria-metrics-k8s-stack/templates/victoria-metrics-operator/vmagent.yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAgent
metadata:
  name: test-victoria-metrics-k8s-stack
  namespace: default
  labels: 
    helm.sh/chart: victoria-metrics-k8s-stack-0.18.11
    app.kubernetes.io/name: victoria-metrics-k8s-stack
    app.kubernetes.io/instance: test
    app.kubernetes.io/version: "v1.96.0"
    app.kubernetes.io/managed-by: Helm
spec:
  externalLabels:
    cluster: vmonitor-prod
  extraArgs:
    promscrape.streamParse: "true"
    remoteWrite.maxDiskUsagePerURL: 2GB
  image:
    tag: v1.96.0
  nodeSelector:
    backend: "true"
  remoteWrite:
  - null
  - basicAuth:
      password:
        key: password
        name: vm-k8s-remote-write-prod
      username:
        key: username
        name: vm-k8s-remote-write-prod

```
- I think that `null value` causes the installation process to fail.
- Here are the helm value file that I used:
[values.zip](https://github.com/VictoriaMetrics/helm-charts/files/14326319/values.zip)

tantm3:
More information:
I think the `null value` is caused by disabling the vmsingle component.
When we make the `vmsingle enabled=false`, it will produce the `null values`. In the opposite, the remotewrite will have the value like this
```
# Source: victoria-metrics-k8s-stack/templates/victoria-metrics-operator/vmagent.yaml
apiVersion: operator.victoriametrics.com/v1beta1
kind: VMAgent
metadata:
  name: vm-k8s-victoria-metrics-k8s-stack
  namespace: default
[...]

PR #886, opened 2024-02-23
charts/victoria-metrics-k8s-stack: fix remoteWrite URLs templating

zekker6:
Fixes templating of remoteWrite URLs in case built-in VMSingle and VMCluster are both disabled. See: https://github.com/VictoriaMetrics/helm-charts/issues/865

P.S. I've tested changes on a local cluster and it works just as expected while it fails in CI. I'll investigate why CI fails for k8s-stack and address this in a separate PR.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/CHANGELOG.md
+++ b/charts/victoria-metrics-k8s-stack/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
-- TODO
+- Fix templating of VMAgent `remoteWrite` in case both `VMSingle` and `VMCluster` are disabled. See [this issue](https://github.com/VictoriaMetrics/helm-charts/issues/865) for details.
 
 ## 0.19.1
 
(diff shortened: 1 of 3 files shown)


Haleygo:
> I've tested changes on a local cluster and it works just as expected while it fails in CI

Should caused by incompatible value change in upstream grafana chart, see https://github.com/grafana/helm-charts/pull/2977. 
We can fix this using fixed grafana chart `v7.3.1` instead of `7.3.*`( ps, this wouldn't effect [victoria-metrics-k8s-stack-0.19.1](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-k8s-stack-0.19.1) since it's already packed with grafana 7.3.1.) in dependency. 
And looks like they will fix this with new version soon, I think we can ignore it if no new release from our side, wdyt?@zekker6

Haleygo (approved):
LGTM!

PR #904, opened 2024-03-01
[fix] Boolean for k8s rules

passie:
Added the option so k8s rules can be turned on and off again via a boolean.


Haleygo (approved):
LGTM!
Thank you for contribution!

Issue #931, opened 2024-03-17
Labels should be truncated


roderik:
I was trying to install the cluster chart, and it died during install because of 

```
Pod observability-vm-75ff870e-victoria-metrics-cluster-vmstorage-0 in StatefulSet observability-vm-75ff870e-victoria-metrics-cluster-vmstorage failed error: Pod "obse ││ rvability-vm-75ff870e-victoria-metrics-cluster-vmstorage-0" is invalid: metadata.labels: Invalid value: "observability-vm-75ff870e-victoria-metrics-cluster-vmstorage-64d9d4dbd9": must be no more than 63 characters
```

Haleygo:
related https://github.com/VictoriaMetrics/helm-charts/pull/687#pullrequestreview-1643368066

AndrewChubatiuk:
made changes to k8s-stack labels for dashboards and rules in #1212. it should be available in the next release

AndrewChubatiuk:
hey @ToroNZ . this issue was fixed in release [v0.25](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-k8s-stack-0.25.0)

PR #936, opened 2024-03-20
Labels should be truncated

brandshaide:
Fixing #931

### DIFF
--- a/charts/victoria-metrics-cluster/templates/_helpers.tpl
+++ b/charts/victoria-metrics-cluster/templates/_helpers.tpl
@@ -46,17 +46,17 @@ Create the name of the service account
 Create unified labels for victoria-metrics components
 */}}
 {{- define "victoria-metrics.common.matchLabels" -}}
-app.kubernetes.io/name: {{ include "victoria-metrics.name" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/name: {{ include "victoria-metrics.name" . | trunc 60 }}
+app.kubernetes.io/instance: {{ .Release.Name | trunc 60 }}
 {{- end -}}
 
 {{- define "victoria-metrics.common.metaLabels" -}}
-helm.sh/chart: {{ include "victoria-metrics.chart" . }}
-app.kubernetes.io/managed-by: {{ .Release.Service }}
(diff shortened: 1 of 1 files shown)


charts/victoria-metrics-cluster/templates/_helpers.tpl
```diff
@@ -45,61 +45,61 @@ Create the name of the service account
 {{/*
 Create unified labels for victoria-metrics components
 */}}
-{{- define "victoria-metrics.common.matchLabels" -}}
+{{- define "victoria-metrics.common.matchLabels" | trunc 60 -}}
```
Haleygo, line 48: I don't think this is the right fix, we should limit the length of each lable value.

brandshaide: thanks, you mean truncating it e. g. [here](https://github.com/VictoriaMetrics/helm-charts/blob/504c6d03f17eb9356ba23b7a1ee9fc6f07782ea0/charts/victoria-metrics-cluster/templates/vminsert-deployment.yaml#L34)

zekker6: @brandshaide It means truncating each label value separately. For example: https://github.com/VictoriaMetrics/helm-charts/blob/624f078a10be80669b0e48238ecd9bd5d049bd5a/charts/victoria-metrics-cluster/templates/_helpers.tpl#L49-L50

This could be updated to become:
```
app.kubernetes.io/name: {{ include "victoria-metrics.name" . | trunc 60 }}
app.kubernetes.io/instance: {{ .Release.Name | trunc 60 }}
```


Also, note that using filters in `define` is not allowed by Helm syntax. You can validate your changes locally by using `make lint`(make file [here](https://github.com/VictoriaMetrics/helm-charts/blob/624f078a10be80669b0e48238ecd9bd5d049bd5a/Makefile#L62-L71)).
By default, it uses docker to run linting, you can switch to local helm adding `HELM=helm-local` to the end of command: `make lint HELM=helm-local`

brandshaide: @zekker6 thanks for your help, I've applied the changes (also did run linting without errors)

zekker6 (approved):
@brandshaide Thank you for the contribution!

PR #937, opened 2024-03-20
fix value for retentionPerion and update docs

MadEngineX:
## Description 

I have vm-stack contolled by k8s operator. I deploy it using your Helm Chart [victoria-metrics-k8s-stack](https://github.com/VictoriaMetrics/helm-charts/tree/504c6d03f17eb9356ba23b7a1ee9fc6f07782ea0/charts/victoria-metrics-k8s-stack).

When I specify `vmcluster.spec.retentionPeriod` as integer, nothing changes. Also we if we take look on [victorialogs documentation](https://docs.victoriametrics.com/victorialogs/#retention) `retentionPeriod` must be in Prometheus [time-duration format](https://prometheus.io/docs/prometheus/latest/querying/basics/#time-durations): `1d`, `2w`, `5y`.

When I change `vmcluster.spec.retentionPeriod` value in my Helm release from `7` to `7d` it starts working:

image
 
I wasn't able to check your values for VictoriaLogs Helm charts, so I didn't made any changes in their values (for example [here](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-logs-single/values.yaml#L32))


charts/victoria-metrics-single/values.yaml
```diff
@@ -47,8 +47,8 @@ server:
   priorityClassName: ""
   # -- Overrides the full name of server component
   fullnameOverride:
-  # -- Data retention period in month
```
Haleygo, line 50: I think keeping 1->1m as default value is more suitable for most cases.

Haleygo (commented):
Thanks for the PR!
But for most of the changing descriptions, like 
```
| server.env | list | `[]` | Env variables -- Additional environment variables (ex.: secret tokens, flags) https://github.com/VictoriaMetrics/VictoriaMetrics#environment-variables |
```
or
```
| server.service.type | string | `"ClusterIP"` | Node port nodePort: 30000 -- Service type |
```
I don't see how they be improved.

And please note, the variable description in the doc is generated automatically from the values file. You should change the comments in each values.yaml and run `make gen-docs` to update the docs.

charts/victoria-metrics-single/values.yaml
MadEngineX: Thanks for your review!

Extra changes in documentation was made because I ran `helm-docs` against `make gen-docs` according to the contribution guide:
- https://github.com/VictoriaMetrics/helm-charts/blob/master/CONTRIBUTING.md

charts/victoria-metrics-k8s-stack/README.md
```diff
@@ -509,7 +509,7 @@ Change the values according to the need of the environment in ``victoria-metrics
 | vmcluster.ingress.storage.pathType | string | `"Prefix"` |  |
 | vmcluster.ingress.storage.tls | list | `[]` |  |
 | vmcluster.spec.replicationFactor | int | `2` |  |
-| vmcluster.spec.retentionPeriod | string | `"14"` |  |
+| vmcluster.spec.retentionPeriod | string | `"14d"` | Data retention period starting from "1d" (one day) up to "100y" (100 years). See these [docs](https://prometheus.io/docs/prometheus/latest/querying/basics/#time-durations) for the supported duration formats. |
```
Haleygo, line 512: It has differences with prometheus query duration, like unit `m` is not valid. Please see https://docs.victoriametrics.com/single-server-victoriametrics/#retention.

charts/victoria-metrics-single/values.yaml
Haleygo: Sorry for misleading, I mean keeping 1 which means 1 month, `m` is not a valid unit for retentionPeriod here.

MadEngineX: Reverted this changes.

charts/victoria-metrics-k8s-stack/README.md
Haleygo: Please update description here, retentionPeriod is not limited by 100y and the link is wrong.

MadEngineX: @Haleygo is everything fine now?

Haleygo (approved):
LGTM!
Thanks!

PR #939, opened 2024-03-24
charts/victoria-metrics-cluster: support customize vmstorage VolumeClaimTemplate name

chenlujjj:


charts/victoria-metrics-cluster/templates/vmstorage-statefulset.yaml
```diff
@@ -222,7 +222,7 @@ spec:
   {{- if and .Values.vmstorage.persistentVolume.enabled (not .Values.vmstorage.persistentVolume.existingClaim) }}
   volumeClaimTemplates:
     - metadata:
-        name: vmstorage-volume
+        name: {{ .Values.vmstorage.persistentVolume.name | default "vmstorage-volume" }}
```
Haleygo, line 225: I think it's better to add this to the [values file ](https://github.com/VictoriaMetrics/helm-charts/blob/504c6d03f17eb9356ba23b7a1ee9fc6f07782ea0/charts/victoria-metrics-cluster/values.yaml#L573)with default value `vmstorage-volume`
```
  persistentVolume:
    # -- Create/use Persistent Volume Claim for vmstorage component. Empty dir if false. If true,  vmstorage will create/use a Persistent Volume Claim
    enabled: true
    name: vmstorage-volume
```

chenlujjj: thx, fixed

Haleygo (approved):
LGTM

PR #941, opened 2024-03-25
fix: typo on whether

gabyfulchic:
Was working on the VM k8s stack and observed this typo.


zekker6 (approved):
Nice catch, @gabyfulchic thank you for the PR!

PR #944, opened 2024-03-26
extraArgs slice args support

AndrewChubatiuk:
vmauth has extraAuthHeader, which can be set multiple times. without these changes it can be set only once. propagated this feature to other containers in other charts as well


Haleygo (commented):
LGTM!
Could you help adding description to changelogs of each chart?

Haleygo (approved):

PR #966, opened 2024-04-15
victoria-metrics-operator: clean up vmauth when uninstall chart with …

Haleygo:
…`cleanupCRD: true`

since vmauth also has victoriametrics `finalizers`, thus can't be deleted after operator is deleted


f41gh7 (approved):
LGTM

PR #973, opened 2024-04-17
charts: properly truncate labels for long release names

zekker6:
Properly truncate "app.kubernetes.io/managed-by" and "app.kubernetes.io/instance" labels for long release names. Apply limit of 63 chars for all truncated labels.

This is a follow-up to cd6921b33551bcc274692d8fdeed890f3b441006


Haleygo (approved):
lgtm

Issue #569, opened 2023-06-28
Override namespace through values.yaml


morozovcookie:
Hello! I would like to purpose to add possibility to override namespace via values.yaml. Especially for the k8s-stack chart.

In my case I have a single chart where I have all I need as dependencies. By default Helm Chart will be installed into `{{ .Release.Namespace }}`, but I need to control this behavior. I want to setup any dependencies in the separate namespaces.

Similar [feature](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/_helpers.tpl#L124) exists in the kube-prometheus-stack.

PR #994, opened 2024-04-25
introduce templating for namespace for victoria-metrics-k8-stack

brandshaide:
fixing #569


zekker6 (changes requested):
Hello @brandshaide 

In order to implement support of namespace override properly it is needed to do the following:
- add a helper similar to one on prometheus-k8s-stack - https://github.com/prometheus-community/helm-charts/blob/7d50bd5aa88b73fce522e19053216f9128451be1/charts/kube-prometheus-stack/templates/_helpers.tpl#L139-L148 . Helper should check if `.Values.namespace` if not empty and use it's value in this case, otherwise it should default to `.Release.Namespace` value. Note that default value for `.Values.namespace` must be an empty string as otherwise all deployments will switch namespace after upgrade.
- update all references to `.Release.Namespace` in this chart to use newly added helper

AndrewChubatiuk:
closing this PR in a favour of https://github.com/VictoriaMetrics/helm-charts/pull/1392

PR #995, opened 2024-04-25
openshift_compatibility

Baboulinet-33:
Hello,

I wanted to use victoria logs on openshift but openshift don't like fixed securityContext ("fsGroup" "runAsUser" "runAsGroup").
So I've make some modification on chart to be compatible with openshift.

The modification are clearly a copy/paste of bitnami solution (https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_compatibility.tpl) - you can keep the modification on `_helpers.tpl` or add the following dependency on all charts:
```yaml
dependencies:
- name: common
  repository: oci://registry-1.docker.io/bitnamicharts
  version: 2.x.x
```

For all securityContext parameter, you have a property `enabled` to enable this parameter.
And you have the parameter  `global.compatibility.openshift.adaptSecurityContext ` with value `auto`, `force`, `disabled`:
- auto: apply if the detected running cluster is Openshift
- force: perform the adaptation always
- disabled: do not perform adaptation

An example with auto:
```yaml
spec:
      containers:
        - name: victoria-logs-single-server
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:
              - ALL
            readOnlyRootFilesystem: true
          image: "victoriametrics/victoria-logs:v0.5.2-victorialogs"
          [...]
      securityContext:
        fsGroup: 2000
        runAsNonRoot: true
        runAsUser: 1000
      terminationGracePeriodSeconds: 60
      volumes:
        - name: server-volume
          emptyDir: {}
```

With force:
**openshift: force**
```yaml
spec:
      containers:
        - name: victoria-logs-single-server
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:
              - ALL
            readOnlyRootFilesystem: true
          image: "victoriametrics/victoria-logs:v0.5.2-victorialogs"
          [...]
      securityContext:
        runAsNonRoot: true
      terminationGracePeriodSeconds: 60
      volumes:
        - name: server-volume
          emptyDir: {}
```

### DIFF
--- a/charts/victoria-logs-single/templates/_helpers.tpl
+++ b/charts/victoria-logs-single/templates/_helpers.tpl
@@ -139,3 +139,36 @@ Return if ingress supports pathType.
 []
 {{- end -}}
 {{- end -}}
+
+{{/* 
+Return true if the detected platform is Openshift
+Usage:
+{{- include "common.compatibility.isOpenshift" . -}}
+*/}}
+{{- define "common.compatibility.isOpenshift" -}}
+{{- if .Capabilities.APIVersions.Has "security.openshift.io/v1" -}}
+{{- true -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Render a compatible securityContext depending on the platform. By default it is maintained as it is. In other platforms like Openshift we remove default user/group values that do not work out of the box with the restricted-v1 SCC
+Usage:
+{{- include "common.compatibility.renderSecurityContext" (dict "secContext" .Values.containerSecurityContext "context" $) -}}
+*/}}
+{{- define "common.compatibility.renderSecurityContext" -}}
+{{- $adaptedContext := .secContext -}}
+{{- if .context.Values.global.compatibility -}}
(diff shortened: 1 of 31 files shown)


Baboulinet-33:
Hi, I forgot to respond but I addressed the points mentionned

Haleygo (approved):
LGTM

Haleygo:
Thanks for contribution!

TheMeier:
just a remark for some future similar changes. maybe it is a good idea to introduce new toggles with defaults set so they match the previous behaviour. 
IMHO the default for `server.podSecurityContext.enabled` should have been `true`

Haleygo:
> just a remark for some future similar changes. maybe it is a good idea to introduce new toggles with defaults set so they match the previous behaviour. IMHO the default for `server.podSecurityContext.enabled` should have been `true`

@TheMeier Yes, thank you for notice and sorry for inconvenience!
It shouldn't break any previous customized securityContext, please see https://github.com/VictoriaMetrics/helm-charts/pull/1023.

PR #996, opened 2024-04-25
feat(crd): view and admin aggregate roles for CRDs

reegnz:
Kubernetes has view and admin roles that aggregate permissions from
other clusterroles that have the right annotations.

The advantage to this is that the view clusterrole doesn't need to be
edited whenever a new CRD is installed, but can still be extended so
view and admin also provide view and admin access to new CRDs.

The functionality is similar to drop-in files in systemd and other
linux/unix tooling.

The change adds the aggregate role functionality to the victoria-metrics-operator
helm chart. So whenever the CRDs are installed the user can also choose
to install the clusterroles that extend the view and admin default
clusterroles.

### DIFF
--- a/charts/victoria-metrics-operator/templates/crd_clusterrole.yaml
+++ b/charts/victoria-metrics-operator/templates/crd_clusterrole.yaml
@@ -1,0 +1,64 @@
+{{- /* do not update crds here, please update in /victoria-metrics-operator/crd.yaml */ -}}
+{{- /* this is used to add "helm.sh/resource-policy: keep" annotation for each crd */ -}}
+{{- /* see this pull request https://github.com/VictoriaMetrics/helm-charts/pull/771 for details */ -}}
+{{- if .Values.createCRD }}
+{{- if .Values.rbac.createAggregate }}
+{{- $files := .Files }}
+{{- $fileContentsList := $files.Get "crd.yaml" | splitList "---" }}
+{{- $groups := dict }}
+{{- range $fileContentsList }}
+    {{- $fileContents := . | fromYaml }}
+    {{- $group := $fileContents.spec.group }}
+    {{- $plural:= $fileContents.spec.names.plural }}
+    {{- $resources := get $groups $group | default (list) }}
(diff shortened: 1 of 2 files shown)


reegnz:
I tested it on a k8s cluster the following way:
- apply rendered clusterroles to cluster
- `kubectl get clusterrole view -o yaml`
- validate that victoriametrics readonly permissions are now part of the `view` clusterrole.

Haleygo:
Hello @reegnz ,
Do you have other application which need to use clusterrole `victoriametrics:admin` or `victoriametrics:readonly` apart from vm-operator?

reegnz:
Yeah, CD apps like ArgoCD, fluxcd, Spinnaker, etc.

charts/victoria-metrics-operator/templates/crd_clusterrole.yaml
```diff
@@ -0,0 +1,64 @@
+{{- /* do not update crds here, please update in /victoria-metrics-operator/crd.yaml */ -}}
+{{- /* this is used to add "helm.sh/resource-policy: keep" annotation for each crd */ -}}
+{{- /* see this pull request https://github.com/VictoriaMetrics/helm-charts/pull/771 for details */ -}}
```
Haleygo, line 3: Could you replace this statement with relavant info about clusterroles here?

charts/victoria-metrics-operator/values.yaml
```diff
@@ -40,6 +40,14 @@ rbac:
   create: true
   # Note: The PSP will only be deployed, if Kubernetes (<1.25) supports the resource.
   pspEnabled: true
+  # -- create agregate clusterroles for CRD readonly and admin permissions
+  createAggregate: true
```
Haleygo, line 44: nit: maybe rename it to `createAggregateClusterRoles`?

Haleygo (commented):
LGTM!
Please see some comments.

Haleygo (approved):
LGTM!

PR #1002, opened 2024-05-07
Update service_account.yaml with proper end statement placement

tylerturk:
The prior configuration allowed the generation of an empty service account yaml file with indented labels or annotations.

image


Haleygo (approved):
LGTM!

Haleygo:
Thanks for the fix!

Issue #341, opened 2022-07-10
Prometheus Operator CRDs


maxpain:
Hello. I have `victoria-metrics-k8s-stack` deployed into my cluster. I'm trying to install some other helm charts, like `nginx` with the `serviceMonitor` option enabled, which creates `ServiceMonitor` CRD.
And I'm getting this error:

```
Error: UPGRADE FAILED: resource mapping not found for name: "nginx-ingress-nginx-controller" namespace: "" from "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"
ensure CRDs are installed first
```

Why doesn't `victoria-metrics-k8s-stack` include Prometheus Operator CRDs like `ServiceMonitor`?

f41gh7:
@k1rk  WDYT, Should we embed it into k8s-stack?

evaneaston:
I'm curious about this too.  I'm running vm k8s stack instead of prometheus, tried to add `openebs-monitoring` and ran into the same issue.  With vm offering feature parity in metrics collection and querying plus a bunch of other stuff, one could imagine wanting to maintain operator feature parity by supporting the `monitoring.coreos.com` CRDs, or at least some of them such as `PodMonitor`, `PrometheusRule` ,  `ServiceMonitor`, and `Probe`

But I have no idea how much work this would be or if operator-level compatibility is something the VM team wants to chase.

evaneaston:
Sure.  I can add the CRDs to get the other monitoring chart to load.  But I still need to either translate it's CRDs into VM operator CRDs or run a `kube-prometheus-stack` and configure it to forward data to `vm` and manually copy any dashboards it creates from the  `kube-prometheus-stack`  into my  `victoria-metrics-k8s-stack` grafana.

evaneaston:
Oh, nice!  I hadn't waded through the vm stack chart enough to spot this.  Thank you!

b-a-t:
I was under the impression that `victoriametrics-operator` installs those CRDs(as well as vm-k8s-stack), but apparently, that's not yet true.

Please implement some workaround for this issue, I see that operator already has all those CRDs in one file in the `operator/hack/prom_crd/prometheus_operator_crd.yaml`, so just needs to install them if enabled.

b-a-t:
[...]

PR #1022, opened 2024-05-14
[victoria-metrics-k8s-stack] Prometheus Operator CRDs (#341)

beliys:
Hi! Small patch for https://github.com/VictoriaMetrics/helm-charts/issues/341


Haleygo:
I have one concern though, CRDs in prometheus-operator-crds are under `/crds/templates`, so they are treated as regular templates and will be managed by Helm. That means prometheus CRDs also will be deleted when uninstall VM release[with prometheus-operator-crds.enabled], and it could break all the prometheus CR resources user has, which is not good for VM-k8s-stack IMO.

beliys:
@Haleygo Yes, but this is an optional feature. It is always possible to make prometheus-operator-crds independent, it sets prometheus-operator-crds.enabled=false and sets prometheus-operator-crds as a separate chart.
But I think there will not be many such users. Most users of victoria-metrics-k8s-stack appreciate being able to set everything with one chart.

Haleygo:
>Most users of victoria-metrics-k8s-stack appreciate being able to set everything with one chart.

Thanks, I'm ok with the default value `prometheus-operator-crds.enabled=false`.
Do you have second opinion? @zekker6 @f41gh7 @Amper

zekker6 (approved):
LGTM

f41gh7 (approved):
LGTM

Issue #1001, opened 2024-05-03
helm install failed when serviceMonitor is enabled


ChristianGeie:
i'am using chart victoria-metrics-operator@0.30.3. While applying the chart first time in my k8s cluster i'am faced the error

`error":"unable to build kubernetes objects from release manifest: resource mapping not found for name: \"victoria-metrics-operator\" namespace: \"monitoring\" from \"\": no matches for kind \"VMServiceScrape\" in version \"operator.victoriametrics.com/v1beta1\"\nensure CRDs are installed first"`

Seems helm wants to create a VMServiceScrape resource for the operator itself because i setup `serviceMonitor.enabled: true` in helm chart values, but all related crd's are not in place.

If i disable the servicemonitor using the default values out of your helm chart, crds are all ceated and everything works like a charm.
Afterwards I can also switch on the servicemonitor in the Helm Chart again.

However, this requires changing the configuration; I would expect the installation of the operator to be successful within one run.

ChristianGeie:
As far as I have found out, it is not possible to get a health status for cr's out of Helm.
My "solution" so far: Applying the helmchart twice, one only for deploying all related crds, the other one deploys the operator but has a dependency to the other helmchart..
See https://fluxcd.io/flux/components/helm/helmreleases/#dependencies

PR #1037, opened 2024-05-28
Move CRD to victoria-metrics-operator-crd chart

beliys:
Hello! Created `victoria-metrics-operator-crd` similarly to `prometheus-operator-crds` in the prometheus chart. It seems to me that this patch should solve the following problems:

1. No need to sync CRD for victoria-metrics-k8s-stack
2. It will be possible to install separately from victoria-metrics-operator and removing the victoria-metrics-operator chart will not remove the CRD.
3.  Avoiding issues like https://github.com/VictoriaMetrics/helm-charts/issues/1001


beliys:
I need help with fixing broken tests.

Haleygo:
Hello!
Now operator and k8s-stack maintain crds in different ways:
1. In victoria-metrics-operator, the crd yaml is under /templates, it won't be deleted when `helm uninstall` since there is annotation `helm.sh/resource-policy: keep`.
The bright side of this, users don't need to upgrade crd by hand when `helm upgrade`, and the installation won't break with default values, but there is problem like https://github.com/VictoriaMetrics/helm-charts/issues/1001.
2. In victoria-metrics-k8s-stack, the crd yaml is always under `/crds` which is not managed by helm after install, so crd upgrade needs to be performed by hand.

It looks not good but works for a long time, and afaik there is no "perfect" way in helm3 to manage CRD.

This PR solves https://github.com/VictoriaMetrics/helm-charts/issues/1001, but also breaks the automatic crd upgrade in victoria-metrics-operator chart.

beliys:
Thanks, but how to fix https://github.com/VictoriaMetrics/helm-charts/issues/1001 without breaks the automatic crd upgrade in victoria-metrics-operator chart?

beliys:
Is it possible to specify victoria-metrics-operator-crd as a dependency of victoria-metrics-operator but also automatically update the CRD if needed?

Haleygo:
>but how to fix https://github.com/VictoriaMetrics/helm-charts/issues/1001 without breaks the automatic crd upgrade in victoria-metrics-operator chart?

As far as I know it's not possible(

beliys:
@Haleygo What do you think about not changing victoria-metrics-operator? But for victoria-metrics-k8s-stack to make a dependency on victoria-metrics-operator-crd as in this patch?
It seems to me that victoria-metrics-operator-crd in a separate chart is still a little better than inside victoria-metrics-k8s-stack. In this case, we can also update the CRD in the chart, as you suggested by automatic pipeline in operator repo. I'll redo this PR if you agree.

Haleygo:
>It seems to me that victoria-metrics-operator-crd in a separate chart is still a little better than inside victoria-metrics-k8s-stack.

I don't see the difference, could you share why you think it's better?
The victoria-metrics-operator-crd is never meant to be a independent chart, as there is no need to publish a new chart to allow installing crds without operator or k8s-stack, it was introduced to [reduce the release secret size of k8s-stack](https://github.com/VictoriaMetrics/helm-charts/pull/603).

beliys:
@Haleygo I'm sorry for the delay in replying. In fact, there are not many advantages. The biggest advantage of a separate chart is that if someone needs to maintain their own fork of victoria-metrics-k8s-stack, they won't need to update the CDR manually. They will simply add victoria-metrics-operator-crd to the dependency. Just a little more flexibility. But at the same time, I don't see any disadvantages compared to having the CRD folder inside victoria-metrics-k8s-stack.

Haleygo:
I think the disadvantages would be:
1. need to maintain a new chart, and release new CRD chart before upgrade k8s-stack chart;
2. it's harder for users to know which version of CRD chart and k8s-stack chart are compatible.

beliys:
You're right. Thank you, I'm closing this PR.

Haleygo:
@towolf 
No, the disadvantages listed [here](https://github.com/VictoriaMetrics/helm-charts/pull/1037#issuecomment-2146894404) are still valid.
You can just use [victoria-metrics-operator](https://github.com/VictoriaMetrics/helm-charts/tree/master/charts/victoria-metrics-operator) as CRD chart, it puts CRD resources under `/template` instead of `/crd`.

PR #1048, opened 2024-06-06
victoria-metrics-cluster: change components image template

Haleygo:
This change syncs the image tags with `.Chart.AppVersion` directly, and makes it possible to upgrade chart without modifying image tag for users who use customized images with suffix `-enterprise` or `-scratch`. 
But it brings breaking change for user who is using non-default tags, they need to remove the `-cluster` suffix in tags.

### DIFF
--- a/charts/victoria-metrics-agent/CHANGELOG.md
+++ b/charts/victoria-metrics-agent/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## Next release
 
-- TODO
+**Update note: VictoriaMetrics components image tag template have been changed, tag suffix like `-scratch`, `-cluster`, `-enterprise` can be specified in `.Values.<component>.image.variant`, `.Values.<component>.image.tag` can be omitted if it's the same as `.Chart.AppVersion`.
+
+- support specifying image tag suffix like "-enterprise" for VictoriaMetrics components using `.Values.<component>.image.variant`.
 
 ## 0.10.8
 
--- a/charts/victoria-metrics-agent/CHANGELOG.md
+++ b/charts/victoria-metrics-agent/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
-**Update note: VictoriaMetrics components image tag template have been changed, tag suffix like `-scratch`, `-cluster`, `-enterprise` can be specified in `.Values.<component>.image.variant`, `.Values.<component>.image.tag` can be omitted if it's the same as `.Chart.AppVersion`.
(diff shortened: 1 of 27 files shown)


Haleygo:
@zekker6 @Amper  If you're ok with the change, I will do the same for other charts.

charts/victoria-metrics-cluster/values.yaml
```diff
@@ -51,9 +51,13 @@ vmselect:
...
     repository: victoriametrics/vmselect
     # -- Image tag
-    tag: v1.101.0-cluster
+    # override Chart.AppVersion
+    tag: ""
     # -- Image pull policy
     pullPolicy: IfNotPresent
+    # Variant of the image to use.
+    # e.g. cluster, enterprise-cluster
+    variant: "cluster"
```
zekker6, line 60: I would suggest treating `cluster` as a part of default tag and using `variant` to provide `-scratch` or `-enterprise-scratch` only.
This will be backward compatible as well as easier to use as user will only need to add the required prefix instead of populating an existing value and also adding suffix: `cluster-enterprise-scratch` vs `enterprise-scratch`

Haleygo: We have two tags for cluster component, `v1.101.1-cluster` and `v1.101.1-enterprise-cluster`. So if `-cluster` is provided by `tag` instead of `variant`, user will have to update `tag` when they upgrade.

Haleygo: Changed the default values, it's backward compatible now.

zekker6:
@Haleygo This approach looks good to me, could you proceed to update other charts as well?

Haleygo:
@zekker6 updated, plz check

charts/victoria-metrics-agent/CHANGELOG.md
```diff
@@ -2,7 +2,9 @@
 
 ## Next release
 
-- TODO
+**Update note: VictoriaMetrics components image tag template have been changed, tag suffix like `-scratch`, `-cluster`, `-enterprise` can be specified in `.Values.<component>.image.variant`, `.Values.<component>.image.tag` can be omitted if it's the same as `.Chart.AppVersion`.
```
zekker6, line 5: Rephrased a bit to avoid making an impression that this is a breaking change:
```suggestion
**Update note**: The VictoriaMetrics components image tag template has been updated. This change introduces the option to specify tag suffixes like `-scratch`, `-cluster`, `-enterprise` using `.Values.<component>.image.variant`. Additionally, you can now omit `.Values.<component>.image.tag` to automatically use the version specified in `.Chart.AppVersion`.

zekker6 (approved):

PR #1062, opened 2024-06-13
charts/victoria-metrics-k8s-stack: do not add "cluster" label by default

zekker6:
Adding "cluster" label by default is only useful for multi-cluster setups, which is a rare case. It is also hard to override this value to an empty map due to merge logic in Helm.


Haleygo (approved):
lgtm

PR #1069, opened 2024-06-16
Prometheus converter disabled by default in stack chart

SQLJames:
Based on the documentation in the values file and in the [operator migration documentation](https://docs.victoriametrics.com/operator/migration/#objects-conversion). I believe that this should be set to false initially. This is similar to the [values in the operator stand alone chart](https://github.com/VictoriaMetrics/helm-charts/blob/dae3850b4cd9b1ca8a742a72dde9e0be0ddffc11/charts/victoria-metrics-operator/values.yaml#L69) where it is set to disabled false.

```yaml
operator:
    # -- By default, operator converts prometheus-operator objects.
    disable_prometheus_converter: true
```

### DIFF
--- a/charts/victoria-metrics-k8s-stack/CHANGELOG.md
+++ b/charts/victoria-metrics-k8s-stack/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
-- TODO
+- Enable [conversion of Prometheus CRDs](https://docs.victoriametrics.com/operator/migration/#objects-conversion) by default. See [this](https://github.com/VictoriaMetrics/helm-charts/pull/1069) pull request for details.
 
 ## 0.23.2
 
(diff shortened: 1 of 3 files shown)


zekker6:
Hi @SQLJames, could you add a line into a [changelog](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-k8s-stack/CHANGELOG.md) about this change as well?

SQLJames:
Happy to add it, Do you want me to create a minor version of the chart also or just add the value to the `TODO` list E.G.
```
# CHANGELOG for `victoria-metrics-k8s-stack` helm-chart

## Next release

- Updated Value for `disable_prometheus_converter`. By default, operator should convert prometheus-operator objects.

```

zekker6:
@SQLJames I've updated the changelog and bumped chart version in https://github.com/VictoriaMetrics/helm-charts/pull/1069/commits/f016cbe194c8fc145e9f97786c2ebeeb1e7e82e5 in this PR. 

LGTM, I'll wait for a peer review before merging this one

zekker6 (approved):

Haleygo (approved):

PR #1110, opened 2024-07-04
Create unified svc for every vm agent shard

moraresearch:


Haleygo:
Hi,
Could you elaborate on why you need a separate service for each vmagent shard?

moraresearch:
Hello! the reason is that after their creation nothing targets to the shards, that is why I need to play lottery with trying to reach the shard that collects from the node I am interesting using svc that targets to all of them (and when we are speaking about really reasonably sharded vmagent in could be 5 new endpoints in a group of 8 thousand) or doing something manually. 
But if there is a configured svc to every shard I can configure for example vmauth to proxy me between them using path and thats it. 

> Hi, Could you elaborate on why you need a separate service for each vmagent shard?

Haleygo:
Thanks for the info!
But I don't think this should be done in chart, instead, it should be implemented by operator itself. I've created the [issue](https://github.com/VictoriaMetrics/operator/issues/1022) in operator.

f41gh7:
> Hello! the reason is that after their creation nothing targets to the shards, that is why I need to play lottery with trying to reach the shard that collects from the node I am interesting using svc that targets to all of them (and when we are speaking about really reasonably sharded vmagent in could be 5 new endpoints in a group of 8 thousand) or doing something manually. But if there is a configured svc to every shard I can configure for example vmauth to proxy me between them using path and thats it.
> 
> > Hi, Could you elaborate on why you need a separate service for each vmagent shard?

I don't think it's a good solution actually.

I suggest to use following workarounds:
1) use vmagent web-ui to check targets picked and skipped by current shard. See this issue for details https://github.com/VictoriaMetrics/VictoriaMetrics/issues/4018
2) create direct pod routes at vmauth with pod name routing (statefulmode is required for that at vmagent side and headless service https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) like:
 dns-name-0 = vmagent-pod-0.sts-name.service-name.svc
dns-name-1 = vmagent-pod-1.sts-name.service-name.svc
 etc

PR #1114, opened 2024-07-08
fix: add metadata.namespace in serviceaccount template

fengxsong:
When we only use `helm template`, the `metadata.namespace` field will not be rendered, which can cause GitOps sync operations to fail.


Haleygo (approved):
LGTM

Haleygo:
Thanks!

PR #1137, opened 2024-07-11
victoria-metrics-distributed: fix default multitenancy mode and disable it by default

Haleygo:
Disable multitenancy mode by default since it increases memory usage in vmagent, and most of the users not use it.
See how to enable it in https://github.com/VictoriaMetrics/helm-charts/tree/master/charts/victoria-metrics-distributed#how-to-use-multitenancy.

### DIFF
--- a/charts/victoria-metrics-distributed/README.md
+++ b/charts/victoria-metrics-distributed/README.md
@@ -71,7 +71,7 @@ And to avoid getting incomplete responses from `zone-eu-1` which gets recovered 
 
 ### How to use [multitenancy](https://docs.victoriametrics.com/cluster-victoriametrics/#multitenancy)?
 
-By default, all the data that written to `vmauth-global-write` belong to tenant `0`. To write data to different tenants, creating new tenant user in `vmauth-global-write`.    
+By default, all the data that written to `vmauth-global-write` belong to tenant `0`. To write data to different tenants, set `.Values.enableMultitenancy=true` and create new tenant users for `vmauth-global-write`.  
 For example, writing data to tenant `1088` with following steps:
 1. create tenant VMUser for vmauth `vmauth-global-write` to use:
 ```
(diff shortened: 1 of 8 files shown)


zekker6 (approved):

Haleygo:
> If `templates/per-az/vmagent yaml` hardcodes using multi-tenancy to `true`, does introducing a values.yaml enableMultiTenantMode in this PR potentially cause some confusion (because if users set it to `false`, the VMAgent will break since it will still be using multi-tenancy mode)

Yes, you're right! We must provide default suffix when multitenancy mode is disabled, fixed in latest commit.

And I disabled multitenancy mode by default now, the reason is: 
In our internal test, enable multitenancy could lead to non-trivial memory increase in vmagent, and it might be useless for most of the users.
@zekker6 please check.

charts/victoria-metrics-distributed/README.md
```diff
@@ -214,6 +214,7 @@ Change the values according to the need of the environment in ``victoria-metrics
 | availabilityZones[1].topologySpreadConstraints | list | `[{"maxSkew":1,"topologyKey":"kubernetes.io/hostname","whenUnsatisfiable":"ScheduleAnyway"}]` | topologySpreadConstraints allows to customize the default topologySpreadConstraints. |
 | availabilityZones[1].vmagent | object | `{"annotations":{},"enabled":true,"name":"","spec":{}}` | vmagent only meant to proxy write requests to each az, doesn't support customized remote write address |
 | availabilityZones[1].vmcluster.spec | object | `{"replicationFactor":2,"retentionPeriod":"14","vminsert":{"extraArgs":{},"replicaCount":2,"resources":{}},"vmselect":{"extraArgs":{},"replicaCount":2,"resources":{}},"vmstorage":{"replicaCount":2,"resources":{},"storageDataPath":"/vm-data"}}` | spec for VMCluster crd, see https://docs.victoriametrics.com/operator/api.html#vmclusterspec |
+| enableMultitenancy | bool | `true` | enable multitenancy mode see https://github.com/VictoriaMetrics/helm-charts/tree/master/charts/victoria-metrics-distributed#how-to-use-multitenancy |
```
zekker6, line 217: nit: would you mind syncing this with an updated value?
```suggestion
| enableMultitenancy | bool | `false` | enable multitenancy mode see https://github.com/VictoriaMetrics/helm-charts/tree/master/charts/victoria-metrics-distributed#how-to-use-multitenancy |
```

zekker6 (approved):

zekker6 (approved):

PR #1158, opened 2024-07-18
Vmanomaly: config validation error fix

Cambaza:
- leave empty schedulers and models section to fix aliases error
- update to 1.13.3 with fixes of Holt-Winters config validation error

### DIFF
--- a/charts/victoria-metrics-anomaly/values.yaml
+++ b/charts/victoria-metrics-anomaly/values.yaml
@@ -126,27 +126,10 @@ config:
   preset: ""
   # -- [Scheduler section](https://docs.victoriametrics.com/anomaly-detection/components/scheduler/)
   schedulers:
-    # Scheduler alias
-    1w_1m:
-      # -- Scheduler class
-      class: "periodic"
-      # -- How often to completely retrain the models. If missing value of infer_every is used and retrain on every inference run.
-      fit_every: ""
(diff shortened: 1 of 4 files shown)


charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -126,27 +126,10 @@ config:
   preset: ""
   # -- [Scheduler section](https://docs.victoriametrics.com/anomaly-detection/components/scheduler/)
   schedulers:
```
zekker6, line 128: Let's set it to an empty map to make it explicit.
```suggestion
  schedulers: {}
```

charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -126,27 +126,10 @@ config:
...
-      class: "periodic"
-      # -- How often to completely retrain the models. If missing value of infer_every is used and retrain on every inference run.
-      fit_every: ""
-      # -- What time range to use for training the models. Must be at least 1 second.
-      fit_window: ""
-      # -- How often a model will write its conclusions on newly added data. Must be at least 1 second.
-      infer_every: ""
 
   # -- [Models section](https://docs.victoriametrics.com/anomaly-detection/components/models/)
   models:
```
zekker6, line 131:
```suggestion
  models: {}
```

zekker6 (approved):

PR #1162, opened 2024-07-22
relabel __meta_kubernetes_namespace to kubernetes_namespace

0xbilko:
Hello!
The reasons why I created this PR:
1) Do the same as for `victoria-metrics-single` (https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-single/values.yaml#L556)
2) Fix an issue when the exporter passes the namespace field and this rule override it with a value from `__meta_kubernetes_namespace`. For acutally at least for [trivy-operator](https://github.com/aquasecurity/trivy-operator) and [k8s-image-availability-exporter](https://github.com/deckhouse/k8s-image-availability-exporter).

### DIFF
--- a/charts/victoria-metrics-agent/values.yaml
+++ b/charts/victoria-metrics-agent/values.yaml
@@ -447,7 +447,7 @@ config:
         - source_labels: [__meta_kubernetes_pod_container_name]
           target_label: container
         - source_labels: [__meta_kubernetes_namespace]
-          target_label: namespace
+          target_label: kubernetes_namespace
         - source_labels: [__meta_kubernetes_service_name]
           target_label: service
         - source_labels: [__meta_kubernetes_service_name]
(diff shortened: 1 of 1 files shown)


Haleygo:
Thanks for the pull request and for pointing out the inconsistency between the two charts!
But I'm not sure whether we should align the `target_label` in the victoria-metrics-vmagent chart with victoria-metrics-single or vice versa.
Based on other usages in [our repo](https://github.com/search?q=org%3AVictoriaMetrics+source_labels%3A+%5B__meta_kubernetes_namespace%5D&type=code) or [prometheus repo](https://github.com/prometheus-community/helm-charts/blob/main/charts/prometheus/values.yaml), it seems the common practice is to use
```
          - source_labels: [__meta_kubernetes_namespace]
            action: replace
            target_label: namespace
```

Haleygo:
Closing this in favor of https://github.com/VictoriaMetrics/helm-charts/pull/1210, see [this comment](https://github.com/VictoriaMetrics/helm-charts/pull/1162#issuecomment-2244779558) for reason.

>Fix an issue when the exporter passes the namespace field

@0xbilko `namespace` should be a reserved field in kubernetes, I'd suugest to fix this on the exporter side instead.

PR #1168, opened 2024-07-24
Fix false positive AlertmanagerConfigInconsistent on datasources with multiple vm clusters

dimaslv:
Count of config hashes will always be more than one if there are more than one cluster writing in current datasource.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/files/rules/generated/alertmanager.rules.yaml
+++ b/charts/victoria-metrics-k8s-stack/files/rules/generated/alertmanager.rules.yaml
@@ -84,8 +84,8 @@ rules:
     summary: 'Alertmanager instances within the same cluster have different configurations.'
   condition: '{{ true }}'
   expr: |-
-    count by (namespace,service) (
-      count_values by (namespace,service) ("config_hash", alertmanager_config_hash{job="{{ include "victoria-metrics-k8s-stack.alertmanager.name" . }}",namespace="{{ .Release.Namespace }}"})
+    count by (namespace,service,cluster) (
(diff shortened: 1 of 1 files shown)


charts/victoria-metrics-k8s-stack/files/rules/generated/alertmanager.rules.yaml
```diff
@@ -84,8 +84,8 @@ rules:
     summary: 'Alertmanager instances within the same cluster have different configurations.'
   condition: '{{ true }}'
   expr: |-
-    count by (namespace,service) (
-      count_values by (namespace,service) ("config_hash", alertmanager_config_hash{job="{{ include "victoria-metrics-k8s-stack.alertmanager.name" . }}",namespace="{{ .Release.Namespace }}"})
+    count by (namespace,service,cluster) (
```
AndrewChubatiuk, line 87: this file is managed by `hack/sync_rules.py` script, so it'll be overwritten during the next run. 
You need to update this value using `replacement_map` [here](https://github.com/VictoriaMetrics/helm-charts/blob/aa3ff6324e17d325976f228301715da214aa50f2/charts/victoria-metrics-k8s-stack/hack/sync_rules.py#L79)

dimaslv: It's appeared that cluster is our custom label from vmagent.spec.externalLabels and there is no any default label which could be used in this case, so it's not possible to distinguish alertmanagers from different clusters in this VMRule. I'm cancelling my PR for now.
Thanks.

AndrewChubatiuk: you can override expr at `defaultRules.rules.AlertmanagerConfigInconsistent.spec.expr`

dimaslv: Wow, this is a great feature! Thanks!

PR #1169, opened 2024-07-24
Fix image tag in victoria-metrics-operator/templates/uninstall_hook.yaml

dimaslv:
.Capabilities.KubeVersion.Minor returns version with plus sign, but for example there is no such tag as 1.27+ for bitnami/kubectl image (https://hub.docker.com/r/bitnami/kubectl/tags?page=&page_size=&ordering=&name=1.27%2B).


AndrewChubatiuk (approved):

Haleygo (approved):
LGTM

PR #1177, opened 2024-07-30
victoria-metrics-k8s-stack: separate alerting and recording rule common annotations and labels

AndrewChubatiuk:


charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -43,13 +43,22 @@ defaultRules:
       # -- Optional HTTP URL parameters added to each rule request
       params: {}
 
-  # -- Common properties for VMRules
   rule:
-    spec:
```
Haleygo, line 48: Shall we also reserve the function of adding common properties for both recording and alerting rules?
Otherwise, this will be a breaking change and user need to define recording&alerting common properties twice if wanted.

AndrewChubatiuk: done

Haleygo (approved):
LGTM!

Haleygo:
Thanks!

Issue #1178, opened 2024-07-30
Minor bug in vmagent helm chart clusterrole/clusterrolebinding template


guptaachin:
**Context**
We have one template for clusterrole and clusterrolebinding which also create Role and RoleBindings based on values files configs. So when `rbac.create` is set to `true`, ClusterRole and ClusterRoleBinding is rendered. 
When `namespaced` is also set to `true` along with `rbac.create` it renders `Role` and `RoleBinding`
Here are the links to templates
- ClusterRoleBinding - [link](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-agent/templates/clusterrolebinding.yaml)
- ClusterRole - [link](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-agent/templates/clusterrole.yaml)

**Problem**
On setting the rbac to render Role and RoleBindings, namespace is not rendered in `metadata`
```
rbac:
  create: true
  namespaced: true
```
This causes the helm chart to fail installation.

**Root cause**
The namespaced flag is not set right
ClusterRole - [link](https://github.com/VictoriaMetrics/helm-charts/blob/20e3142a0f079083457892e4e7051d97758915ed/charts/victoria-metrics-agent/templates/clusterrole.yaml#L7)
ClusterRoleBinding - [link](https://github.com/VictoriaMetrics/helm-charts/blob/20e3142a0f079083457892e4e7051d97758915ed/charts/victoria-metrics-agent/templates/clusterrolebinding.yaml#L7)

I will try to quickly do a PR for it.

guptaachin:
Here is the PR to fix the issue 
https://github.com/VictoriaMetrics/helm-charts/pull/1179

Haleygo:
JFYI, the fix has been released in [victoria-metrics-agent-0.10.12](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-agent-0.10.12).

PR #1179, opened 2024-07-30
Fix vmagent ClusterRole and ClusterRoleBinding to correctly render Role and RoleBinding

guptaachin:
Reported here - https://github.com/VictoriaMetrics/helm-charts/issues/1178

Issue Summary
On setting the rbac to render Role and RoleBindings, namespace is not rendered in metadata
```
rbac:
  create: true
  namespaced: true
```
This causes the helm chart to fail installation.

### DIFF
--- a/charts/victoria-metrics-agent/CHANGELOG.md
+++ b/charts/victoria-metrics-agent/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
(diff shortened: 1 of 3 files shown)


Haleygo (approved):
LGTM!

Haleygo:
Thank you for contribution!

PR #1193, opened 2024-08-01
update envs and flags documentation link

flenoir:
Hi, 

i did noticed that the link explaining how to use flags through envs vars was not working. 

I did find it in docs so i updated the ReadMe and values.yaml accordingly

regards


zekker6 (approved):
LGTM, thank you @flenoir!

Issue #1180, opened 2024-07-31
[victoria-metrics-single] serviceMonitor template doesn't have basicAuth


mdesouky:
We're deploying the `victoria-metrics-single` chart and we set `server.serviceMonitor.enabled=true` to deploy the service monitor but when it's deployed it doesn't have the basicAuth block under `spec.endpoints` which results in vmagent not being able to scrape the service with the service returning 401 errors because of failed auth.
```
2024-07-31T03:19:15.569Z	warn	VictoriaMetrics/lib/promscrape/scrapework.go:382	cannot scrape target "http://10.40.78.36:8428/metrics" ({container="victoria-metrics-single-server",endpoint="http",instance="10.40.78.36:8428",job="victoria-metrics-victoria-metrics-single-server",namespace="monitor",pod="victoria-metrics-victoria-metrics-single-server-0",service="victoria-metrics-victoria-metrics-single-server"}) 1 out of 1 times during -promscrape.suppressScrapeErrorsDelay=0s; the last error: unexpected status code returned when scraping "http://10.40.78.36:8428/metrics": 401; expecting 200; response body: "\n"
```
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  annotations:
    meta.helm.sh/release-name: victoria-metrics
    meta.helm.sh/release-namespace: monitor
  creationTimestamp: "2024-07-31T02:27:58Z"
  generation: 1
  labels:
    app: server
    app.kubernetes.io/instance: victoria-metrics
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: victoria-metrics-single
    helm.sh/chart: victoria-metrics-single-0.9.24
  name: victoria-metrics-victoria-metrics-single-server
  namespace: monitor
  resourceVersion: "49596418"
  uid: 7df4920c-dbac-46b5-8dd0-9dec83684fe3
spec:
  endpoints:
  - port: http
  namespaceSelector:
    matchNames:
    - monitor
  selector:
    matchLabels:
      app: server
      app.kubernetes.io/instance: victoria-metrics
      app.kubernetes.io/name: victoria-metrics-single
```
We have to manually edit the service monitor and add basicAuth for the service monitor to work

AndrewChubatiuk:
starting from [v0.10.0](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-single-0.10.0) it's possible to configure basic auth in ServiceMonitor

PR #1199, opened 2024-08-06
added basic auth to service monitor

AndrewChubatiuk:
Related https://github.com/VictoriaMetrics/helm-charts/issues/1180


charts/victoria-metrics-agent/templates/service-monitor.yaml
```diff
@@ -1,48 +1,44 @@
...
-  {{- if .Values.serviceMonitor.annotations }}
-  annotations:
-{{ toYaml .Values.serviceMonitor.annotations | indent 4 }}
-  {{- end }}
-  labels:
-  {{- include "chart.labels" . | nindent 4 }}
-  {{- with .Values.serviceMonitor.extraLabels -}}
-  {{ toYaml . | nindent 4 }}
+  {{- with $serviceMonitor.annotations }}
+  annotations: {{ toYaml $serviceMonitor.annotations | nindent 4 }}
```
zekker6, line 7: As far as I can see, that should be:
```suggestion
  annotations: {{ toYaml . | nindent 4 }}
```

charts/victoria-metrics-cluster/templates/vmselect-service-monitor.yaml
```diff
@@ -1,48 +1,44 @@
...
-  {{- if .Values.vmselect.serviceMonitor.annotations }}
-  annotations:
-{{ toYaml .Values.vmselect.serviceMonitor.annotations | indent 4 }}
-  {{- end }}
-  labels:
-  {{- include "victoria-metrics.vmselect.labels" . | nindent 4 }}
-  {{- if .Values.vmselect.serviceMonitor.extraLabels }}
-{{ toYaml .Values.vmselect.serviceMonitor.extraLabels | indent 4 }}
+  {{- with $serviceMonitor.annotations }}
+  annotations: {{ toYaml | nindent 4 }}
```
zekker6, line 7:
```suggestion
  annotations: {{ toYaml . | nindent 4 }}
```

zekker6 (commented):
I see `extraLabels` and `annotations` templating have also been updated, and it seems like `annotations` were not always applied properly. If that is correct could you also reflect this in a changelog?

Also see a few small fixes. It's quite hard to check such large change manually, wdyt about adding `serviceMonitor` config for values files used in CI in order to make sure there are no typos left unnoticed?

AndrewChubatiuk:
there were no fixes in annotations an extraLabels, just minor refactoring
regarding ServiceMonitor, adding it to helm chart requires to setup prometheus operator additionally

zekker6:
>  regarding ServiceMonitor, adding it to helm chart requires to setup prometheus operator additionally

What I mean is to run a "helm template" part of CI just to make sure that syntax is correct.
This is using `make lint` and `make template` make tasks separate from `Install chart` workflow.

AndrewChubatiuk:
> This is using `make lint` and `make template` make tasks separate from `Install chart` workflow.

updated

zekker6 (dismissed):

Haleygo (approved):
LGTM

Issue #520, opened 2023-04-24
[victoriametrics-k8s-stack] Make datasource name configurable or disable provisioning


D3luxee:
Hi,
i just started to look into victoria-metrics-k8s-stack as an replacement for our kube-prometheus-stack.
We have a lot of provisioned dashboards that rely on a specific name for the prometheus datasource.
At the moment it is not possible to disable the provisioning of the default datasource or change its name.

https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-k8s-stack/templates/grafana/datasource.yaml#L1

The current if clause means that i have to disable the deployment of grafana entirely or disable the collection of datasources entirely.
It would be nice if i could change the name of the default datasource and disable it without disabling grafana or the sidecar.

PR #1201, opened 2024-08-06
Allow to disable default grafana datasource

stieglma:
We have a proxy inbetween grafana and the victoria metrics prometheus endpoint. This proxy handles authentication and authorization based on forwarded oauth identities. As all datasources available in Grafana can be used by everbody, I need to be able to remove the default datasource, so that there is not direct connection to the prometheus endpoint.

This was also requested by another guy in #520

### DIFF
--- a/charts/victoria-metrics-k8s-stack/CHANGELOG.md
+++ b/charts/victoria-metrics-k8s-stack/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next release
 
 - Moved dashboards templating logic out of sync script to Helm template
+- Allow to disable default grafana datasource
 
 ## 0.24.5
 
--- a/charts/victoria-metrics-k8s-stack/README.md
+++ b/charts/victoria-metrics-k8s-stack/README.md
@@ -420,6 +420,7 @@ Change the values according to the need of the environment in ``victoria-metrics
 | grafana.ingress.path | string | `"/"` |  |
 | grafana.ingress.pathType | string | `"Prefix"` |  |
 | grafana.ingress.tls | list | `[]` |  |
+| grafana.provisionDefaultDatasource | bool | `true` |  |
 | grafana.sidecar.dashboards.additionalDashboardAnnotations | object | `{}` |  |
 | grafana.sidecar.dashboards.additionalDashboardLabels | object | `{}` |  |
 | grafana.sidecar.dashboards.enabled | bool | `true` |  |
--- a/charts/victoria-metrics-k8s-stack/templates/grafana/datasource.yaml
+++ b/charts/victoria-metrics-k8s-stack/templates/grafana/datasource.yaml
@@ -17,7 +17,7 @@ data:
   datasource.yaml: |-
     apiVersion: 1
     datasources:
-{{- if or .Values.vmsingle.enabled  .Values.vmcluster.enabled }}
+{{- if and (.Values.grafana.provisionDefaultDatasource) (or .Values.vmsingle.enabled  .Values.vmcluster.enabled) }}
     - name: VictoriaMetrics
       type: {{ default "prometheus" .Values.grafana.defaultDatasourceType }}
       {{- $readEndpoint:= (include "victoria-metrics-k8s-stack.vmReadEndpoint" . | fromYaml) }}
--- a/charts/victoria-metrics-k8s-stack/values.yaml
+++ b/charts/victoria-metrics-k8s-stack/values.yaml
@@ -774,6 +774,8 @@ grafana:
 
   ## ForceDeployDatasource Create datasource configmap even if grafana deployment has been disabled
   forceDeployDatasource: false
+  # Set to false to disable the default datasource connected directly to victoria-metrics
+  provisionDefaultDatasource: true
 
   ## Configure additional grafana datasources (passed through tpl)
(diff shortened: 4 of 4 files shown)


AndrewChubatiuk:
hey @stieglma 
please add description of your change to a chart's changelog

Haleygo:
Hi @stieglma,
Is the [victoria-metrics-k8s-stack.vmReadEndpoint](https://github.com/VictoriaMetrics/helm-charts/blob/fix-operator-role/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl#L90) not available for you, or you want to set other [additionalDataSources](https://github.com/VictoriaMetrics/helm-charts/blob/fix-operator-role/charts/victoria-metrics-k8s-stack/values.yaml#L780) as the default one?
If it's the latter, I'd suggest to add option to allow set datasource `VictoriaMetrics` as non-default, like
```
# values.yaml
grafana:
  enabled: true
  ## all values for grafana helm chart can be specified here
  sidecar:
    datasources:
      enabled: true
      initDatasources: true
      createVMReplicasDatasources: false
      useVMAsDefaultDatasource: true

#/templates/grafana/datasource.yaml
{{- if or .Values.vmsingle.enabled  .Values.vmcluster.enabled }}
    - name: VictoriaMetrics
      type: {{ default "prometheus" .Values.grafana.defaultDatasourceType }}
      {{- $readEndpoint:= (include "victoria-metrics-k8s-stack.vmReadEndpoint" . | fromYaml) }}
      url: {{ $readEndpoint.url }}
      access: proxy
      isDefault:  {{ .Values.grafana.sidecar.datasources.useVMAsDefaultDatasource }}
      jsonData: {{ toYaml .Values.grafana.sidecar.datasources.jsonData | nindent 8 }}
```
this way, user can still create VictoriaMetrics datasource and use it when needed.

Haleygo:
Maybe we should reuse the [.Values.externalVM.read](https://github.com/VictoriaMetrics/helm-charts/blob/d26e12f52d97c6eb4cd9493b9f61b0b3054c6a22/charts/victoria-metrics-k8s-stack/values.yaml#L220), to make it prior to the default VM endpoints in [victoria-metrics-k8s-stack.vmReadEndpoint](https://github.com/VictoriaMetrics/helm-charts/blob/d26e12f52d97c6eb4cd9493b9f61b0b3054c6a22/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl#L90)
```
{{- define "victoria-metrics-k8s-stack.vmReadEndpoint" -}}
{{- if .Values.externalVM.read.url -}}
{{ .Values.externalVM.read | toYaml }}
{{- else if .Values.vmsingle.enabled -}}
url: {{ printf "http://%s.%s.svc:%s%s" (include "victoria-metrics-k8s-stack.vmsingleName" .) .Release.Namespace (.Values.vmsingle.spec.port | default "8429") (index .Values "vmsingle" "spec" "extraArgs" "http.pathPrefix" | default "/") }}
{{- else if .Values.vmcluster.enabled -}}
url: {{ printf "http://%s.%s.svc:%s%s/select/%s/prometheus" (include "victoria-metrics-k8s-stack.vmselectName" .) .Release.Namespace (.Values.vmcluster.spec.vmselect.port | default "8481") (index .Values "vmcluster" "spec" "vmselect" "extraArgs" "http.pathPrefix" | default "") (.Values.tenant | default "0") }}
{{- end }}
{{- end }}
```
So user can modify the endpoint with vmsingle or vmcluster deployed.
It will also change the [datasource and remoteRead](https://github.com/VictoriaMetrics/helm-charts/blob/d26e12f52d97c6eb4cd9493b9f61b0b3054c6a22/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl#L120) endpoints of vmalert, and I think it's ok.
wdyt @AndrewChubatiuk @zekker6

stieglma:
I'm not sure if this is helping everywhere, for example:
1. the use case from #520 would still not work, as they need their datasource to either have a different name or be disabled, so they can create their own datasource as a default datasource with a specific name
2. if we change the url via the template function I'm not sure about the consequences for alertmanager, e.g. our auth-proxy that is used via grafana, requires the oauth headers to be forwarded from grafana to the proxy. I doubt that this is something alertmanager does, or where it even makes sense, alertmanager should just have direct access to the underlying datasource
3. we even don't "run" the auth proxy as a separate deployment, it is a sidecar in grafana and we therefore reference the endpoint via localhost:port instead of a service. So this would even not be visible for alertmanager if the authentication was no issue

Haleygo:
>I'm not sure if this is helping everywhere

Yeah, it doesn't solve all the issues. And I'm not against allowing disable default VM datasource as this pull request does. 
I just want to utilize our existing options, to see if they can be beneficial for more cases. 
If users add an extra simple proxy in front of vmsingle/vmcluster, using `.Values.externalVM.read` as the default [vmReadEndpoint](https://github.com/VictoriaMetrics/helm-charts/blob/bbe961de756397a3cdd1bfbedd161de2c590a3b2/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl#L90) seems reasonable to me. The other option would be disable the default VM datasource totally and add new one under `.Values.grafana.additionalDataSources`.

zekker6 (approved):

PR #509, opened 2023-04-04
Enable latest etcd dashboard, use jsonnet mixin

kvaster:
Update etcd rules to latest, add etcd dashboard using latest source - jsonnext mixin.

PR #1209, opened 2024-08-08
support jsonnet import

AndrewChubatiuk:
replacement for #509

### DIFF
--- a/charts/victoria-metrics-k8s-stack/CHANGELOG.md
+++ b/charts/victoria-metrics-k8s-stack/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Moved dashboards templating logic out of sync script to Helm template
 - Allow to disable default grafana datasource
+- Synchronize Etcd dashboards and rules with mixin provided by Etcd
 
 ## 0.24.5
 
--- a/charts/victoria-metrics-k8s-stack/hack/sync_rules.py
+++ b/charts/victoria-metrics-k8s-stack/hack/sync_rules.py
@@ -2,17 +2,22 @@
 """Fetch alerting and aggregation rules from provided urls into this chart."""
 from os import makedirs
 from os.path import realpath, dirname, join
+from urllib.parse import urlparse
 
+import json
+import pathlib
+import _jsonnet
 import requests
 import yaml
 from yaml.representer import SafeRepresenter
-import re
 
-rulesDir = join(dirname(realpath(__file__)), '..', 'files', 'rules', 'generated')
+rulesDir = join(dirname(realpath(__file__)), "..", "files", "rules", "generated")
+
 
 # https://stackoverflow.com/a/20863889/961092
 class literal(str):
     pass
+
 
 class quoted(str):
     pass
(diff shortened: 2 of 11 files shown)


charts/victoria-metrics-k8s-stack/hack/sync_rules.py
```diff
@@ -158,33 +178,70 @@ def write_group_to_file(group, url, destination):
...
 
+def jsonnet_import(directory, file_name):
+    src = join(directory, file_name)
+    if file_name.startswith("github.com"):
+        parsed = urlparse("https://%s" % file_name).path.split("/", 3)
+        src = "https://raw.githubusercontent.com/%s/%s/master/%s" % (
+            parsed[1],
+            parsed[2],
+            parsed[3],
+        )
```
zekker6, line 195: Could you add a comment to clarify which cases are fixed by this?
As far as I can see, it is most likely related to jsonnet being able to reference other files in github repo directly, but it would be great to get this noted.

Also, is it ok to hardcode `master` branch here?
Maybe it is possible to fetch original branch name from the `file_name` in order to avoid issues in case branch will be renamed in upstream.

charts/victoria-metrics-k8s-stack/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Next release
 
 - Moved dashboards templating logic out of sync script to Helm template
+- Support importing jsonnet dashboards and rules
```
zekker6, line 6: I would suggest to mention the change which is useful for users:
```suggestion
- Synchronize Etcd dashboards and rules with mixin provided by Etcd.
```

The fact that it is now possible to import `.jsonnet` is most likely not very useful to the end users, but only to the chart developers.

zekker6 (approved):
LGTM in general, see a few comments.

This PR also introduces code style changes like using single or double quotes, spacing around arguments and similar ones. What do you think about adding a linter for formatting for these scripts? (In a separate PR)
That would help to enforce some code style once and ensure that it will be followed in future changes as well.

charts/victoria-metrics-k8s-stack/hack/sync_rules.py
AndrewChubatiuk: jsonnet has remote data directives in a format, like in go (check [here](https://github.com/etcd-io/etcd/blob/main/contrib/mixin/dashboards/g.libsonnet)). this code invoked by jsonnet on each met import and converts imports to exact data from remote sources (jsonnet lib does not perform it by default)

AndrewChubatiuk:
> What do you think about adding a linter for formatting for these scripts?
added in this PR

charts/victoria-metrics-k8s-stack/hack/sync_rules.py
AndrewChubatiuk: > Also, is it ok to hardcode master branch here?
Added api call to get a default branch name

zekker6 (approved):

PR #1212, opened 2024-08-09
added ability to override probes

AndrewChubatiuk:
- replacement for #757. To change a probe type old type should be set to `null`, e.g: 
   for a default probe config:
   ```
   probe:
      liveness:
         httpGet:
            path: /health
            port: http
   ```
   to change probe type to `tcpSocket` `httpGet` should be set to `null`
   ```
   probe:
      liveness:
         httpGet: null
         tcpSocket:
            port: http
   ```
- fixed multiple `imagePullSecrets` and `env` sections indent
- minor refactoring
- introduced victoria-metrics-common library chart


charts/victoria-logs-single/values.yaml
```diff
@@ -150,35 +150,35 @@ server:
...
-    #periodSeconds: 15
-    #successThreshold: 1
-    #timeoutSeconds: 5
+  probe:
+    # Indicates whether the Container is ready to service requests. If the readiness probe fails, the endpoints controller removes the Pod's IP address from the endpoints of all Services that match the Pod. The default state of readiness before the initial delay is Failure. If a Container does not provide a readiness probe, the default state is Success.
+    readiness:
+      httpGet:
+        path: '{{ include "vm.probe.http.path" . }}'
+        scheme: '{{ include "vm.probe.http.scheme" . }}'
+        port: '{{ include "vm.probe.port" . }}'
```
Haleygo, line 159: The tpl doesn't seem to work in my test:
```
➜  victoria-logs-single git:(vmcluster-override-probes) ✗ helm template .
Error: template: victoria-logs-single/templates/server-statefulset.yaml:61:30: executing "victoria-logs-single/templates/server-statefulset.yaml" at <include "vm.probe" (dict "app" $app "type" "readiness")>: error calling include: template: victoria-logs-single/templates/_helpers.tpl:155:4: executing "vm.probe" at <tpl (toYaml $probe) .>: error calling tpl: cannot retrieve Template.Basepath from values inside tpl function: failureThreshold: 3
httpGet:
  path: '{{ include "vm.probe.http.path" . }}'
  port: '{{ include "vm.probe.port" . }}'
  scheme: '{{ include "vm.probe.http.scheme" . }}'
initialDelaySeconds: 5
periodSeconds: 15
timeoutSeconds: 5: "BasePath" is not a value

Use --debug flag to render out invalid YAML
```

charts/victoria-logs-single/values.yaml
```diff
@@ -150,35 +150,35 @@ server:
...
-
-  # Indicates whether the Container is done with potentially costly initialization. If set it is executed first. If it fails Container is restarted. If it succeeds liveness and readiness probes takes over.
-  startupProbe: {}
-    #tcpSocket:
-    #  port: http
-    #failureThreshold: 30
-    #periodSeconds: 15
-    #successThreshold: 1
-    #timeoutSeconds: 5
+  probe:
```
Haleygo, line 153: Can we not adding extra `probe:` but using the old `readinessProbe`, `livenessProbe` fields?
So this doesn't become a breaking change.

charts/victoria-logs-single/values.yaml
AndrewChubatiuk: are you trying to run it from my branch?

Haleygo: yeah, I pulled the latest

charts/victoria-logs-single/values.yaml
AndrewChubatiuk: want to make it consistent across all charts. I don't care if it should be `probe.readiness` or `readinessProbe`, but I would like to have same option across all charts, instead of these two, that we're having at the moment

charts/victoria-logs-single/values.yaml
AndrewChubatiuk: could you please upgrade helm locally?

charts/victoria-metrics-agent/templates/podsecuritypolicy.yaml
```diff
@@ -3,17 +3,14 @@ apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
```
Haleygo, line 3: I think we can remove PodSecurityPolicy from our charts totally. It was removed in k8s v1.25, and we claimed to support >v1.27 recently.
wdyt?

charts/victoria-logs-single/values.yaml
AndrewChubatiuk: there were issues with old versions on CI as well

Haleygo: Hmm, upgrading from helm v3.12.3 to v3.15.3 solves the error, should related to https://github.com/helm/helm/issues/11168#issuecomment-1935567803.
We might need to clarify the helm version somewhere.

charts/victoria-logs-single/values.yaml
Haleygo: okay, thanks

charts/victoria-metrics-agent/templates/podsecuritypolicy.yaml
AndrewChubatiuk: removed

charts/victoria-logs-single/values.yaml
AndrewChubatiuk: added to requirements.md

charts/victoria-logs-single/Chart.yaml
```diff
@@ -25,9 +25,11 @@ annotations:
...
-  version: "0.46.7"
-  repository: "https://fluent.github.io/helm-charts"
-  condition: fluent-bit.enabled
+  - name: fluent-bit
+    version: "0.46.7"
+    repository: "https://fluent.github.io/helm-charts"
+    condition: fluent-bit.enabled
+  - name: victoria-metrics-common
+    version: "0.0.*"
+    repository: file://../victoria-metrics-common
```
zekker6, line 35: Reference to local file system will not work for external users without cloning a repo.
It is needed to publish `common` chart first and then add a reference with proper `repository` + `version` here.

.github/workflows/run-testing.yaml
```diff
@@ -75,10 +79,11 @@ jobs:
...
         uses: helm/chart-testing-action@v2.6.1
 
-      - name: Create k3d cluster
-        uses: nolar/setup-k3d-k3s@v1
+      - name: Create Kind cluster
+        uses: helm/kind-action@v1
         with:
-          version: ${{ matrix.k8s_version }}
+          node_image: kindest/node:${{ matrix.k8s_version }}
+          kubectl_version: ${{ matrix.k8s_version }}
```
zekker6, line 86: What is the reason to switch from one to another?

.github/workflows/run-lint.yaml
```diff
@@ -6,11 +6,15 @@ on:
...
       - '**.md'
+    branches:
+      - master
   pull_request:
     paths-ignore:
       - 'packages/**'
       - 'index.yaml'
       - '**.md'
+    branches:
+      - master
```
zekker6, line 17: Why limiting this to PR and pushes to master?
Github Actions in public repos are free of charge, and it might be more convenient to see CI results before even opening a PR when working on something. Seems to me like it is fine to run it on all commits.

.github/workflows/run-testing.yaml
AndrewChubatiuk: nolar/setup-k3d-k3s@v1 action fails frequently

.github/workflows/run-lint.yaml
AndrewChubatiuk: previous configuration produced redundant pipelines (for branch and for PR), this change leaves pull requests pipelines, cause branches filter filters a base branch and disables branch pipelines for non-master branches.

charts/victoria-logs-single/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: v0.28.0-victorialogs
+appVersion: v0.28.0
```
zekker6, line 2: VictoriaLogs releases have `-victorialogs` suffix in published docker images, see: https://hub.docker.com/r/victoriametrics/victoria-logs/tags?page=&page_size=&ordering=&name=v0.28.0

AndrewChubatiuk: using common template for image name generation, victorialogs was moved to `image.variant`, how it's done in the rest of the charts

charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -7,10 +7,12 @@ global:
...
       
 image:
+  # -- Victoria Metrics anomaly Docker registry
+  registry: ""
   # -- Victoria Metrics anomaly Docker repository and image name
-  repository: docker.io/victoriametrics/vmanomaly
+  repository: victoriametrics/vmanomaly
   # -- Tag of Docker image
-  tag: "" # rewrites Chart.AppVersion
+  tag: v1.13.3 # rewrites Chart.AppVersion
```
zekker6, line 15: As far as I can see, charts are not specifying `.image.tag` anymore and using `.Chart.AppVersion` instead. Is this an intentional change?

AndrewChubatiuk: vmanomaly tagging policy differs from other apps. despite it's an enterprise product, but it has no `-enterprise` suffix in a tag name. I've recently added a PR to vmanomaly repo to fix it
https://github.com/VictoriaMetrics/vmanomaly/pull/243

zekker6: Thanks for clarification. Could you update a tag to `v1.15.4` then?
That's the latest version based on [changelog](https://docs.victoriametrics.com/anomaly-detection/changelog/)

AndrewChubatiuk: [code in a common template](https://github.com/VictoriaMetrics/helm-charts/blob/0735f3ce0e7a74fa384d0d93bb3123a55cec8b16/charts/victoria-metrics-common/templates/_image.tpl#L7) checks if `image.tag` wasn't set and for enterprise products (which have license.key, license.secret.name or eula) it sets `image.variant` to `enterprise`

zekker6 (approved):

zekker6 (approved):

PR #1215, opened 2024-08-10
fixed exact rule annotation propagation to other rules

### LINKED ISSUE #1214: Overriding rules will apply to all rules in group and type
So if you add an annotation to an exact rule, it will apply to all rules in the group and even type (recording/alert).

Seems that `mergeOverwrite` in Helm will mutate the underlying Go maps. So all arguments needs a `deepCopy`:

https://github.com/VictoriaMetrics/helm-charts/blob/4040c7e6f9a134972535fb1fdf9c973a2dfb53be/charts/victoria-metrics-k8s-stack/templates/rules/rule.yaml#L88

AndrewChubatiuk:
FIxes #1214


Haleygo (approved):
LGTM

PR #1221, opened 2024-08-14
charts/victoria-metrics-anomaly: update version

zekker6:
- update vmanomaly to 1.15.3
- update configuration examples to match new version


fred-navruzov (approved):
thanks, LGTM!

PR #1233, opened 2024-08-15
use static container names

AndrewChubatiuk:
removed tempating from container names to simplify debugging

### DIFF
--- a/charts/victoria-logs-single/CHANGELOG.md
+++ b/charts/victoria-logs-single/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next release
 
 **Update note**: The VictoriaMetrics components image tag template has been updated. This change introduces `.Values.<component>.image.variant` to specify tag suffixes like `-scratch`, `-cluster`, `-enterprise`. Additionally, you can now omit `.Values.<component>.image.tag` to automatically use the version specified in `.Chart.AppVersion`.
+**Update note**: main container name was changed to `vlogs`, which will recreate a pod.
 
 - Added basicAuth support for ServiceMonitor
 - Set minimal kubernetes version to 1.25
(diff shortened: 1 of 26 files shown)


zekker6 (commented):
Should we maybe use static container names related to the component name?
AFAIK, there are cases when container name is used for targets relabeling to drop all containers apart from the main one (e.g. there is oauth2-proxy running next to the main component and relabeling is used to drop oauth2-proxy as a target).

Also, this is going to be a breaking change for such use-cases so it would be great to add this to "upgrade notes" secton so that it is easier to find.

AndrewChubatiuk:
@zekker6 
> related to the component name
e.g: in victoria-metrics-single
```
- name: single
```
?
for a case, you've mentioned, looks like it's easier to have the same main container name across all charts

Haleygo:
I agree with @zekker6 to keep the names related to components, like vmsingle, vmagent, alertmanager...
I don't see how different names complicate debugging, but can be more clear to see and understand the container without correlating with pod. 
e.g.`kube_pod_container_status_ready{container="alertmanager",endpoint="http",instance="10.71.2.9:8080",job="kube-state-metrics",namespace="monitoring",pod="vmalertmanager-main-0"}`

AndrewChubatiuk:
@Haleygo @zekker6 updated container names

zekker6 (approved):

charts/victoria-logs-single/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Next release
 
 **Update note**: The VictoriaMetrics components image tag template has been updated. This change introduces `.Values.<component>.image.variant` to specify tag suffixes like `-scratch`, `-cluster`, `-enterprise`. Additionally, you can now omit `.Values.<component>.image.tag` to automatically use the version specified in `.Chart.AppVersion`.
+**Update note**: main container name was changed to `vlogs`
```
Haleygo, line 6: It would be nice to specify that this will recreate the pod.
> change main container name to `vlogs`, which will recreate the pod.

Issue #1244, opened 2024-08-17
Can't set vmagent resources in chart values.yaml


yunzck8s:
```
vmagent:
  enabled: true
  annotations: {}
  # https://docs.victoriametrics.com/operator/api.html#vmagentremotewritespec
  # defined spec will be added to the remoteWrite configuration of VMAgent
  additionalRemoteWrites:
    []
    #- url: http://some-remote-write/api/v1/write
  # spec for VMAgent crd
  # https://docs.victoriametrics.com/operator/api.html#vmagentspec
  spec:
    selectAllByDefault: true
    image:
      tag: v1.102.1
    scrapeInterval: 20s
    externalLabels: {}
      # For multi-cluster setups it is useful to use "cluster" label to identify the metrics source.
      # For example:
      # cluster: cluster-name
    extraArgs:
      promscrape.streamParse: "true"
      # Do not store original labels in vmagent's memory by default. This reduces the amount of memory used by vmagent
      # but makes vmagent debugging UI less informative. See: https://docs.victoriametrics.com/vmagent/#relabel-debug
      promscrape.dropOriginalLabels: "true"
  ingress:
    enabled: false
    # For Kubernetes >= 1.18 you should specify the ingress-controller via the field ingressClassName
    # See https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#specifying-the-class-of-an-ingress
    # ingressClassName: nginx
    # Values can be templated
    annotations:
      {}
      # kubernetes.io/ingress.class: nginx
      # kubernetes.io/tls-acme: "true"
    labels: {}
    path: /
    # pathType is only for k8s > 1.19
    pathType: Prefix

    hosts:
      - vmagent.domain.com
    ## Extra paths to prepend to every host configuration. This is useful when working with annotation based services.
    extraPaths: []
    # - path: /*
    #   backend:
    #     serviceName: ssl-redirect
    #     servicePort: use-annotation
    ## Or for k8s > 1.19
    # - path: /*
    #   pathType: Prefix
    #   backend:
    #     service:
    #       name: ssl-redirect
    #       port:
    #         name: service
    tls: []
    #  - secretName: vmagent-ingress-tls
    #    hosts:
    #      - vmagent.domain.com
```

I want to add a resource field here and set the agent's resources. Now the agent's cpu resource triggers an early warning and I cannot adjust it through values.

AndrewChubatiuk:
[...]

Issue #255, opened 2021-11-09
Add more description about the parameters


woodliu:
Hi, could you add more description about the parameters? There is so few infromation about them, we are not sure how use them...
![image](https://user-images.githubusercontent.com/9976943/140851197-8af8a5ae-1129-4354-bc04-bac19edba9a6.png)

AndrewChubatiuk:
in PR #1245 updated documentation, grouped sections with default configuration together to provide a link to a full full list of allowed arguments

PR #1245, opened 2024-08-17
Added spec params description

AndrewChubatiuk:
related issues #1244 #255
- expose spec urls in params description
- render param object in REAMDE as yaml


Haleygo (commented):
hmm, I'm not sure about using HTML tags, they make README better but the values.yaml is a bit messy.

AndrewChubatiuk:
I would also like to keep markdown tables instead of html ones, but multiline content breaks markdown tables and markdown is not rendered inside html. I can add to hugo-docs a PR, which adds function, which allow markdown to html conversion

AndrewChubatiuk:
removed HTML tags from values and and rendered documentation by helm-docs binary from [this](https://github.com/norwoodj/helm-docs/pull/262) PR

AndrewChubatiuk:
@Haleygo
Since there's no activity in helm-docs repo for a month, so I've built a docker image with my fork
could you please review it again?

Haleygo:
> `#` -- Extra Volume Mounts for the container
  extraVolumeMounts:
    []
    # - name: example
    #   mountPath: /example

A minor question: is there a way to support value's comment without adding the prefix `# --`? 
I'm just concerned that I or other contributors might forget to add this for new values in the future.

Haleygo (approved):
LGTM

PR #1259, opened 2024-08-20
support both service port names and numbers in ingress

AndrewChubatiuk:
followup for https://github.com/VictoriaMetrics/helm-charts/pull/1257

### DIFF
--- a/charts/victoria-logs-single/templates/server-ingress.yaml
+++ b/charts/victoria-logs-single/templates/server-ingress.yaml
@@ -20,6 +20,11 @@ spec:
   {{- end }}
   rules:
     {{- range $app.ingress.hosts }}
+    {{- $port := dict "name" "http" }}
+    {{- with .port }}
+    {{- $numberTypes := list "int" "float64" }}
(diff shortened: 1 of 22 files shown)


charts/victoria-logs-single/templates/server-ingress.yaml
```diff
@@ -20,6 +20,11 @@ spec:
   {{- end }}
   rules:
     {{- range $app.ingress.hosts }}
+    {{- $port := dict "name" "http" }}
+    {{- with .port }}
+    {{- $numberTypes := list "int" "float64" }}
+    {{- $port = dict (ternary "number" "name" (has (kindOf .) $numberTypes)) . }}
+    {{- end }}
```
zekker6, line 27: Seems like this part is quite repetitive, and we now have `common` chart. Would that make sense to move it there, wdyt?

AndrewChubatiuk: moved logic to a common chart, added values for chart's ingress linting

PR #1272, opened 2024-08-21
operator: generate webhook certs without cert-manager, enable webhook by default

AndrewChubatiuk:


f41gh7 (approved):
LGTM

f41gh7:
Thanks!

farcaller:
This broke the cert-manager setup, because the webhook options ended in Issuer/victoria-metrics-operator-root. Can you please make sure this renders properly for `admissionWebhooks.certManager.enabled = true`?

farcaller:
I think it's the incorrect resource split; the suggested fix is in https://github.com/VictoriaMetrics/helm-charts/pull/1315.

Isn't helm fun?

Issue #1287, opened 2024-08-22
Broken dependencies in vmauth chart 0.5.0


cwstrommer:
The chart includes `victoria-metrics-common` as a relative file dependency, as follows:
```
dependencies:
- name: victoria-metrics-common
  repository: file://../victoria-metrics-common
  version: 0.0.*
```
However, this breaks things when doing the following:
```
helm fetch vm/victoria-metics-auth
tar xzf victoria-metrics-auth-0.5.0.tgz
cd victoria-metrics-auth/
helm dep up
```
The dependency should instead be something like
```
dependencies:
- name: victoria-metrics-common
  repository: https://victoriametrics.github.io/helm-charts/
  version: 0.0.*
```

AndrewChubatiuk:
hey @cwstrommer
what is your usecase?
when this upgrade happens?

cwstrommer:
@AndrewChubatiuk we are required for regulatory reasons to perform a SAST scan against all charts we use.  Effectively we do a `helm fetch`, `helm dep build/up`, and `helm template` to render the chart, and then scan the resulting manifests before caching the chart in our private registry (since we can then attest that all charts from our registry have been scanned).

I would also expect that user using gitops tools like argocd or flux might run into similar issues, since their behaviour is similar to our process.

One thing that I have seen other developers do with dependencies that have file paths is to package these dependencies when creating the chart artifact.  In other words their CI pipeline does the `helm dep up`, and stores both `Chart.lock` (which I see vmauth already does) and the `charts` subdirectory.  This would be acceptable to us, because we do `helm dep build` instead of `helm dep up` when the chart has a `Chart.lock` file.

PR #1289, opened 2024-08-22
publish common chart

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1287

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -88,7 +88,9 @@ jobs:
                 echo "" >> $RN_FILE
                 echo "**Release date:** $(date -I)" >> $RN_FILE
                 echo "" >> $RN_FILE
-                echo "![AppVersion: $APP_VERSION](https://img.shields.io/static/v1?label=AppVersion&message=$APP_VERSION&color=success&logo=)" >> $RN_FILE
+                if [[ $APP_VERSION != "null" ]]; then
+                  echo "![AppVersion: $APP_VERSION](https://img.shields.io/static/v1?label=AppVersion&message=$APP_VERSION&color=success&logo=)" >> $RN_FILE
+                fi
                 echo "![Helm: v3](https://img.shields.io/static/v1?label=Helm&message=v3&color=informational&logo=helm)" >> $RN_FILE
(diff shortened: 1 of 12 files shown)


charts/victoria-logs-single/Chart.lock
```diff
@@ -4,6 +4,6 @@ dependencies:
   version: 0.46.7
 - name: victoria-metrics-common
   repository: file://../victoria-metrics-common
```
zekker6, line 6: Based on an issue related to this PR it seems like this link should also be updated (in `Chart.yaml`) to point to `https://victoriametrics.github.io/helm-charts/`, otherwise an error user is seeing will be the same. Could you update the link?

AndrewChubatiuk: added patching before release

.github/workflows/release.yaml
```diff
@@ -82,13 +83,19 @@ jobs:
                 # ArtifactHub annotation in Chart.yaml file
                 export CHANGELOG
                 yq -M -i '.annotations["artifacthub.io/changes"] = (load_str(strenv(CHANGELOG)) | capture("## Next release[\s]*(?P<changes>[\s\S]*?)[\s]*##") | .changes | split("\n") | .[] |= sub("^[\s]*-[\s]+", "") | filter(length > 0) | to_yaml)' $CHART_FILE
+                yq -M -i '(.dependencies[] | select(.name == "victoria-metrics-common") | .repository) = "https://victoriametrics.github.io/helm-charts"' $CHART_FILE
+                if [ -e $CHART_LOCK ]; then
+                  yq -M -i '(.dependencies[] | select(.name == "victoria-metrics-common") | .repository) = "https://victoriametrics.github.io/helm-charts"' $CHART_LOCK
+                fi
```
zekker6, line 89: It is seems like a `helm dep update` is needed after replacing repository URL. Otherwise, helm will throw a `digest` mismatch error.

AndrewChubatiuk: moved common chart dependency update to another PR, will create it as soon as changes from this PR merged and published

zekker6 (approved):

Issue #207, opened 2021-09-24
victoria-metrics-single: add config support for relabelConfig


belm0:
the chart values should support a global relabel config for ingested metrics, similar to the `scrape` variable (generating a configMap, etc.)

https://docs.victoriametrics.com/#relabeling

currently we're using `--relabelConfig` and a separately deployed file, but I'd like a coherent deployment including a configMap for relabel config

(Note that the victoria-metrics-agent chart supports relabel configmap, but victoria-metrics-single does not.)

AndrewChubatiuk:
global relabel config setting is available in release [0.11.0](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-single-0.11.0)

PR #1290, opened 2024-08-23
added global relabel config

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/207


charts/victoria-metrics-single/values.yaml
```diff
@@ -372,6 +372,12 @@ server:
     # -- Service Monitor metricRelabelings
     metricRelabelings: []
 
+  # -- Global relabel configuration
+  relabel:
+    enabled: false
+    config: []
+    configMap: ""
```
zekker6, line 379: WDYT about either making this:

```
configMap:
  name: ""
  key: ""
```

Or adding a comment that key is expected to be a `relabel.yml` for this to work?

AndrewChubatiuk: done

zekker6 (approved):

PR #1306, opened 2024-08-23
moved templates to a common chart

AndrewChubatiuk:
removed identical templates from all charts, that are already present in a common chart


f41gh7 (approved):
LGTM

PR #1325, opened 2024-08-26
charts/victoria-metrics-operator: fix CRD naming in RBAC

zekker6:
Follow-up for https://github.com/VictoriaMetrics/helm-charts/pull/1310


AndrewChubatiuk:
@zekker6 why not just rename everything to a plural form?

zekker6:
@AndrewChubatiuk These values are also used in order to generate entries for `ValidatingWebhookConfiguration` and it uses singular form in order to define resource `name`.

PR #1326, opened 2024-08-26
removed eula support

AndrewChubatiuk:
- remove eula flag support
- updated `vm.args` template structure to embed all args in to a dict, which allows to unmarshal values later


zekker6 (commented):
I would suggest to stop support of eula flag in helm charts.
It was deprecated in VictoriaMetrics a while ago and will be completely removed soon.

AndrewChubatiuk:
dropped eula support from everywhere

zekker6 (approved):

zekker6 (approved):

PR #1348, opened 2024-08-27
vmanomaly: add pull port to podMonitor condition

AndrewChubatiuk:


charts/victoria-metrics-anomaly/templates/pod-monitor.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.podMonitor.enabled -}}
+{{- if and .Values.podMonitor.enabled (((.Values.config).monitoring).pull).port -}}
```
zekker6, line 1: Maybe it's better to throw an error in this case?
While disabling `PodMonitor` will make sure there are no scrape errors it will make it harder for user to find an error. Explicitly failing here will give user more feedback and enforce using correct configuration.

AndrewChubatiuk: changed this. also added a check for `monitoring.pull.enabled property`

zekker6 (approved):

charts/victoria-metrics-anomaly/Chart.yaml
```diff
@@ -3,7 +3,7 @@ type: application
 name: victoria-metrics-anomaly
 description: Victoria Metrics Anomaly Detection - a service that continuously scans Victoria Metrics time series and detects unexpected changes within data patterns in real-time.
 version: 1.4.2
-appVersion: v1.15.6
+appVersion: v1.15.9
```
zekker6, line 6: Could you also add a changelog entry for `appVersion` change?

AndrewChubatiuk: done

zekker6 (dismissed):

zekker6 (approved):

PR #1356, opened 2024-08-28
common: added more generic template functions

AndrewChubatiuk:
- Added `vm.service` for unified service name generation
- Added `vm.url` to construct service DNS name
- Added `vm.name` for chart name
- Added `vm.fullname` which is actively used in resource name construction
- Added `vm.chart` to construct chart name label value
- Added `vm.labels` for common labels
- Added `vm.sa` for service account name
- Added `vm.release` for release name
- Added `vm.selectorLabels` for common selector labels

### DIFF
--- a/charts/victoria-metrics-common/templates/_helpers.tpl
+++ b/charts/victoria-metrics-common/templates/_helpers.tpl
@@ -1,0 +1,59 @@
+{{- /* Expand the name of the chart. */ -}}
+{{- define "vm.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{- /*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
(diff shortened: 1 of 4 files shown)


charts/victoria-metrics-common/templates/_helpers.tpl
```diff
@@ -0,0 +1,59 @@
...
+{{- end }}
+{{- end }}
+
+{{- /* Create chart name and version as used by the chart label. */ -}}
+{{- define "vm.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{- /* Create the name of the service account to use */ -}}
+{{- define "vm.sa" -}}
```
zekker6, line 30:
```suggestion
{{- define "vm.sa.name" -}}
```

Let's add a `name` suffix as otherwise it seems like this template will render full SA config.

AndrewChubatiuk: thanks for review, will recreate this PR later, after making these templates more generic

PR #1358, opened 2024-08-28
common: added more templates

AndrewChubatiuk:
- Added `vm.service` to construct vmoperator-managed service name
- Added `vm.url` to construct vmoperator-managed service base url
- Added `vm.name` for chart name
- Added `vm.fullname` which is actively used in resource name construction
- Added `vm.chart` to construct chart name label value
- Added `vm.labels` for common labels
- Added `vm.sa.name` for service account name
- Added `vm.metaLabels` for compatibility with most of the charts
- Added `vm.release` for release name
- Added `vm.selectorLabels` for common selector labels
- Expect built-in variables in both root and `.helm` context

### DIFF
--- a/charts/victoria-metrics-common/templates/_enterprise.tpl
+++ b/charts/victoria-metrics-common/templates/_enterprise.tpl
@@ -1,13 +1,16 @@
 {{- define "vm.license.secret.key" -}}
-{{- ((.Values.license).secret).key | default (((.Values.global).license).secret).key | default "" -}}
+  {{- $Values := (.helm).Values | default .Values -}}
+  {{- (($Values.license).secret).key | default ((($Values.global).license).secret).key | default "" -}}
 {{- end -}}
 
 {{- define "vm.license.secret.name" -}}
-{{- ((.Values.license).secret).name | default (((.Values.global).license).secret).name | default "" -}}
+  {{- $Values := (.helm).Values | default .Values -}}
+  {{- (($Values.license).secret).name | default ((($Values.global).license).secret).name | default "" -}}
 {{- end -}}
 
 {{- define "vm.license.key" -}}
-{{- (.Values.license).key | default ((.Values.global).license).key | default "" -}}
+  {{- $Values := (.helm).Values | default .Values }}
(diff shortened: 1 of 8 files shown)


charts/victoria-metrics-common/templates/_enterprise.tpl
```diff
@@ -27,38 +30,36 @@
...
-{{- $licenseSecretName := (include "vm.license.secret.name" .) -}}
-{{- if and $licenseSecretName $licenseSecretKey -}}
-- name: license-key
-  mountPath: /etc/vm-license-key
-  readOnly: true
-{{- end -}}
+  {{- $licenseSecretKey := (include "vm.license.secret.key" .) -}}
+  {{- $licenseSecretName := (include "vm.license.secret.name" .) -}}
+  {{- if and $licenseSecretName $licenseSecretKey -}}
+    {{- $mount := list (dict "name" "license-key" "mountPath" "/etc/vm-license-key" "readOnly" true) -}}
```
zekker6, line 48: Same as previous comment

charts/victoria-metrics-common/templates/_enterprise.tpl
```diff
@@ -27,38 +30,36 @@
...
-- name: license-key
-  secret:
-    secretName: {{ $licenseSecretName }}
-{{- end -}}
+  {{- $licenseSecretKey := (include "vm.license.secret.key" .) -}}
+  {{- $licenseSecretName := (include "vm.license.secret.name" .) -}}
+  {{- if and $licenseSecretName $licenseSecretKey -}}
+    {{- $volume := list (dict "name" "license-key" "secret" (dict "secretName" $licenseSecretName)) -}}
+    {{- toYaml $volume -}}
+  {{- end -}}
```
zekker6, line 38: Won't this produce exactly the same result as it was before?
It seems to me like plain YAML will be easier to read and understand compared to helm `dict` definition, so I'd keep previous version.

AndrewChubatiuk: restored

zekker6 (approved):

PR #1392, opened 2024-09-02
common: added more templates

AndrewChubatiuk:
- Allow `appKey` argument to be a list to support deeply nested objects
- Added `vm.namespace`, which returns `namespaceOverride` or `global.namespaceOverride` or `Release.Namespace` as a default. See [related issue](https://github.com/VictoriaMetrics/helm-charts/issues/569)
- Added `vm.managed.fullname`, which returns default fullname prefixed by `appKey`
- Added `vm.plain.fullname`, which returns default fullname suffixed by `appKey`

### DIFF
--- a/charts/victoria-metrics-common/CHANGELOG.md
+++ b/charts/victoria-metrics-common/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ## Next release
 
-- TODO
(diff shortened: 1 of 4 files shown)


f41gh7 (approved):
LGTM

PR #1297, opened 2024-08-23
fixed statefulset pvc mount

AndrewChubatiuk:
Fixes https://github.com/VictoriaMetrics/helm-charts/issues/189

PR #1402, opened 2024-09-02
fixed PVC in statefulset

AndrewChubatiuk:
related https://github.com/VictoriaMetrics/helm-charts/pull/1297#pullrequestreview-2276092046

### DIFF
--- a/charts/victoria-metrics-single/templates/server-statefulset.yaml
+++ b/charts/victoria-metrics-single/templates/server-statefulset.yaml
@@ -257,13 +257,15 @@ spec:
           configMap:
             name: {{ include "victoria-metrics.server.relabel.configname" . }}
         {{- end }}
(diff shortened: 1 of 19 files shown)


charts/victoria-metrics-single/templates/server-statefulset.yaml
```diff
@@ -257,13 +257,15 @@ spec:
           configMap:
             name: {{ include "victoria-metrics.server.relabel.configname" . }}
         {{- end }}
+        {{- if or (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
```
mkilchhofer, line 260: Eventually it would be easier to understand if you define a helper variable for this on the top of the file.
- use it for this check
- use the inverted variable for the volumeClaimTemplate

zekker6 (approved):

PR #1431, opened 2024-09-04
cluster bugfix and drop empty endpoints from scrape

AndrewChubatiuk:
- fixed resources templating for vmstorage in victoria-metrics-cluster chart
- drop empty "endpoints" from spec to support switching to different scrape config types


f41gh7 (approved):
LGTM

PR #1435, opened 2024-09-04
feature: Add initcontainer capability to initialize vmalert pod

rakesh-von:


charts/victoria-metrics-alert/values.yaml
```diff
@@ -308,6 +308,8 @@ server:
     #         cpu: 1
     #         memory: 500Mi
     #       controlledResources: ["cpu", "memory"]
+  ## Additional initContainers to initialize the pod
```
AndrewChubatiuk, line 311: please use '# -- ' in front of comment messages to be parsed correctly by helm docs and run 'make gen-docs' from a repo root

AndrewChubatiuk (changes requested):

charts/victoria-metrics-alert/values.yaml
rakesh-von: done.

AndrewChubatiuk (approved):

PR #1444, opened 2024-09-05
charts/victoria-metrics-alert: remove license args for AM

zekker6:
AM does not use VictoriaMetrics' licensing flags, so remove templating of these flags.


AndrewChubatiuk (approved):
Good catch. Thanks

Issue #1449, opened 2024-09-05
[victoria-metrics-k8s-stack] Type mismatch `($spec.extraArgs).tls` after refactoring


psmolkin:
Hello,
Unfortunately, the last refactoring resulted in a type mismatch error:
https://github.com/VictoriaMetrics/helm-charts/blame/victoria-metrics-k8s-stack-0.25.14/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl#L133-L135
```yaml
# ./victoria-metrics-k8s-stack/templates/_helpers.tpl:133, {{- define "vm.url" -}} 
    {{- $isSecure = ($spec.extraArgs).tls | default $isSecure -}}
    {{- $proto = (ternary "https" "http" $isSecure) -}}
```
This part references the wrong type, since the `extraArgs` field must `object (keys:string, values:string)`:
see [extraArgs](https://docs.victoriametrics.com/operator/api/index.html#vminsert) at `VMClusterSpec.VMInsert`

This results in two mutually exclusive situations:
- String param
  ```yaml
  vmcluster:
    enabled: false
    spec:
      vminsert:
        extraArgs:
          tls: "true"
  ```
   leads to `Error: UPGRADE FAILED: template: victoria-metrics-k8s-stack/templates/victoria-metrics-operator/vmagent/vmagent.yaml:13:9: executing "victoria-metrics-k8s-stack/templates/victoria-metrics-operator/vmagent/vmagent.yaml" at <include "vm.agent.spec" .>: error calling include: template: victoria-metrics-k8s-stack/templates/_helpers.tpl:279:16: executing "vm.agent.spec" at <include "vm.agent.remote.write" .>: error calling include: template: victoria-metrics-k8s-stack/templates/_helpers.tpl:271:56: executing "vm.agent.remote.write" at <include "vm.write.endpoint" $ctx>: error calling include: template: victoria-metrics-k8s-stack/templates/_helpers.tpl:170:37: executing "vm.write.endpoint" at <include "vm.url" $ctx>: error calling include: template: victoria-metrics-k8s-stack/templates/_helpers.tpl:134:41: executing "vm.url" at <$isSecure>: wrong type for value; expected bool; got string`
- Bool param
  ```yaml
  vmcluster:
    enabled: false
    spec:
      vminsert:
        extraArgs:
          tls: true
  ```
[...]

PR #1451, opened 2024-09-05
fixed proto, when TLS is enabled

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1449


Haleygo (approved):
LGTM

PR #1459, opened 2024-09-09
k8s-stack: removed CRDs subchart as they are already included in operator

AndrewChubatiuk:
moved CRDs to operators `crds` folder to avoid importing them in both k8s-stack and operator charts

### DIFF
--- a/charts/victoria-metrics-agent/Chart.lock
+++ b/charts/victoria-metrics-agent/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: victoria-metrics-common
   repository: https://victoriametrics.github.io/helm-charts
-  version: 0.0.15
-digest: sha256:28ebee3a7c14718527f72d912db85bb53fa05a0ffce17dfaab6ea836d7b495d6
-generated: "2024-10-11T07:18:51.909855212Z"
+  version: 0.0.16
+digest: sha256:7549c73bc949e7f40e67441c4a3a151471d4e258e9d32cc1c0f879f70825cce4
+generated: "2024-10-22T13:22:55.643975002Z"
--- a/charts/victoria-metrics-alert/Chart.lock
+++ b/charts/victoria-metrics-alert/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: victoria-metrics-common
   repository: https://victoriametrics.github.io/helm-charts
-  version: 0.0.15
-digest: sha256:28ebee3a7c14718527f72d912db85bb53fa05a0ffce17dfaab6ea836d7b495d6
-generated: "2024-10-11T07:18:55.704052714Z"
+  version: 0.0.16
+digest: sha256:7549c73bc949e7f40e67441c4a3a151471d4e258e9d32cc1c0f879f70825cce4
+generated: "2024-10-22T13:23:00.238306671Z"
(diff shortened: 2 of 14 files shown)


Haleygo:
address https://github.com/VictoriaMetrics/helm-charts/issues/901

charts/victoria-metrics-operator/CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## Next release
 
-- TODO
+- Moved crds to a shared chart and import them as a dependency
+- replaced `crd.*` properties to `crds.*` as they now belong to a subchart
```
Haleygo, line 6: Please mention this as a upgrade note too, as it requires user to change their values if they used those properties before.

Haleygo (approved):
LGTM!
See one comment.

charts/victoria-metrics-operator/templates/crd.yaml
```diff
@@ -1,18 +0,0 @@
-{{- /* do not update crds here, please update in /victoria-metrics-operator/crd.yaml */ -}}
```
f41gh7, line 1: It's breaking change that will have strongly negative impact on chart users.

 Helm cannot perform CRD updates according to the documentation:
https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#some-caveats-and-explanations

 And it was the main point of keeping CRDs at templates.

Haleygo: I had the same wrong impression( but resources with annotation `helm.sh/resource-policy: keep` won't be upgraded either during `helm upgrade`. So this pull request doesn't bring breaking change.
>The annotation helm.sh/resource-policy: keep instructs Helm to skip deleting this resource when a helm operation (such as helm uninstall, helm upgrade or helm rollback) would result in its deletion.

Haleygo: We might need to add some instructions to explain this and tell users how to upgrade crds for operator and k8s-stack charts, wdyt @AndrewChubatiuk

AndrewChubatiuk: i've just checked helm code, this annotations prevents removal during upgrade, but seems like resource update itself should happen
https://github.com/helm/helm/blob/49819b4ef782e80b0c7f78c30bd76b51ebb56dc8/pkg/kube/client.go#L214

f41gh7: Sounds like a bug at helm side. Because documentation wording mentions only `deleting` resource case. It should be upgrade on regular basis.

AndrewChubatiuk: > Helm cannot perform CRD updates according to the documentation
but crds are in templates in operator chart only, k8s-stack has them in `crds` folder of a subchart
I can update PR and leave it for user to decide if he wants crd as templates or as immutable object

Haleygo: >Because documentation wording mentions only deleting resource case. It should be upgrade on regular basis.

Sorry, I did the wrong test, that's true and we shouldn't change it for operator chart.

AndrewChubatiuk: @f41gh7 @Haleygo 
reverted templating back, added boolean`crds.plain` instead, which is set to `false` by default to render CRDs from templates or when it's set to `true` it renders crds subchart `crds` folder

f41gh7 (approved):
LGTM

PR #1460, opened 2024-09-09
updated webhook service port

AndrewChubatiuk:
made service port for webhook the same as pod port
can be related to 
https://github.com/VictoriaMetrics/operator/issues/1094


Haleygo:
As I understand, the old values are correct, but [the user](https://github.com/VictoriaMetrics/operator/issues/1094) didn't expose 443 on their nodes for webhook svc?
If so, maybe we should provide a way to customize service port, wdyt?

AndrewChubatiuk:
as we do not customize `http` service port, so guess there's no need to customize `webhook` one as well, just made the value of webhook port the same as on container in the same manner as we have for http

Haleygo (approved):

Issue #1476, opened 2024-09-12
[victoria-metrics-anomaly] Invalid option loggerFormat for vmanomaly


TheDJVG:
Latest release of VMAnomaly does not have a `--loggerFormat` option but it's set in the default `values.yaml`:
https://github.com/VictoriaMetrics/helm-charts/blob/f5a8450c3454e19065b63899a58fd009e018b280/charts/victoria-metrics-anomaly/values.yaml#L42-L43

This generates a startup error:
```bash
# docker run --rm victoriametrics/vmanomaly --loggerFormat=json /config
usage: vmanomaly.pyc [-h] [--log_level {INFO,ERROR,WARNING,FATAL,DEBUG}]
                     [--watch] [--license STRING | --license-file PATH]
                     [--license-verify-offline]
                     config [config ...]
vmanomaly.pyc: error: unrecognized arguments: --loggerFormat=json
```

AndrewChubatiuk:
fix is available in https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-anomaly-1.4.5

PR #1477, opened 2024-09-12
vmanomaly: removed not loggerFormat flag from extraArgs

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1476

### DIFF
--- a/charts/victoria-metrics-anomaly/values.yaml
+++ b/charts/victoria-metrics-anomaly/values.yaml
@@ -15,7 +15,7 @@ image:
   # -- Victoria Metrics anomaly Docker repository and image name
   repository: victoriametrics/vmanomaly
   # -- Tag of Docker image
-  tag: v1.15.4 # rewrites Chart.AppVersion
+  tag: v1.15.9 # rewrites Chart.AppVersion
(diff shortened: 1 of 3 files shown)


charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -15,7 +15,7 @@ image:
   # -- Victoria Metrics anomaly Docker repository and image name
   repository: victoriametrics/vmanomaly
   # -- Tag of Docker image
-  tag: v1.15.4 # rewrites Chart.AppVersion
+  tag: v1.15.9 # rewrites Chart.AppVersion
```
Haleygo, line 18: nit: should we leave this null to use `Chart.AppVersion`?

Haleygo (approved):

charts/victoria-metrics-anomaly/values.yaml
AndrewChubatiuk: vmanomaly should change their tagging policy and start using `-enterprise` suffix, till that moment tag should be set explicitly

Issue #829, opened 2024-01-17
`vmauth` missing from `victoria-metrics-k8s-stack` helm chart


kbweave:
Is there a particular reason technical or otherwise that `vmauth` isn't exposed and configurable in the `victoria-metrics-k8s-stack` helm chart?

Haleygo:
Thanks for notice!
I think there is no particular reason, we just don't have too many usercases where vmauth is needed for load balancing or authorization in single vmcluster before.
We might add it to `victoria-metrics-k8s-stack` chart after we [create multi-AZ chart](https://github.com/VictoriaMetrics/helm-charts/issues/789).

AndrewChubatiuk:
vmauth is available in k8s-stack 0.25.17

PR #1488, opened 2024-09-13
added vmauth to k8s-stack

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/829

### DIFF
--- a/charts/victoria-metrics-k8s-stack/values.yaml
+++ b/charts/victoria-metrics-k8s-stack/values.yaml
@@ -716,6 +716,31 @@ vmalert:
     #    hosts:
     #      - vmalert.domain.com
 
+vmauth:
+  enabled: false
+  annotations: {}
+  # -- (object) full spec for VMAuth CRD. Allowed values described [here](https://docs.victoriametrics.com/operator/api#vmauthspec)
+  spec:
+    discover_backend_ips: true
+    port: "8427"
+  # -- (object) vmagent ingress configuration
+  ingress:
+    enabled: false
+    # For Kubernetes >= 1.18 you should specify the ingress-controller via the field ingressClassName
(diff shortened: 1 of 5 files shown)


charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -716,6 +716,31 @@ vmalert:
     #    hosts:
     #      - vmalert.domain.com
 
+vmauth:
+  enabled: false
+  annotations: {}
+  # -- (object) full spec for VMAuth CRD. Allowed values described [here](https://docs.victoriametrics.com/operator/api#vmauthspec)
+  spec:
+    discover_backend_ips: true
```
f41gh7, line 724: vmauth spec doesn't have such setting at spec https://docs.victoriametrics.com/operator/api/

charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -716,6 +716,31 @@ vmalert:
...
+    port: "8427"
+  # -- (object) vmagent ingress configuration
+  ingress:
```
f41gh7, line 727: Ingress is part of vmauth.spec

charts/victoria-metrics-k8s-stack/values.yaml
AndrewChubatiuk: it does https://docs.victoriametrics.com/operator/api/#vmauthspec

charts/victoria-metrics-k8s-stack/values.yaml
AndrewChubatiuk: done

f41gh7 (approved):
LGTM

PR #1490, opened 2024-09-13
Enable to add annotations to vmanomaly  configMap

mohamedragab2024:
**Description**

This pull request introduces the ability to add annotations to the anomaly ConfigMap. Annotations are crucial for integrating the ConfigMap with tools like Vault, among others. By allowing annotations, this change enhances the flexibility of the ConfigMap, enabling seamless integrations with external systems that rely on metadata

### DIFF
--- a/charts/victoria-metrics-anomaly/templates/configmap.yaml
+++ b/charts/victoria-metrics-anomaly/templates/configmap.yaml
@@ -4,7 +4,7 @@ metadata:
   name: {{ include "chart.fullname" . }}
   namespace: {{ include "vm.namespace" . }}
   labels: {{- include "chart.labels" . | nindent 4 }}
-  {{- with .Values.config.annotations }}
+  {{- with .Values.configMapAnnotations }}
   annotations: {{ toYaml . | nindent 10 }}
   {{- end }}
 data:
--- a/charts/victoria-metrics-anomaly/templates/configmap.yaml
+++ b/charts/victoria-metrics-anomaly/templates/configmap.yaml
@@ -4,5 +4,8 @@ metadata:
   name: {{ include "chart.fullname" . }}
   namespace: {{ include "vm.namespace" . }}
   labels: {{- include "chart.labels" . | nindent 4 }}
+  {{- with .Values.config.annotations }}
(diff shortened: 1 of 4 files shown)


charts/victoria-metrics-anomaly/templates/configmap.yaml
```diff
@@ -4,5 +4,8 @@ metadata:
   name: {{ include "chart.fullname" . }}
   namespace: {{ include "vm.namespace" . }}
   labels: {{- include "chart.labels" . | nindent 4 }}
+  {{- with .Values.config.annotations }}
```
AndrewChubatiuk, line 7: please change this parameter to `.Values.configMeta.annotations` to avoid having these parameters in configuration file

mohamedragab2024: Can you elaborate more What is `configMeta` ? What do you think about  annotations in the root level to be `configMapAnnotations` similar to `podAnnotations` ?

charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -161,6 +161,8 @@ config:
     datasource_url: ""
     # -- For VictoriaMetrics Cluster version only, tenants are identified by accountID or accountID:projectID. See VictoriaMetrics Cluster multitenancy docs
     tenant_id: ""
+    # -- ConfigMap annotations
```
miry, line 164: I think there is an extra space.

charts/victoria-metrics-anomaly/templates/configmap.yaml
AndrewChubatiuk:
```
What do you think about annotations in the root level to be configMapAnnotations similar to podAnnotations
```
this also works

mohamedragab2024: I changed it , Could you take a look

AndrewChubatiuk:
@mohamedragab2024 could you please also add a changelog entry to `charts/victoria-metrics-anomaly/CHANGELOG.md`?

AndrewChubatiuk (approved):
LGTM, thanks for contribution

Issue #376, opened 2022-09-09
Consider using nginx or vmauth for balancing read load to vmselects for cluster chart


hagen1778:
Looks like balancing via default k8s service is not always optimal. During the benchmarks I found that vmselects are receving non equal number of queries. Especially, in cases when read queries have different execution time. This may result into excessive resource usage for some vmselects. Would be great to try using nginx for better balancing of the read load. Least-loaded strategy seems like a better fit here.

cloud-66:
at least, Open source nginx can't aware resource consumption by backends. How can this be implemented ?

hagen1778:
@cloud-66 we can try using [vmauth](https://docs.victoriametrics.com/vmauth.html) instead, as it provides [load-balancing feature out-of-box](https://docs.victoriametrics.com/vmauth.html#load-balancing). 

We can try to validate the assumption for vmauth to improve balancing by testing it [on our playground](https://play-grafana.victoriametrics.com/explore?left=%7B%22datasource%22:%22P38648FE0F8C5BEA2%22,%22queries%22:%5B%7B%22datasource%22:%7B%22type%22:%22victoriametrics-datasource%22,%22uid%22:%22P38648FE0F8C5BEA2%22%7D,%22editorMode%22:%22code%22,%22expr%22:%22sum%28rate%28vm_http_requests_total%7Bjob%3D~%5C%22vmselect-benchmark-vm-cluster%5C%22,%20instance%3D~%5C%22.%2A%5C%22,%20path%3D~%5C%22.%2Aquery%5C%22%7D%5B$__rate_interval%5D%29%29%20by%20%28pod%29%20%3E%200%22,%22format%22:%22time_series%22,%22intervalFactor%22:1,%22range%22:true,%22refId%22:%22A%22,%22interval%22:%22%22%7D%5D,%22range%22:%7B%22from%22:%22now-3h%22,%22to%22:%22now%22%7D%7D&orgId=1).

AndrewChubatiuk:
vmauth was included in release [0.14.1](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-cluster-0.14.1)

Issue #273, opened 2022-01-27
Multi Retention Setup


Alon-Katz:
Hey,

I was wondering, how can the Multi Retention Setup described here: https://docs.victoriametrics.com/guides/guide-vmcluster-multiple-retention-setup.html be implemented using the helm chart?
I cant deploy different retention periods in the same helm chart deployment, but I can deploy multiple helm charts. The issue is the vmselect that should communicate with all the vmstorages that was deployed in multiple helm charts..

Thanks for the help!

Alon-Katz:
Thanks @f41gh7!

Are there any plans to support providing only the headless service names for vmstorage? Instead of manually providing the list of vmstorage instances.

This will make it much easier to maintain, and will open a lot more possibilities. For example using HPA (with downscale behavior disabled) for vmstorage instances, this way when the HPA will autoscale the vmstorage then the headless service will provide all the new instances and it makes maintaining the vm cluster much easier for heavy production clusters

AndrewChubatiuk:
hey @Alon-Katz
with enterprise license enabled vmstorage SRV records can be passed to vminsert and vmselect, where all nodes will be automatically discovered. This feature is available in victoria-metrics-cluster chart with enterprise license configured and `autoDiscovery: true` starting from release [0.14.1](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-cluster-0.14.1)

PR #1494, opened 2024-09-14
cluster: storage node addesses discovery, added vmauth

AndrewChubatiuk:
- added ability to set additional storage nodes to default ones using `<vmselect|vminsert>.extraArgs.storageNode` array
- added storage node srv discovery if enterprise license is enabled. fixes https://github.com/VictoriaMetrics/helm-charts/issues/273
- added optional vmauth in front of vmselect and vminsert. fixes https://github.com/VictoriaMetrics/helm-charts/issues/376

related issue https://github.com/VictoriaMetrics/VictoriaMetrics/issues/7040


Haleygo (commented):
>added optional vmauth in front of vmselect and vminsert. fixes https://github.com/VictoriaMetrics/helm-charts/issues/376

User can only benefit more from vmauth than k8s service when address under `url_prefix` is pointing to a headless service, that way vmauth can resolve all the backend addresses in advance and proxy request to the least-loaded endpoint. But headless svc is not the default service type for both vmselect and vminsert here.
And I don't think we should add more logic to change the service type when enable the vmauth without making user understand why we do that. What about adding some instructions in readme?

charts/victoria-metrics-cluster/values.yaml
```diff
@@ -11,12 +11,18 @@ global:
...
+  # -- k8s cluster domain suffix, uses for building storage pods' FQDN. Details are [here](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/)
+  cluster:
+    dnsDomain: cluster.local
 
 # -- K8s cluster domain suffix, uses for building storage pods' FQDN. Details are [here](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/)
 clusterDomainSuffix: cluster.local
 # -- Print information after deployment
 printNotes: true
 
+# use SRV discovery for storageNode and selectNode flags for enterprise version
```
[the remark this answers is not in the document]
AndrewChubatiuk: license can also be defined at `global.license`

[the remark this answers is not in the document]
AndrewChubatiuk: added failure if no license is defined with autoDiscovery enabled

AndrewChubatiuk:
added for `vmauth.enabled` param description about headless services

Haleygo:
Need to set `vmselect.name` to "" for the new name template to avoid creating new service.
https://github.com/VictoriaMetrics/helm-charts/blob/d478a2454d81cea5025e16f5b1c02289751a7326/charts/victoria-metrics-cluster/values.yaml#L71

btw, templates seem to complicate debugging and make value modifications more fragile. 
I don't understand why we need to consolidate all the component' services into one template yaml, and they have very different port settings in the `_helpers.tpl`. The previous resources look easier to me only with some duplicated fields.
https://github.com/VictoriaMetrics/helm-charts/blob/d478a2454d81cea5025e16f5b1c02289751a7326/charts/victoria-metrics-cluster/templates/_helpers.tpl#L261-L288

charts/victoria-metrics-cluster/templates/_helpers.tpl
```diff
@@ -106,131 +67,135 @@ app: {{ .Values.vmselect.name }}
...
+  {{- $_ := set . "style" "plain" }}
+  {{- $_ := set . "appKey" "vmstorage" }}
+  {{- $args = mergeOverwrite $args (fromYaml (include "vm.license.flag" .)) -}}
+  {{- $args = mergeOverwrite $args $app.extraArgs -}}
+  {{- $storage := $Values.vmstorage }}
+  {{- if and (not $app.suppressStorageFQDNsRender) (and $storage.enabled $storage.replicaCount) }}
+    {{- $storageNodes := default list }}
+    {{- $fqdn := include "vm.fqdn" . }}
+    {{- if $Values.autoDiscovery }}
+      {{- if (include "vm.enterprise.disabled" . ) "true" }}
```
zekker6, line 104:
```suggestion
      {{- if eq (include "vm.enterprise.disabled" . ) "true" }}
```

charts/victoria-metrics-cluster/templates/_helpers.tpl
```diff
@@ -106,131 +67,135 @@ app: {{ .Values.vmselect.name }}
...
+  {{- $_ := set . "appKey" "vmstorage" }}
   {{- $_ := set $args "cacheDataPath" $app.cacheMountPath -}}
   {{- $args = mergeOverwrite $args (fromYaml (include "vm.license.flag" .)) -}}
   {{- $args = mergeOverwrite $args $app.extraArgs -}}
+  {{- $storage := $Values.vmstorage }}
+  {{- if and (not $app.suppressStorageFQDNsRender) (and $storage.enabled $storage.replicaCount) }}
+    {{- $storageNodes := default list }}
+    {{- $fqdn := include "vm.fqdn" . }}
+    {{- if $Values.autoDiscovery }}
+      {{- if (include "vm.enterprise.disabled" . ) "true" }}
```
zekker6, line 150:
```suggestion
      {{- if eq (include "vm.enterprise.disabled" . ) "true" }}
```

charts/victoria-metrics-cluster/templates/_helpers.tpl
```diff
@@ -106,131 +67,135 @@ app: {{ .Values.vmselect.name }}
...
+      {{- $_ := unset . "appIdx" }}
+    {{- end }}
+    {{- $_ := set $args "storageNode" (concat ($args.storageNode | default list) $storageNodes) }}
+  {{- end }}
+  {{- if and $app.statefulSet.enabled $app.enabled $app.replicaCount }}
+    {{- $selectNodes := default list }}
+    {{- $_ := set . "appKey" "vmselect" }}
+    {{- $fqdn := include "vm.fqdn" . }}
+    {{- if $Values.autoDiscovery }}
+      {{- if (include "vm.enterprise.disabled" . ) "true" }}
```
zekker6, line 171:
```suggestion
      {{- if eq (include "vm.enterprise.disabled" . ) "true" }}
```

zekker6 (commented):
LGTM in general.
Added a few suggestions to fix a typo in configs. I would expect CI to detect this, but it seems like we don't have a test set to cover this templating path. Could you add some tests to cover this part?

AndrewChubatiuk:
> Could you add some tests to cover this part?
thanks for review. added this scenario

zekker6 (approved):

Issue #1260, opened 2024-08-20
[victoria-metrics-k8s-stack] Add `routePrefix` to VM Components


alihoda:
Hi. It would be nice if vm-k8s-stack has a option like `routePrefix` in *kube-prometheus-stack* that helps to set a root path for each components and so just use one domain for entire stack.
I currently should handle it with `ingress.annotations`:

```
ingress:
  select:
    path: /s/?(.*)
    annotations:
      kubernetes.io/ingress.class: "nginx"
      nginx.ingress.kubernetes.io/rewrite-target: /$1
      nginx.ingress.kubernetes.io/use-regex: "true"
```

As it seems, it is a little messy!! And if use something like anchors to reuse it for other ingresses, then for grafana and vmalertmanger you can not use it because they support the root path option. In the end you need to repeat the necessary annotations for these two!!

AndrewChubatiuk:
could you please elaborate more about this case?
which services do you want to have ingress for?
in a final configuration do you expect all services to be behind a single ingress? (i'm not aware if nginx ingress supports multiple ingresses with the same host:port, but different paths, like alb controller does)

alihoda:
> could you please elaborate more about this case? which services do you want to have ingress for? in a final configuration do you expect all services to be behind a single ingress? (i'm not aware if nginx ingress supports multiple ingresses with the same host:port, but different paths, like alb controller does)


In the `vm-k8s-stack`, specifically for `vmcluster`, we typically have Ingress resources for `vminsert`, `vmstorage`, and `vmselect`, with `vmselect` being the most crucial. Additionally, in both single and cluster modes, `vmagent` and `vmalert` also have Ingress options, mainly to provide detailed insights beyond what is available in Grafana dashboards.

For our scenario, the goal is to use a single domain, `vm.example.com`, with Ingress paths for three key services: `vmselect`, `vmagent`, and `vmalert` (we also have VMAlertMnager and Grafana for that domain too but they are handled with `routePrefix` and `grafana.ini.server.root`). The desired paths are as follows:

- `/s` for `vmselect`
- `/a` for `vmagent`
- `/al` for `vmalert`

[...]

PR #1496, opened 2024-09-14
k8s-stack: use path prefix from args as a default ingress path

AndrewChubatiuk:
related issue https://github.com/VictoriaMetrics/helm-charts/issues/1260

### DIFF
--- a/charts/victoria-metrics-k8s-stack/values.yaml
+++ b/charts/victoria-metrics-k8s-stack/values.yaml
@@ -416,7 +416,7 @@ vmcluster:
         # kubernetes.io/ingress.class: nginx
         # kubernetes.io/tls-acme: "true"
       labels: {}
-      path: '/select{{ dig "extraArgs" "http.pathPrefix" "/" .Values.vmcluster.spec.vmselect }}'
+      path: '{{ dig "extraArgs" "http.pathPrefix" "/" .Values.vmcluster.spec.vmselect }}'
       # pathType is only for k8s > 1.19
       pathType: Prefix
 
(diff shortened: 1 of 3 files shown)


charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -416,7 +416,7 @@ vmcluster:
         # kubernetes.io/ingress.class: nginx
         # kubernetes.io/tls-acme: "true"
       labels: {}
-      path: /
+      path: '/select{{ dig "extraArgs" "http.pathPrefix" "/" .Values.vmcluster.spec.vmselect }}'
```
Haleygo, line 419: why adding `/select` as fixed prefix here? vmselect also exposes urls without `/select`.

charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -451,7 +451,7 @@ vmcluster:
         # kubernetes.io/ingress.class: nginx
         # kubernetes.io/tls-acme: "true"
       labels: {}
-      path: /
+      path: '/insert{{ dig "extraArgs" "http.pathPrefix" "/" .Values.vmcluster.spec.vminsert }}'
```
Haleygo, line 454: same here

charts/victoria-metrics-k8s-stack/values.yaml
AndrewChubatiuk: updated pr

zekker6 (approved):

PR #1523, opened 2024-09-21
sort targets of operator's cleanup job

ushitora-anqou:
The current implementation of the operator's cleanup job uses the Helm's `keys` function to find which custom resources should be deleted. However, the array returned by `keys` isn't in a predictable order, and may change each time a user runs a Helm command.

This commit fixes that issue by sorting the array returned by `keys` before using it. It utilizes a Helm's function `sortAlpha` for this purpose.


AndrewChubatiuk (approved):

ushitora-anqou:
Thank you!

Issue #1525, opened 2024-09-23
[victoria-logs-single] Extra quotes on fluent-bit default host definition, no upstream connection


k7k0:
**Chart name and version**
chart: victoria-logs-single
version: v0.6.3

**Describe the bug**

[#1245](https://github.com/VictoriaMetrics/helm-charts/pull/1245) added quotes around host [output host section](https://github.com/VictoriaMetrics/helm-charts/blob/victoria-logs-single-0.6.3/charts/victoria-logs-single/values.yaml#L319) breaking connectivity:

```
2024/09/23 11:48:15] [ warn] [net] getaddrinfo(host=''victoria-logs-server'', err=8): Misformatted domain name
[2024/09/23 11:48:15] [error] [output:http:http.0] no upstream connections available to 'victoria-logs-server':9428
```

**Custom values**
Please provide only custom values (excluding default ones):
```yaml
fluent-bit:
  enabled: true
```

AndrewChubatiuk:
hey @k7k0 
Thanks for submitting an issue
Will be fixed in the next release

AndrewChubatiuk:
available in 0.6.4

PR #1526, opened 2024-09-23
fixed default fluentbit config, use headers instead of query args

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1525


zekker6 (approved):

Andrey36652:
What is the reason to use headers instead of query params?

AndrewChubatiuk:
better readability, do you have any objections?

Andrey36652:
No, I thought it is some kind of functional difference.

PR #1532, opened 2024-09-25
vmagent: added remoteWrite param as a replacement for remoteWriteUrls

AndrewChubatiuk:
removed `multiTenantUrls` as it's deprecated in VictoriaMetrics 1.102.0
added `remoteWrite` param, which replaces `remoteWriteUrls`and besides adding`remoteWrite.*` prefixed cmd flags to vmagent process also generates config map with content of values if values types are array or map. e.g:
```
remoteWrite:
 - url: http://localhost/api/v1/write
   urlRelabelConfig:
    - action: keep
      source_labels: [env]
      regex: "dev"
```
`--remoteWrite.url=http://localhost/api/v1/write` and `--remoteWrite.urlRelabelConfig=/config/rw/0-urlRelabelConfig.yaml` flags will be added together with config map, which contains
```
data:
  0-urlRelabelConfig.yaml: |
     - action: keep
       source_labels: [env]
       regex: "dev"
```
which is mounted to a pod

### DIFF
--- a/charts/victoria-metrics-agent/CHANGELOG.md
+++ b/charts/victoria-metrics-agent/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## Next release
 
+**Update note**: `remoteWriteUrls` and `multiTenantUrls` parameters were replaced by `remoteWrite`. Please follow [upgrade guide](./README.md#upgrade-to-0130)
+
 - Fail if no remoteWriteUrls set
+- Added `remoteWrite` array param, which can contain all `remoteWrite.*` flag values. Please check chart docs for details.
 
 ## 0.12.2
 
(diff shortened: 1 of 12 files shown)


charts/victoria-metrics-agent/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## Next release
 
+**Update note**: `remoteWriteUrls` and `multiTenantUrls` were deprecated and replaced by `remoteWrite`.
```
zekker6, line 3: Based on code changes an update note should be:

```suggestion
**Update note**: `remoteWriteUrls` and `multiTenantUrls` parameters were replaced by `remoteWrite`.
```

Deprecation is not appliable here since parameters are now removed in favor of a new one. 

Wdyt about adding a section to a readme called "upgrade guide"?
This change can be a first item to note there. We can put a short guidance like update X to Y so that users won't have to check an updated version of `values.yaml`.
Alternatively, we can do the same in changelog but it will be easier find multiple pending changes when upgrading between multiple versions if such guides would be at one place.

AndrewChubatiuk: @zekker6 updated PR, changed upgrade guide

zekker6 (approved):

Issue #1543, opened 2024-10-01
bug: etcd monitoring does not work properly on vanilla kubeadm v1.31.1


gecube:
Good day!

I am using the latest chart version of `victoria-metrics-k8s-stack`:

```bash
root@master-1:~# helm ls -A | grep victoria
victoria-logs-single      	monitoring-system        	2       	2024-09-30 18:07:35.47983054 +0000 UTC 	deployed	victoria-logs-single-0.6.4       	v0.29.0    
victoria-metrics-k8s-stack	monitoring-system        	9       	2024-09-30 20:56:38.409079006 +0000 UTC	deployed	victoria-metrics-k8s-stack-0.26.0	v1.102.1   
victoria-metrics-operator 	victoria-metrics-operator	1       	2024-09-30 18:00:02.135494903 +0000 UTC	deployed	victoria-metrics-operator-0.35.2 	v0.48.3   
```

I found that etcd monitoring does not work with the defaults as static manifest looks like:

```bash
root@master-1:~# cat /etc/kubernetes/manifests/etcd.yaml 
apiVersion: v1
kind: Pod
metadata:
  annotations:
    kubeadm.kubernetes.io/etcd.advertise-client-urls: https://49.13.236.131:2379
  creationTimestamp: null
  labels:
    component: etcd
    tier: control-plane
  name: etcd
  namespace: kube-system
spec:
  containers:
  - command:
    - etcd
    - --advertise-client-urls=https://49.13.236.131:2379
    - --cert-file=/etc/kubernetes/pki/etcd/server.crt
    - --client-cert-auth=true
    - --data-dir=/var/lib/etcd
    - --experimental-initial-corrupt-check=true
    - --experimental-watch-progress-notify-interval=5s
    - --initial-advertise-peer-urls=https://49.13.236.131:2380
    - --initial-cluster=master-1=https://49.13.236.131:2380
    - --key-file=/etc/kubernetes/pki/etcd/server.key
    - --listen-client-urls=https://127.0.0.1:2379,https://49.13.236.131:2379
    - --listen-metrics-urls=http://0.0.0.0:2381
    - --listen-peer-urls=https://49.13.236.131:2380
    - --name=master-1
    - --peer-cert-file=/etc/kubernetes/pki/etcd/peer.crt
    - --peer-client-cert-auth=true
    - --peer-key-file=/etc/kubernetes/pki/etcd/peer.key
[...]

PR #1544, opened 2024-10-01
Updated minikube ETCD default port

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1543

### DIFF
--- a/charts/victoria-metrics-agent/README.md
+++ b/charts/victoria-metrics-agent/README.md
@@ -6,7 +6,7 @@ Victoria Metrics Agent - collects metrics from various sources and stores them t
 
 ## Prerequisites
 
-* Install the follow packages: ``git``, ``kubectl``, ``helm``, ``helm-docs``. See this [tutorial](../../REQUIREMENTS.md).
+* Install the follow packages: ``git``, ``kubectl``, ``helm``, ``helm-docs``. See this [tutorial](https://docs.victoriametrics.com/helm/requirements/).
 
 ## How to install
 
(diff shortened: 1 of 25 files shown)


charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -1127,9 +1127,9 @@ kubeEtcd:
     # -- Enable service for ETCD metrics scraping
     enabled: true
     # -- ETCD service port
-    port: 2379
+    port: 2381
```
zekker6, line 1130: Won't this break etcd scraping for older versions where the port is still 2379?

AndrewChubatiuk: updated minikube values port instead, since GKE still relies on 2381

charts/victoria-metrics-agent/values.yaml
```diff
@@ -27,7 +27,7 @@ deployment:
 
 # -- [K8s StatefulSet](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/) specific variables
 statefulset:
-  enabled: false
+  enabled: true
```
zekker6, line 30: is this expected in this PR?

AndrewChubatiuk: removed this change

charts/victoria-metrics-agent/README.md
```diff
@@ -1445,7 +1445,7 @@ periodSeconds: 15
       <td>object</td>
       <td><pre class="helm-vars-default-value" language-yaml" lang="plaintext">
 <code class="language-yaml">clusterMode: false
-enabled: false
+enabled: true
```
zekker6, line 1448: looks like this doc was also updated by accident

AndrewChubatiuk: updated

zekker6 (approved):

PR #1546, opened 2024-10-02
bump to 1.104.0

hagen1778:
I followed https://docs.victoriametrics.com/release-guide/#helm-charts


AndrewChubatiuk (dismissed):

AndrewChubatiuk:
do we need to release new chart versions as well?

zekker6:
@AndrewChubatiuk Yes, that is correct. A full procedure would be to include changelog + chart version update. Would you like to take this? If not I can handle this by myself.

AndrewChubatiuk (approved):
LGTM

PR #1550, opened 2024-10-03
charts/victoria-metrics-anomaly: update vmanomaly, bugfixes

zekker6:
- update vmanomaly to 1.16.1
- update configuration for testing to be able to run vmanomaly
- fix configuration templating
- add option to dump models and data

### DIFF
--- a/charts/victoria-metrics-anomaly/templates/configmap.yaml
+++ b/charts/victoria-metrics-anomaly/templates/configmap.yaml
@@ -1,3 +1,10 @@
+{{- if not .Values.config.reader.tenant_id -}}
+{{- $ := unset .Values.config.reader "tenant_id" }}
+{{- end -}}
+{{- if not .Values.config.writer.tenant_id -}}
+{{- $ := unset .Values.config.writer "tenant_id" }}
+{{- end -}}
+---
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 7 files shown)


AndrewChubatiuk (approved):

charts/victoria-metrics-anomaly/templates/configmap.yaml
```diff
@@ -1,3 +1,10 @@
+{{- if not .Values.config.reader.tenant_id -}}
+{{- $ := unset .Values.config.reader "tenant_id" }}
+{{- end -}}
+{{- if not .Values.config.writer.tenant_id -}}
+{{- $ := unset .Values.config.writer "tenant_id" }}
+{{- end -}}
+---
```
fred-navruzov, line 7: can you please also add the same for `monitoring.push` as we also have `tenant_id` there ?

zekker6: It is not included as a part of default values so unless user sets it manually that won't be a problem for vmanomaly.

fred-navruzov: oh, got it, thanks for clarification!

fred-navruzov (commented):
thanks, LGTM!

Issue #1592, opened 2024-10-13
feature: Add AlertManager datasource to Grafana to victoria-metrics-k8s-stack chart


maxpain:
`kube-prometheus-stack` has this integration:

https://github.com/prometheus-community/helm-charts/blob/0529c1011be5ffd26419ef941920368ebdc2ee6f/charts/kube-prometheus-stack/values.yaml#L1103

Issue #1590, opened 2024-10-13
feature: Add some dashboards and rules to victoria-logs-single chart


maxpain:

AndrewChubatiuk:
hey @maxpain 
regarding dashboards your request sounds reasonable as it can be provided as a configmap, but how are you going to use rules and which format do you expect?

AndrewChubatiuk:
available in 0.7.0

maxpain:
@AndrewChubatiuk, could you please enable `grafana_dashboard: "1"` by default in the ConfigMap with the dashboard JSON file?

PR #1593, opened 2024-10-13
k8s-stack: added alertmanager datasource; vlogs: added dashboard

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1592
partially resolves https://github.com/VictoriaMetrics/helm-charts/issues/1590


charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -57,6 +57,8 @@ defaultDashboards:
     allowCrossNamespaceImport: false
   # -- Create dashboards as ConfigMap despite dependency it requires is not installed
   dashboards:
+    victorialogs:
```
Haleygo, line 60: Since VM k8s-stack chart doesn't install victorialogs, I don't think we should include victorialogs dashboard here.

charts/victoria-logs-single/values.yaml
```diff
@@ -359,3 +359,12 @@ fluent-bit:
 
 # -- Add extra specs dynamically to this chart
 extraObjects: []
+
+dashboards:
+  # -- Create VictoriaLogs dashboards
```
Haleygo, line 364: Maybe we should also support creating it as a GrafanaDashboard CR like k8s-stack [does](https://github.com/VictoriaMetrics/helm-charts/blob/ac89332d907dd7da7843541c52018bb36e06394c/charts/victoria-metrics-k8s-stack/templates/grafana/dashboard.yaml#L41), wdyt?

And I think we should disable it by default now, currently it's not clear or convenient to tell users how to get access to dashboard or rule in vlogs chart, since the chart doesn't contain grafana or ruler components.

charts/victoria-metrics-k8s-stack/values.yaml
AndrewChubatiuk: added vlogs to k8s-stack chart

charts/victoria-logs-single/values.yaml
AndrewChubatiuk: added this as well

charts/victoria-metrics-k8s-stack/values.yaml
Haleygo: I don't believe we should include victorialogs workload in victoria-metrics-k8s-stack chart:
1. it doesn't integrate with other components, like collecting their logs without other configuration.
2. I think user would prefer to install log and metrics management components separately, since they're not bundled.

I think we could have another new chart for victoriaLogs in the furture, which contains vm-operator, single&cluster version of victoriaLogs, grafana, vmalert&alertmanager and possibly some common optional log collectors.

wdyt @AndrewChubatiuk @zekker6 @f41gh7

AndrewChubatiuk: > I think we could have another new chart for victoriaLogs in the furture, which contains vm-operator, single&cluster version of victoriaLogs, grafana, vmalert&alertmanager and possibly some common optional log collectors.

the only difference between existing k8s-stack chart and suggested one is vmcluster/vmsingle or vlcluster/vlsingle and most of the templates are the same. The only benefit of a separate chart - no need to think about chart size limit, but even this problem could be solved with dashboard subcharts. Since we use a single operator for both VL and VM, many components are expected to be the same for VL chart, I don't see many benefits from having a separate chart. Maybe it makes sense to have one for distributed setup in a future

Haleygo: >the only difference between existing k8s-stack chart and suggested one is vlcluster/vlsingle instead of vmcluster/vmsingle and most of the templates are the same. 

No, the collectors(vmagent&log collector) and configurations(vmScrapeConf&logCollectConf) are different, the vmalert configurations(flags&rules) are different. 
If we have the vm-stack and vlogs-stack charts, the only shared resources I can see are grafana and vm-operator.
Except from avoiding copying the shared resources and some templates to the new chart, I see no good thing to have one vm-family chart instead of multiple charts for different stacks. vm-family chart makes the chart values&templates complicated. If users are new to vm or vlogs, or they only want to install one stack, they see a chart with lots of templates and values they don't need and wonder why. Plus we need to test more cases for the vm-family chart, like installing the entire stack and then uninstalling the metrics stack.

AndrewChubatiuk: okay, i can create a separate chart, but first would like to merge https://github.com/VictoriaMetrics/helm-charts/pull/1560 to avoid copying this CRD to a new chart as well

AndrewChubatiuk: @Haleygo 
Updated this PR, moved hack scripts to a repo root hack directory and made dashboards and alerts destination configurable, removed vlogs and added vlogs dashboards and rules to victoria-logs-single chart only

charts/victoria-metrics-k8s-stack/templates/grafana/dashboard.yaml
```diff
@@ -65,19 +62,17 @@ metadata:
     {{- if $.Values.grafana.sidecar.dashboards.label }}
```
Haleygo, line 62: Btw, where does `$.Values.grafana.sidecar.dashboards.label` come from?

charts/victoria-metrics-k8s-stack/CHANGELOG.md
```diff
@@ -1,6 +1,12 @@
 ## Next release
 
-- TODO
+- Added alertmanager datasource. See [this issue](https://github.com/VictoriaMetrics/helm-charts/issues/1592)
+- Renamed `grafana.sidecar.dashboards.additionalDashboardLabels` to `grafana.sidecar.dashboards.labels`
+- Renamed `grafana.sidecar.dashboards.additionalDashboardAnnotations` to `grafana.sidecar.dashboards.annotations`
+- Renamed `grafana.sidecar.datasources.default` to `grafana.sidecar.datasources.victoriametrics`
+- Renamed `grafana.additionalDataSources` to `grafana.sidecar.datasources.extra`
```
Haleygo, line 7: I prefer not to change this, the original value name looks ok to me. I remember some users using the additionalDataSource as the default one and this change could breaks their setup.

charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -785,61 +786,64 @@ grafana:
   sidecar:
     datasources:
       enabled: true
-      # -- List of default prometheus compatible datasource configurations.
-      # VM `url` will be added to each of them in templates and `type` will be set to defaultDatasourceType if not defined
-      default:
+      # -- List of prometheus compatible datasource configurations.
```
Haleygo, line 789: Just notice we're adding our own values to the thrid-party chart `grafana`, which I believe we should avoid.
Since those values are only for vm-stack chart templates, and it breaks if the Grafana chart ever uses any of the same value names. 
It would be better to move below datasources from `grafana.sidecar.datasources`.

charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -785,61 +786,64 @@ grafana:
...
       initDatasources: true
       createVMReplicasDatasources: false
     dashboards:
+      defaultTimezone: utc
       provider:
         name: default
         orgid: 1
-      additionalDashboardLabels: {}
+      labels: {}
+      annotations: {}
```
Haleygo, line 828: As described above, these value names could be used by grafana chart itself.

charts/victoria-metrics-k8s-stack/templates/grafana/dashboard.yaml
AndrewChubatiuk: its grafana chart variable

charts/victoria-metrics-k8s-stack/values.yaml
AndrewChubatiuk: moved to defaultDatasources

charts/victoria-metrics-k8s-stack/CHANGELOG.md
AndrewChubatiuk: moved this value to defaultDatasources as it's not a part of grafana chart as well

charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -779,68 +783,75 @@ vmagent:
     #    hosts:
     #      - vmagent.domain.com
 
+defaultDatasources:
```
Haleygo, line 786: Why not reuse the [grafana.datasources](https://github.com/grafana/helm-charts/blob/07419bd5220eef66e1c4906390cb27d153de0bb6/charts/grafana/values.yaml#L612), should be more easy to see and understand by user.
And I don't see the need for `perReplica` (the old `createVMReplicasDatasources`), as it seems to serve only for debug purposes and works only on vmsingle, which I doubt is used. We can remove the field and consolidate all datasources under `grafana.datasources` instead of creating our own.

AndrewChubatiuk: https://github.com/grafana/helm-charts/blob/07419bd5220eef66e1c4906390cb27d153de0bb6/charts/grafana/values.yaml#L858

AndrewChubatiuk: createVMReplicasDatasources -> perReplica is needed when vmsingle instance count is > 1

Haleygo: >https://github.com/grafana/helm-charts/blob/07419bd5220eef66e1c4906390cb27d153de0bb6/charts/grafana/values.yaml#L858

ok, I don't see they have check or restriction in code though.

>createVMReplicasDatasources -> perReplica is needed when vmsingle instance count is > 1

For what? I can only see for debug, and it's pretty strange way to create datasources for that.

AndrewChubatiuk: > For what?
for those, who are using VMsingle in HA mode, the same setup as we're now suggesting for VLogs users till cluster version is not available
> ok, I don't see they have check or restriction in code though.
There were complaints from users regarding dashboards, so we already moved `node-exporter-full` dashboard to configmap

Haleygo (approved):

PR #1595, opened 2024-10-15
common: use <appSection>.fullnameOverride as alternative source for resource name prefix

AndrewChubatiuk:
currenly default name prefix override can be extracted from:
- global.fullnameOverride
- global.<Chart.Name>.fullnameOverride:
- fullnameOverride
- <appSection>.name

Added `<appSection>.fullnameOverride` additionally to fix backward compatibility issue for cluster chart


f41gh7 (approved):
LGTM

PR #1604, opened 2024-10-18
charts/victoria-metrics-k8s-stack: add info about adding external dashboards

zekker6:


AndrewChubatiuk (approved):

AndrewChubatiuk (approved):

ascknx:
Why did you do this? I want to use both methods of loading dashboards. It is working with the original Grafana helm chart perfectly

PR #1623, opened 2024-10-22
vmsingle: use common templates

AndrewChubatiuk:


charts/victoria-metrics-single/values.yaml
```diff
@@ -56,7 +56,7 @@ server:
   # -- Enable deployment of server component. Deployed as StatefulSet
   enabled: true
   # -- Server container name
-  name: server
```
Haleygo, line 59: This name doesn't actually set the container's name, should we fix that?

charts/victoria-metrics-anomaly/CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
 
+- TODO
+
+## 1.6.2
+
+**Release date:** 2024-10-22
+
+![AppVersion: v1.17.2](https://img.shields.io/static/v1?label=AppVersion&message=v1.17.2&color=success&logo=)
+![Helm: v3](https://img.shields.io/static/v1?label=Helm&message=v3&color=informational&logo=helm)
+
```
zekker6, line 11: Are these changes expected in this PR?
Just want to avoid breaking release bot if that's not expected

zekker6 (commented):
Please, add a changelog entry and check if changes from other changelogs are expected to be in this PR.
Other than that, looks good to me.

charts/victoria-metrics-anomaly/CHANGELOG.md
AndrewChubatiuk: rebasing fixed it

f41gh7 (approved):
LGTM

PR #1624, opened 2024-10-23
common: added vm.podLabels template

AndrewChubatiuk:
added vm.podLabels template, which is used in cluster chart


Haleygo (approved):
LGTM

PR #1625, opened 2024-10-23
vlogs: use service url for single replica and deployment

AndrewChubatiuk:
use k8s service url while using deployments for vlogs or replica count == 1
fixed fluent-bit hot-reload when victorialogs relicacount changes
added temporary workaround till https://github.com/fluent/helm-charts/pull/561 is merged

### DIFF
--- a/charts/victoria-logs-single/templates/extra-outputs.yaml
+++ b/charts/victoria-logs-single/templates/extra-outputs.yaml
@@ -9,9 +9,12 @@ metadata:
   namespace: {{ include "vm.namespace" . }}
   labels: {{- include "vm.labels" $ctx | nindent 4 }}
 data:
-  output_vlogs.conf: |
-    {{- range $i := until (int .Values.server.replicaCount) }}
+  output_vl.conf: |
+    {{- $replicaCount := ternary (int .Values.server.replicaCount) 1 .Values.server.statefulSet.enabled }}
+    {{- range $i := until $replicaCount }}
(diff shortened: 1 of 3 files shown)


charts/victoria-logs-single/templates/extra-outputs.yaml
```diff
@@ -5,13 +5,16 @@
...
 metadata:
-  name: victorialogs-outputs
+  name: vl-outputs
   namespace: {{ include "vm.namespace" . }}
   labels: {{- include "vm.labels" $ctx | nindent 4 }}
 data:
-  output_vlogs.conf: |
-    {{- range $i := until (int .Values.server.replicaCount) }}
+  output_vl.conf: |
+    {{- $replicaCount := ternary (int .Values.server.replicaCount) 1 .Values.server.statefulSet.enabled }}
```
Haleygo, line 13: In statefulsete mode: each victorialogs instance stores all data, and during query, we can get complete data from any instance;
In deployment mode: data is divided among multiple victorialogs instances due to a single write entrance. In order to get complete data, we must query all the victorialogs instances, which requires an additional proxy as the read entry, and the proxy is missing now. 
Should we just disable deployment mode for HA now, since it only returns partial responses?

AndrewChubatiuk: done

Haleygo (approved):

PR #1627, opened 2024-10-24
k8s-stack: fixed default alertmanager template in additionalNotifiersConfig

AndrewChubatiuk:
- fixed default alertmanager rendering while using additionalNotifiersConfig
- added ability to customize vmauth config for vmcluster and externalVM

### DIFF
--- a/charts/victoria-metrics-k8s-stack/values.yaml
+++ b/charts/victoria-metrics-k8s-stack/values.yaml
@@ -246,6 +246,19 @@ additionalVictoriaMetricsMap:
 
 # -- External VM read and write URLs
 externalVM:
+  # -- Custom VMAuth config, url_prefix requires only path, which will be appended to a read and write base URL.
+  # To disable auth for read or write empty list for component config `externalVM.vmauth.<component>: []`
+  vmauth:
+    read:
+      - src_paths:
+          - /select/.*
+        url_prefix:
+          - /
+    write:
+      - src_paths:
+          - /insert/.*
+        url_prefix:
+          - /
   read:
     url: ""
     # bearerTokenSecret:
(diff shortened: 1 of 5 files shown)


charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -722,7 +740,7 @@ vmalert:
 
 vmauth:
   # -- Enable VMAuth CR
-  enabled: false
+  enabled: true
```
zekker6, line 743: Seems like this was added by accident

charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -317,6 +325,15 @@ vmcluster:
...
   annotations: {}
+  # -- Custom VMAuth config, url_prefix requires only path, which will be appended to a select and insert base URL.
+  # To disable auth for vmselect or vminsert empty list for component config `vmcluster.vmauth.<component>: []`
+  vmauth:
+    vmselect:
+      - src_paths:
+          - /select/.*
+        url_prefix:
+          - /
+    vminsert: []
```
zekker6, line 336: I'd suggest enabling both routes by default in order to avoid confusion. 
There is already a comment clarifying how to disable routing for some of the components.

charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -246,6 +246,14 @@ additionalVictoriaMetricsMap:
...
 # -- External VM read and write URLs
 externalVM:
+  # -- VMAuth config sections for external VM urls
+  vmauth:
+    read:
+      - src_paths:
+          - /.*
+        url_prefix:
+          - /
+    write: []
```
zekker6, line 256: I'd suggest enabling both routes by default in order to avoid confusion and add a comment clarifying a way to disable some of paths (similar to the same section in vmcluster config).

zekker6 (approved):

PR #1633, opened 2024-10-25
logs: support multiple path for each host in ingress

AndrewChubatiuk:
support mulitple paths for each ingress host

### DIFF
--- a/charts/victoria-logs-single/values.yaml
+++ b/charts/victoria-logs-single/values.yaml
@@ -219,7 +219,9 @@ server:
     # -- Array of host objects
     hosts: []
     #   - name: vmselect.local
-    #     path: /select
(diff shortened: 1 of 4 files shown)


charts/victoria-logs-single/values.yaml
```diff
@@ -219,7 +219,9 @@ server:
     # -- Array of host objects
     hosts: []
     #   - name: vmselect.local
-    #     path: /select
+    #     path:
+    #       - /select
+    #       - /ui
```
zekker6, line 224: Let's use `/insert` as an example as `/ui` does not exist and might confuse some users.
```suggestion
    #       - /insert
```

zekker6 (approved):

Issue #1641, opened 2024-10-28
feature:


vhajdukd:
**Chart name and version, where you feel a lack of requested feature**
chart: victoria-metrics-agent
version: v0.12.1

**Is your feature request related to a problem? Please describe.**
Allow statefulset to be configured as the target for hpa. One may want to retain tmp queue for gracefulshutdowns during for example image tag changes / upgrades to pods.

**Describe the solution you'd like**
Allow statefulset to be configured as the target for hpa. 

**Describe alternatives you've considered**
Deployment won't fulfil that role as emptyDir will be cleaned (tmp queue) on pods rollout.

**Additional context**
This is to facilitate this part of the documentation:
- https://docs.victoriametrics.com/vmagent/#disabling-on-disk-persistence
```
vmagent still may write pending in-memory data to -remoteWrite.tmpDataPath on graceful shutdown if -remoteWrite.disableOnDiskQueue command-line flag is specified. It may also read buffered data from -remoteWrite.tmpDataPath on startup.
```

AndrewChubatiuk:
hey @vhajdukd 
how are you going to use HPA with VMAgent?

vhajdukd:
Thanks for the fast fix. How we use it ? - This is in case there will be a surge in traffic (ppl are sending to our vmagents) and queue will build up before rollout of new version.

AndrewChubatiuk:
starting the next release HPA will target StatefulSet if you have `horizontalPodAutoscaler.enabled: true` and `statefulset.enabled: true`

vhajdukd:
When can I expect a release of "0.14.4" ?

AndrewChubatiuk:
It's already available

PR #1642, opened 2024-10-28
support HPA for statefulset

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1641


f41gh7 (approved):
LGTM

f41gh7:
It'd be great to mention some where at docs, that `HPA` for vmagent cannot be used to scale `scrapping` workload. In this case VPA should be used.

 `HPA` is really useful for the cases:
1) streaming aggregation. 
2) data ingestion via push protocols
3) data ingestion from kafka/pubsub at ent version

PR #1645, opened 2024-10-29
k8s-stack: use common templates

AndrewChubatiuk:
use common templates for k8s stack chart


charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -36,16 +36,6 @@ victoria-metrics-operator:
     # -- By default, operator converts prometheus-operator objects.
     disable_prometheus_converter: false
 
-serviceAccount:
```
Haleygo, line 39: Would you mind mentioning in the changelog that we removed this? Even though it didn't do anything, it might still be used by some users.

Haleygo (approved):
LGTM

charts/victoria-metrics-k8s-stack/values.yaml
AndrewChubatiuk: added this here https://github.com/VictoriaMetrics/helm-charts/pull/1662

PR #1652, opened 2024-10-29
Drop extraLabels to avoid carrying them over the next render

aacevedoosorio:


AndrewChubatiuk:
Thanks for contribution

AndrewChubatiuk (approved):

PR #1654, opened 2024-10-30
ci: remove already published packages to prevent OCI releases from overwriting

AndrewChubatiuk:
currently all latest OCI packages are overwriting during each run of release job, which causes issues for users, who are using OCI


f41gh7 (approved):
LGTM

PR #1655, opened 2024-10-30
common: replace app label with component for managed labels

AndrewChubatiuk:
use app.kubernetes.io/component label with <char.name>-<app> for managed labels


Haleygo (approved):
LGTM

PR #1658, opened 2024-11-01
operator's crd: fix typo in crd field description

evsigneevk:
solve the issue that prevent correct upgrade or install operator's crd


AndrewChubatiuk (approved):
LGTM, Thank you

PR #1667, opened 2024-11-04
victoria-metrics-cluster: chart cleanup

AndrewChubatiuk:
victoria-metrics-cluster  chart cleanup

### DIFF
--- a/charts/victoria-metrics-cluster/values.yaml
+++ b/charts/victoria-metrics-cluster/values.yaml
@@ -15,8 +15,6 @@ global:
   cluster:
     dnsDomain: cluster.local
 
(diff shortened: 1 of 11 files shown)


charts/victoria-metrics-cluster/values.yaml
```diff
@@ -89,6 +87,7 @@ vmselect:
     envflag.enable: true
     envflag.prefix: VM_
     loggerFormat: json
+    httpListenAddr: "0.0.0.0:8481"
```
zekker6, line 90: Not sure if I understand why this option needs to be added?

AndrewChubatiuk: removed it

zekker6 (approved):

PR #1679, opened 2024-11-05
charts/k8s-stack: fix templating of the labels for VMAlertmanager

zekker6:
`indent` does not add a new line char, which results into templating labels right after `labels`:
```
labels: app.kubernetes...
```
`nindent` adds a new line so labels are properly templated starting from the next line.


f41gh7 (approved):
LGTM

Issue #1660, opened 2024-11-03
[victoria-logs-single] feature: Enable JSON parser


maxpain:
https://t.me/VictoriaMetrics_ru1/61686

PR #1692, opened 2024-11-06
logs: added json parsing

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1660, https://github.com/VictoriaMetrics/helm-charts/issues/1693

### DIFF
--- a/charts/victoria-logs-single/templates/extra-outputs.yaml
+++ b/charts/victoria-logs-single/templates/extra-outputs.yaml
@@ -20,7 +20,7 @@ data:
       Name             http
       Match            kube.*
       Host             {{ include "vm.fqdn" $ctx }}
-      port             9428
+      port             {{ $.Values.server.service.servicePort }}
       compress         gzip
       uri              /insert/jsonline
       format           json_lines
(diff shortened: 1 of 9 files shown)


charts/victoria-logs-single/templates/extra-outputs.yaml
```diff
@@ -20,14 +20,14 @@ data:
       Name             http
       Match            kube.*
       Host             {{ include "vm.fqdn" $ctx }}
-      port             9428
+      port             {{ $.Values.server.service.servicePort }}
```
zekker6, line 23: Could you add this bugfix to a changelog?

charts/victoria-logs-single/values.yaml
```diff
@@ -364,6 +364,16 @@ fluent-bit:
...
+      [FILTER]
+          Name          parser
+          Match         *
+          Parser        json
+          Key_Name      log
+          Reserve_Data  On
+      [FILTER]
+          Name   modify
+          Match  *
+          Rename log msg
```
zekker6, line 376: Will this affect log lines which are not formatted in json?
Will a parser just skip non-json entries?

AndrewChubatiuk: it extracts parsed entries only and removes original fields it it parsed successfully

zekker6: It would be great to clarify this in changelog as the data format will change and users will need to take an action to use an updated data format.

zekker6 (approved):

PR #1704, opened 2024-11-07
fixed release notes path

AndrewChubatiuk:
cannot release new charts as wrong path was mistakenly set in actions conditions


f41gh7 (approved):
LGTM

Issue #1706, opened 2024-11-07
bug: Invalid URL for VMAlert remoteWrite when remoteWriteVMAgent is set to true


filipbnkd:
**Chart name and version**
chart: victoria-metrics-k8s-stack
version: 0.28.2

**Describe the bug**

When you set `remoteWriteVMAgent` to `true` for vmalert component it renders invalid URL for remote write. URL has double `/` at the begging of the path, e.g. `http://vmagent-vm-stack.monitoring.svc:8429//api/v1/write`

Here is screenshot from diff in ArgoCD
![Screenshot 2024-11-07 at 12 13 38](https://github.com/user-attachments/assets/4d58b8b0-3e16-422b-925d-4040563656b3)

**Custom values**
Please provide only custom values (excluding default ones):
```yaml
vmsingle:
  enabled: false

vmcluster:
  enabled: true

vmalert:
  enabled: true
  remoteWriteVMAgent: true

vmagent:
  enbled: true
```

```sh
> helm template vm/victoria-metrics-k8s-stack -f values.yaml | grep "url: http://vmagent"
url: http://vmagent-release-name-victoria-metrics-k8s-stack.monitoring.svc:8429//api/v1/write
```

AndrewChubatiuk:
0.28.3 should contain a fix

PR #1710, opened 2024-11-07
updated common chart

AndrewChubatiuk:
updated common chart dependency in all charts
common chart contains a fix for https://github.com/VictoriaMetrics/helm-charts/issues/1706


zekker6 (dismissed):
LTGM

Could about add an entry to all charts about the dependency update with a link to common chart changelog?
That should help users to discover what has changed/fixed since previous release easier.

zekker6 (approved):

Issue #1713, opened 2024-11-08
[victoria-metrics-k8s-stack] bug: etcd dashboard doesn't work in multicluster setups


maxpain:
**Chart name and version**
chart: victoria-metrics-k8s-stack
version: v0.28.2

**Describe the bug**
<img width="1162" alt="image" src="https://github.com/user-attachments/assets/b32c6225-ba23-43ee-b0e2-1ee99d197aeb">

<img width="775" alt="image" src="https://github.com/user-attachments/assets/de395939-aca5-4ed6-9e74-576ab50eb512">

**Custom values**
Please provide only custom values (excluding default ones):
```yaml
vmagent:
  spec:
    externalLabels:
      cluster: hetzner-hel

grafana:
  sidecar:
    dashboards:
      searchNamespace: ALL
      multicluster: true
```

PR #1714, opened 2024-11-08
fixed etcd dashboard cluster variable

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1713


hack/rules-and-dashboards/sync_dashboards.py
```diff
@@ -258,13 +258,13 @@ def patch_dashboard(dashboard, name):
                 variable["hide"] = (
                     "[[ ternary 0 2 $Values.grafana.sidecar.dashboards.multicluster ]]"
                 )
+                query = json.dumps(variable["query"])
+                print(query)
```
zekker6, line 262: Seems like `print` left after debugging.

zekker6 (approved):

PR #1718, opened 2024-11-08
vlogs: updated log agent in readme

AndrewChubatiuk:
updated readme for victoria-logs-single chart


charts/victoria-logs-single/README.md
```diff
@@ -18,9 +18,9 @@ This chart will do the following:
 * (optional) Rollout [fluentbit](https://fluentbit.io/) to collect logs from pods.
```
zekker6, line 18: This line needs to be updated as well

Issue #1721, opened 2024-11-08
[victoria-logs-single] feature: PodMonitor, prometheus rules and Grafana dashboard for Vector


maxpain:
:D

AndrewChubatiuk:
according to [docs](https://github.com/vectordotdev/helm-charts/tree/develop/charts/vector)
`vector.podMonitor.enabled: true` enables podmonitor
rules and dashboard are expected to be a part of victoria-logs-k8s-stack chart

PR #1728, opened 2024-11-08
vlogs: added vector dashboard

### LINKED ISSUE #1727: [victoria-logs-single] bug: duplicate entries for key [name="VECTOR_SELF_NODE_NAME"]
```Could not determine release state: unable to determine cluster state: DaemonSet/logging/vls-vector dry-run failed: failed to create typed patch object (logging/vls-vector; apps/v1, Kind=DaemonSet): .spec.template.spec.containers[name="vector"].env: duplicate entries for key [name="VECTOR_SELF_NODE_NAME"]```

I checked the generated `DaemonSet`, and indeed, it has a duplicated `VECTOR_SELF_NODE_NAME` key:
```yaml
containers:
- name: vector
  image: timberio/vector:0.42.0-distroless-libc
  args:
    - '--config-dir'
    - /etc/vector/
  env:
    - name: VECTOR_LOG
      value: info
    - name: VECTOR_SELF_NODE_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: spec.nodeName
    - name: VECTOR_SELF_NODE_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: spec.nodeName
    - name: VECTOR_SELF_POD_NAME
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.name
    - name: VECTOR_SELF_POD_NAMESPACE
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: metadata.namespace
```

This key is already included here:
https://github.com/vectordotdev/helm-charts/blob/develop/charts/vector/templates/_pod.tpl#L67

AndrewChubatiuk:
fixes #1727, #1721

### DIFF
--- a/charts/victoria-logs-single/values.yaml
+++ b/charts/victoria-logs-single/values.yaml
@@ -11,7 +11,7 @@ global:
   compatibility:
     openshift:
       adaptSecurityContext: "auto"
-  # -- Global name override
+  # -- Override chart name
   nameOverride: ""
   cluster:
     # -- K8s cluster domain suffix, uses for building storage pods' FQDN. Details are [here](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/)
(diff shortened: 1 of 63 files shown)


zekker6 (approved):

charts/victoria-logs-single/values.yaml
```diff
@@ -307,15 +307,16 @@ vector:
   role: Agent
   dataDir: /vector-data-dir
   resources: {}
+  podMonitor:
+    enabled: true
```
zekker6, line 311: I would suggest keeping it disabled by default as it requires an external CRD to be present in cluster.

AndrewChubatiuk: updated

f41gh7 (approved):
LGTM

Issue #1730, opened 2024-11-09
bug: no option to skip CRD installation by operator chart


gecube:
After upgrade to the latest operator chart I see the next:

https://github.com/VictoriaMetrics/helm-charts/blob/2e8679049f593f61d19f2f57cb9adee603a26b3c/charts/victoria-metrics-operator/values.yaml#L35

so no option to skip installation of CRDs at all.

### Desired behavior

Have an option to skip CRDs installation at all (as I want to manage them by myself), so introduce additional config variable like

```
crds:
  enabled: true # or false
```

PR #1731, opened 2024-11-11
operator: added crds.enabled flag

AndrewChubatiuk:
also reverted removal of option to disable CRD in operator. fixes https://github.com/VictoriaMetrics/helm-charts/issues/1730

### DIFF
--- a/charts/victoria-metrics-operator/values.yaml
+++ b/charts/victoria-metrics-operator/values.yaml
@@ -30,6 +30,8 @@ image:
   pullPolicy: IfNotPresent
 
 crds:
+  # -- manages CRD creation. Disables CRD creation only in combination with `crds.plain: false` due to helm dependency conditions limitation
+  enabled: true
   # -- check if plain or templated CRDs should be created.
   # with this option set to `false`, all CRDs will be rendered from templates.
   # with this option set to `true`, all CRDs are immutable and require manual upgrade.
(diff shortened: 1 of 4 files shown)


charts/victoria-metrics-operator/values.yaml
```diff
@@ -30,6 +30,8 @@ image:
   pullPolicy: IfNotPresent
 
 crds:
+  # -- manages CRD creation. Disables CRD creation only in combination with `crds.plain: false` due to helm dependency conditions limitation
```
Haleygo, line 33: Why not have both `crds.plain` and `crds.enabled` as the conditions of dependency `crds`?
```
  - name: crds
    version: "0.0.*"
    condition: crds.plain, crds.enabled
```

Haleygo: ohh, nvm, it's "or" combination.

Haleygo, line 33: Can we throw a error directly for  `crds.enabled: false` and `crds.plain: true` then, to prevent incorrect usage? It should be fine since the default `crds.plain` is false.

AndrewChubatiuk: comma in `condition` implements `OR` operation

AndrewChubatiuk: so it will not help

Haleygo: yeah, I mean to add an additional check somewhere ourselves, for `crds.enabled: false` with`crds.plain: true`.

AndrewChubatiuk: updated

f41gh7 (approved):
LGTM

PR #1732, opened 2024-11-11
common: added overrideKey to support different override keys for plain an managed resources, clarified params in docs

AndrewChubatiuk:
use `name` overrideKey for managed resources and `fullnameOverride` for plain ones
removed impact of `name` param on plain resources
clarified fullnameOverride, nameOverride and name params in docs


Haleygo (approved):
LGTM

PR #1736, opened 2024-11-12
vmagent: fixed statefulset variable name typo, new releases for vmalert and vmagent

AndrewChubatiuk:
new vmalert and vmagent chart releases, fixed `.Values.statefulSet` variable name in templates


f41gh7 (approved):
LGTM

Issue #1563, opened 2024-10-07
bug: removal of VM k8s chart does not work well.


gecube:
Good day!

The issue is observed on the all versions of `victoria-metrics-k8s-stack` chart.

The steps to reproduce:

1. install `victoria-metrics-k8s-stack` 
2. remove ns where `victoria-metrics-k8s-stack` is installed
3. as the order of deletion of the objects in ns is not defined, there is a possibility that operator would be removed earlier than the resources it manages, effectively making ns stuck in `Terminating` state.

Possible solutions:
- not recommend to install both operator and CRs it manages in the same NS
- split operator chart and CRs into different charts and/or namespaces and recommended such an installation

Also I got the same result when I tried to remove the `victoria-metrics-k8s-stack` chart without ns deletion.
1. install `victoria-metrics-k8s-stack` with FluxCD
2. remove `HelmRelease` from gitops repo and/or invoke `kubectl delete helmrelease -n monitoring-system   victoria-metrics-k8s-stack`
3. get pods stuck

```bash
kubectl get pods -A
...
monitoring-system   vmagent-victoria-metrics-k8s-stack-5d4d45898d-47qwx    2/2     Running            2 (42m ago)   4d19h
monitoring-system   vmalert-victoria-metrics-k8s-stack-846689d7db-r6dh6    2/2     Running            2 (42m ago)   4d19h
monitoring-system   vmalertmanager-victoria-metrics-k8s-stack-0            0/2     Unknown            0             4d22h
monitoring-system   vmsingle-victoria-metrics-k8s-stack-7565649d7c-7v624   1/1     Running            1 (42m ago)   4d19h
```

From written above I am concluding that the process of chart's removal was not testes and it is subject for race conditions (if operator is accidentally removed BEFORE CRs it manages, they stuck).

I am kindly asking to make an investigation and confirm the bug.

AndrewChubatiuk:
hey @gecube 
this behaviour is expected, CRs should not be removed together with operator. To remove CRs operator chart has delete hook, which removes CRs (`victoria-metrics-operator.crd.cleanup.enabled` option in k8s-stack chart), but I have no idea how it works with Flux CD. Regarding CRs deployment in another namespace, you can always set `victoria-metrics-operator.namespaceOverride` to run operator in a separate namespace

gecube:
@AndrewChubatiuk there is a difference between CRD and CR. So let's don't confuse them.

[...]

PR #1744, opened 2024-11-13
operator: disable cleanup, when crds.enabled: false

AndrewChubatiuk:
disable cleanup hook, when crds.enabled: false. related issue https://github.com/VictoriaMetrics/helm-charts/issues/1563, alternatively we can set `victoria-metrics-operator.crds.enabled: false` in k8s-stack chart by default


Haleygo:
The fix looks good to me, but I don't understand how it related to https://github.com/VictoriaMetrics/helm-charts/issues/1563#issuecomment-2464954569, I would expect to see stucked cleanup job pod if that's the case.

>alternatively we can set victoria-metrics-operator.crds.enabled: false in k8s-stack chart by default

I don't think so, many user install victoria-metrics-k8s-stack directly and vm CRDs are needed.

AndrewChubatiuk:
it's not a fix for an issue, but related to it

zekker6 (approved):

PR #1763, opened 2024-11-15
Added default flag "rule" and commented subPath 

xogoodnow:
Neither of the included parameters change the current behavior so no breaking changes.

Reason to add "rule" flag:
this is already set and since almost every user would need to modify this path. So including the flag in the values.yaml is very convenient.

Reason to add subPath:
Consistency with other sections. Conveniently showing how to mount the volume as a file instead of a directory.

### DIFF
--- a/charts/victoria-metrics-alert/values.yaml
+++ b/charts/victoria-metrics-alert/values.yaml
@@ -178,6 +178,7 @@ server:
     envflag.enable: "true"
     envflag.prefix: VM_
     loggerFormat: json
+    rule: "/config/alert-rules.yaml"
 
   # -- Additional hostPath mounts
   extraHostPathMounts:
@@ -200,6 +201,8 @@ server:
     []
     # - name: example
     #   mountPath: /example
+    #   subPath: ""
+
 
   # -- Additional containers to run in the same pod
   extraContainers:


AndrewChubatiuk:
rule flag is added from templates and there's no need to set it explicitly in parameters
extraVolumeMounts represents kubernetes pod volume mounts section and it makes more sense to add to comments above the `extraVolumeMounts` parameter a link to https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.26/#volumemount-v1-core instead

xogoodnow:
Hi @AndrewChubatiuk,

Regarding the "rule" flag:

As almost no one uses the single alert file which is provided by default, it is required to change the rule flag
so by nature it needs to be dynamic and having the option within the template itself is confusing.
but when the flag is provided within the values.yaml file, users can easily change the parameter without any modifications to the templates.

IMHO, when using helm charts, almost all modifications must be made to the values.yaml and changing the template for an official chart is a last resort.



Regarding the subpath comment:
As you can see here this is mentioned in k8s docs, specially when using config maps the option subpath could come in handy.
So I think this is required in both sections (host path and extra volumes)

ps: https://kubernetes.io/docs/concepts/storage/volumes/#using-subpath

Cheers

AndrewChubatiuk:
regarding "flag" I've got your point and moved parameter to values as a list and you can override it with a slice of rule files 
regarding subpath I've added a comment to extraVolume mounts to about where to find all supported params including subPath

xogoodnow:
I see that you opened and merged another PR
https://github.com/VictoriaMetrics/helm-charts/pull/1764

May I ask why didn't you just merge my PR ? :neutral_face:

AndrewChubatiuk:
> May I ask why didn't you just merge my PR ?

sorry, for me it's faster to implement than to explain :)

xogoodnow:
This is so ridiculously unprofessional that I don't even know what to say.

Hi @zekker6, 
I had the pleasure of meeting you in our enterprise features assessment meeting.
I just wanted to mention someone about this ridiculous and arrogant attitude.

Will definitely take this matter into account in the next planning quarter.

AndrewChubatiuk:
@xogoodnow 
first of all thank you for a feedback
sad to hear it watching initial issue you were trying to resolve was mostly related to chart user experience and I've slightly updated your idea in alternative PR as having more experience with this chart as a developer

xogoodnow:
First of all you are welcome.
Second:
My issue is not technical at all. As it is customary within the open source community, you could have provided me with the requirements as you see fit so the PR could have been updated or at the very least add the changes to the same branch and simply merge it.
Opening and closing an alternative branch without any notice or mention (IMHO) is very unprofessional.

AndrewChubatiuk (approved):

hagen1778:
Hey @xogoodnow!
Thanks for your involvement and contributions to the project!
Regarding to the following [comment](https://github.com/VictoriaMetrics/helm-charts/pull/1763#issuecomment-2478462110) let me assure you that our intentions are always focused on the project improvement. Sometimes, communication might take too much time and slow-down the development pace. So project maintainers can decide what needs to be done and how. In the end, maintenance and support responsibility is on their shoulders.

I agree with your other [comment](https://github.com/VictoriaMetrics/helm-charts/pull/1763#issuecomment-2478571811) - good communication is a must. We should have let you know about related changes or made decisions. Here I am on your side. I hope we'll handle it better next time. 

PRs like [this one](https://github.com/VictoriaMetrics/helm-charts/pull/1764) should have better description, mentioning what was changed, why it was changed and what are related tickets or issues. Otherwise, we lose the context of changes and transparency. JFYI @AndrewChubatiuk @f41gh7

xogoodnow:
Hi @hagen1778, 

Thank you for your response.

I completely understand and as I mentioned my issue with the response was non technical. This is a very simple PR and there are no complexities on the technical front.
I think this [comment](https://github.com/VictoriaMetrics/helm-charts/pull/1763#issuecomment-2478442412) could have been handled in a much more professional and respectful manner (response to why there was another PR.

In any case, thank you for your time and I'm glad to know that the approach of the company is quite reasonable. 


Cheers

Issue #1778, opened 2024-11-19
bug: vmcluster name can no longer be overridden


dctrwatson:
**Chart name and version**
chart: victoria-metrics-k8s-stack
version: v0.28.4

**Describe the bug**
The VMCluster CR is now using the fullname from the Chart instead of what is defined in values.vmcluster.name

**Custom values**
Please provide only custom values (excluding default ones):
```yaml
vmcluster:
  name: my-custom-name
```

AndrewChubatiuk:
thanks for an issue
fix is expected in a next release

PR #1780, opened 2024-11-20
k8s-stack: allow CR name override using component.name parameter

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1778


f41gh7 (approved):
LGTM

PR #1783, opened 2024-11-20
common: render template in vm.app.name

AndrewChubatiuk:
render generated name in `vm.app.name` template
minor refactoring

### DIFF
--- a/charts/victoria-logs-single/values.yaml
+++ b/charts/victoria-logs-single/values.yaml
@@ -11,11 +11,13 @@ global:
   compatibility:
     openshift:
       adaptSecurityContext: "auto"
-  # -- Override chart name
-  nameOverride: ""
   cluster:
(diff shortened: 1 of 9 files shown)


charts/victoria-logs-single/values.yaml
```diff
@@ -11,11 +11,13 @@ global:
   compatibility:
     openshift:
       adaptSecurityContext: "auto"
-  # -- Override chart name
```
Haleygo, line 14: As I understand, you want to remove all `.global.nameOverride` usage, would you mind mentioning this in changelog?

Haleygo (approved):

charts/victoria-logs-single/values.yaml
AndrewChubatiuk: just removing it from docs to not confuse users

Haleygo: I don't undersatnd, it's the values file you're changing, now both `.nameOverride` and `.global.nameOverride` options work, but only `.nameOverride` shows in the default values files we provide.

Issue #1809, opened 2024-11-25
bug: DiskRunsOutOfSpace alert does not work


dctrwatson:
**Chart name and version**
chart: victoria-metrics-k8s-stack
version: v0.28.3

**Describe the bug**
As defined `DiskRunsOutOfSpace` will never fire if following the recommendation for `storage.minFreeDiskSpaceBytes` to be at least 20%

**Custom values**

dctrwatson:
> hey @dctrwatson do you expect alert threshold to be connected to `storage.minFreeDiskSpaceBytes` value?

I think so. That way it would work if you set it to 20% or even 35%.

I'm not sure how you could that in Helm though. We purposely set that flag to be an integer so that we can use `label_value(flag{job=~"vmstorage-.*", name="retentionPeriod"}, "value") * 31d` for a capacity alert.

lieberlois:
I don't have the same issue, but as an idea: I think it would adding the PVC Dashboard from the kube-prometheus-stack to the VictoriaMetrics dashboards by default would help a lot with analyzing such alerts

AndrewChubatiuk:
@dctrwatson 
in 0.30.1 added `NodeBecomesReadonlyIn3Days` alert, which should notify, when storage becomes read-only in 3 days

PR #1814, opened 2024-11-26
k8s-stack: added workaround for DiskRunsOutOfSpace alerts for cases, when storage.minFreeDiskSpaceBytes is more than 20% of disk storage

AndrewChubatiuk:
as we don't want to break compatibility with prometheus for dashboards at VictoriaMetrics repo, added hack that takes into an account `storage.minFreeDiskSpaceBytes` flag value
fixes https://github.com/VictoriaMetrics/helm-charts/issues/1809

### DIFF
--- a/charts/victoria-metrics-k8s-stack/files/rules/generated/vmcluster.yaml
+++ b/charts/victoria-metrics-k8s-stack/files/rules/generated/vmcluster.yaml
@@ -35,7 +35,8 @@ rules:
     sum(vm_data_size_bytes) by (job,instance,{{ $clusterLabel }}) /
     (
      sum(vm_free_disk_space_bytes) by (job,instance,{{ $clusterLabel }}) +
-     sum(vm_data_size_bytes) by (job,instance,{{ $clusterLabel }})
+     sum(vm_data_size_bytes) by (job,instance,{{ $clusterLabel }}) -
+     sum(label_value(flag{name="storage.minFreeDiskSpaceBytes"}, "value")) by (job,instance,{{ $clusterLabel }})
     ) > 0.8
   for: 30m
   labels:
--- a/hack/rules-and-dashboards/sync_rules.py
+++ b/hack/rules-and-dashboards/sync_rules.py
@@ -166,6 +166,10 @@ def cluster_label_var(mo):
 
 
 replacement_map = {
+    "( |\n|\r)+\\) > 0.8": {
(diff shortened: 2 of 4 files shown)


charts/victoria-metrics-k8s-stack/files/rules/generated/vmcluster.yaml
```diff
@@ -35,7 +35,8 @@ rules:
     sum(vm_data_size_bytes) by (job,instance,{{ $clusterLabel }}) /
     (
      sum(vm_free_disk_space_bytes) by (job,instance,{{ $clusterLabel }}) +
-     sum(vm_data_size_bytes) by (job,instance,{{ $clusterLabel }})
+     sum(vm_data_size_bytes) by (job,instance,{{ $clusterLabel }}) -
+     sum(label_value(flag{name="storage.minFreeDiskSpaceBytes"}, "value")) by (job,instance,{{ $clusterLabel }})
```
dctrwatson, line 39: when I was adding our own alert for this in the meantime, I saw that `vm_free_disk_space_limit_bytes` already exists for this flag 😍

dctrwatson (commented):
thank you for the workaround!

hack/rules-and-dashboards/sync_rules.py
```diff
@@ -166,6 +166,10 @@ def cluster_label_var(mo):
 
 
 replacement_map = {
+    "( |\n|\r)+\\) > 0.8": {
```
Haleygo, line 169: I prefer not to have such hack for rules belong to victoriametrics, but fix the upstream if it's not expected and avoid inconsistencies between both sources.

Haleygo:
If users set `-storage.minFreeDiskSpaceBytes` to 20% or 30% of disk total, the `DiskRunsOutOfSpace` alert won’t trigger, which makes sense since there’s enough disk space. 
However, with the proposed change, if the total disk size is 1000G and `-storage.minFreeDiskSpaceBytes=200G`, the alert would trigger when VM uses 640G, which is unreasonable.

I think the alert is for smaller `-storage.minFreeDiskSpaceBytes` setting. For example, if the total disk size is 1000G and `-storage.minFreeDiskSpaceBytes=100G`, the alert warns users when VM uses 800G.

AndrewChubatiuk:
closing this for https://github.com/VictoriaMetrics/VictoriaMetrics/pull/7670

hack/rules-and-dashboards/sync_rules.py
AndrewChubatiuk: label_value is not supported by promql, but I've opened another PR, which uses metrics, that reports a value of this flag

PR #1862, opened 2024-12-11
Omitted server-enabled

xogoodnow:
Hi,

Since no one would be using this chart if they did not want the vm-alert server, having this condition is not required.

Cheers


AndrewChubatiuk:
thanks for contribution!

PR #1869, opened 2024-12-16
Fix values rendering for Markdown engines with strict XML parsing

mindw:
* replace PWD with REPODIR in Makefile to fix run on PyCharm remote WSL2

### DIFF
--- a/hack/docs/template.tmpl
+++ b/hack/docs/template.tmpl
@@ -7,7 +7,7 @@
 {{- $notationType = "plaintext" }}
 {{- end -}}
 {{- end -}}
-<pre class="helm-vars-default-value" language-yaml" lang="{{ $notationType }}">
+<pre class="language-yaml" lang="{{ $notationType }}">
 <code class="language-yaml">
 {{- if (eq $notationType "tpl" ) -}}
 {{ .Key }}: |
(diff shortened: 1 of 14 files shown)


hack/docs/template.tmpl
```diff
@@ -7,7 +7,7 @@
 {{- $notationType = "plaintext" }}
 {{- end -}}
 {{- end -}}
-<pre class="helm-vars-default-value" language-yaml" lang="{{ $notationType }}">
```
AndrewChubatiuk, line 10: please revert this change everywhere as we're relying on this class in docs website

mindw: @AndrewChubatiuk , Would `class="language-yaml"` do instead? what it now isn't a legal XML ...

AndrewChubatiuk: why is this important? where do you use it?

AndrewChubatiuk:
@mindw Thanks for contribution!

hack/docs/template.tmpl
mindw: My apologies - I wanted to suggest `helm-vars-default-value` - had a brain meltdown :( . 
Also, My CSS /XHTML is extremely hence missing options that there was an extra `"` .

PR #1896, opened 2024-12-23
charts/operator: Add option to enable hostNetwork for custom CNI deployments

tanmay-bhat:
- Add an option to enable hostNetwork on operator deployment to work with custom CNI on managed Kubernetes clusters such as EKS.


AndrewChubatiuk:
thanks for contribution!

PR #1902, opened 2024-12-26
Fix "minim number os seconds" typo

konstantin-kornienko:


zekker6:
Hello @konstantin-kornienko, thank you for the PR!

CRD fields descriptions are automatically generated from operator source code during the operator release, so this change will be overwritten after the next release.

Could you send a PR to operator repository so that the source of the typo will be fixed?
Here is an example line in operator repo which you'd need to fix - https://github.com/VictoriaMetrics/operator/blob/9cf74c2846a61768379ba09b0ed42eebda439c81/api/operator/v1beta1/vmextra_types.go#L1214

AndrewChubatiuk:
https://github.com/VictoriaMetrics/operator/pull/1208

Issue #1960, opened 2025-02-01
bug: VM operator `securityContext` missing


tropnikovvl:
**Chart name and version**
chart: victoria-metrics-operator
version: 0.41.0

**Describe the bug**
By default option [securityContext](https://github.com/VictoriaMetrics/helm-charts/blob/master/charts/victoria-metrics-operator/values.yaml#L97) is enabled, but in Kubernetes I see that it is empty.

```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "1"
    meta.helm.sh/release-name: vm-stack
    meta.helm.sh/release-namespace: vm-stack
  creationTimestamp: "2025-01-29T11:27:39Z"
  generation: 1
  labels:
    app.kubernetes.io/instance: vm-stack
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: victoria-metrics-operator
    app.kubernetes.io/version: v0.52.0
    helm.sh/chart: victoria-metrics-operator-0.41.0
    helm.toolkit.fluxcd.io/name: vm-stack
    helm.toolkit.fluxcd.io/namespace: vm-stack
  name: vm-stack-victoria-metrics-operator
  namespace: vm-stack
  resourceVersion: "291439363"
  uid: 8df6c9b2-d54f-466b-a290-f6dc32be77d7
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app.kubernetes.io/instance: vm-stack
      app.kubernetes.io/name: victoria-metrics-operator
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/instance: vm-stack
        app.kubernetes.io/managed-by: Helm
        app.kubernetes.io/name: victoria-metrics-operator
    spec:
      automountServiceAccountToken: true
      containers:
      - args:
        - --zap-log-level=info
        - --leader-elect
        - --health-probe-bind-address=:8081
        - --webhook.enable=true
        env:
        - name: POD_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: metadata.name
        - name: OPERATOR_NAME
          value: victoria-metrics-operator
[...]

PR #1963, opened 2025-02-04
operator: added default security context

AndrewChubatiuk:
- operator chart has `securityContext.enabled: true`, but it has no impact as it has no security context defined. adding default one. fixes https://github.com/VictoriaMetrics/helm-charts/issues/1960
- adding GrafanaDatasource support as an alternative to configMap. fixes https://github.com/VictoriaMetrics/helm-charts/issues/1964
- fixed minor issue in vm.securityContext common template

### DIFF
--- a/charts/victoria-metrics-common/CHANGELOG.md
+++ b/charts/victoria-metrics-common/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next release
 
-- TODO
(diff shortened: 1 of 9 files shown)


zekker6 (commented):
Could you add a changelog for this change?

AndrewChubatiuk:
@zekker6 updated

zekker6 (approved):

Issue #1995, opened 2025-02-13
feature: vmauth and vmagent as Daemonset


szibis:
**Chart name and version, where you feel a lack of requested feature**
chart: victoria-metrics-agent
version: v0.15.8

chart: victoria-metrics-auth
version: v0.8.7

**Is your feature request related to a problem? Please describe.**

**Network cost savings**
Those using OpenTelemetry operator or collector chart in Damonset mode and sending data to VM would be great in bigger setups to support vmagent and vmauth as Daemonsets.
We know that VMagent can be sent to other vmagent/cluster via VictoriaMetrics RemoteWrite protocol, which uses zstd similar to RemoteWrite 2.0, and this saves on network traffic, especially with inter-AZ communications. For example, this factor is important in bigger organizations using AWS, and network costs can be significant.
When we use OpenTelemetry collector as daemonset and vmauth/vmagent as deployment/statefulset, we also need to send from other nodes than the VM components that are running, which means we introduce additional network cost.

**Stream Aggregation sharding** 
Additional case sharding. Opentelemetry collector does not use the same sharding as vmagent, and we cannot send properly sharded data for stream aggregation to vmagent statefulset/deployment.

**Describe the solution/alternatives you'd like**

**Network cost savings**
This can be mitigated by using OTLP to VM with zstd compression because with RemoteWrite, we will use Snappy, which increases the cost.
When we need to use something different than OTLP (there is multiple reasons behind this) for example RemoteWrite or different ingestion protocols then having local on same node vmauth/vmagent will allow use localy less efficient option, but later from vmagent to cluster or another vmagent deployment/statefulset we could send using VictoriMetrics RemoteWrite.
The additional cost of resources - vmagent/vmauth are very small in consumption - is smaller than network costs in high throughput installations.

**Stream Aggregation sharding** 
When we will have local vmagent daemonset we will send on same node to vmagent via Service with `internalTrafficPolicy: Local` and this vmagent then can send properly sharded data to external statefulset/deployment which will guarantee always same destination for specific metrics from all nodes in the cluster - means all metric producer pods in cluster.

**Additional context**
![Image](https://github.com/user-attachments/assets/92b19aca-9c37-4a3f-8538-6615b831c8b9)

AndrewChubatiuk:
hey @szibis 
could you explain a purpose of vmauth in your setup?
[...]

PR #1998, opened 2025-02-17
vmagent: added daemonset support

AndrewChubatiuk:
related issue https://github.com/VictoriaMetrics/helm-charts/issues/1995
also did the same change in other charts, which were using `statefulSet.enabled`


charts/victoria-metrics-cluster/CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 ## Next release
 
-- TODO
+**Update note 1**: `.Values.vmselect.statefulSet.updateStrategy` was changed to `.Values.vmselect.statefulSet.spec.updateStrategy`
+**Update node 2**: `.Values.vmselect.podManagementPolicy` was changed to `.Values.vmselect.statefulSet.spec.podManagementPolicy`
+**Update note 3**: `.Values.vmselect.deployment.strategy` was changed to `.Values.vmselect.deployment.spec.strategy`
+**Update note 4**: `.Values.vmselect.statefulSet.enabled` was replaced by `.Values.vmselect.mode`, which accepts `deployment`, `statefulSet` or `daemonSet` values
```
f41gh7, line 6: Can you please add any real world example when vmselect should started as daemonSet?

AndrewChubatiuk: just updated this record

charts/victoria-logs-single/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,16 @@
+{{- $sa := .Values.serviceAccount }}
```
zekker6, line 1: Why does VL pod need a service acount?
It does not use any kind of access to k8s APIs, so it doesn't seem like it is needed

zekker6: Wdyt about disabling it by default than?
Seems like it might only be useful in some rare cases, so keeping it disabled by default will make impact on users minimal but will still allow to use it when needed.

AndrewChubatiuk: done

zekker6 (approved):

charts/victoria-metrics-agent/templates/server.yaml
```diff
@@ -1,113 +1,124 @@
-{{- if .Values.statefulSet.enabled }}
```
mkilchhofer, line 1: This is sadly another breaking change within a Minor upgrade :(
@AndrewChubatiuk

PR #2009, opened 2025-02-24
generate paragraph per chart value instead of table

AndrewChubatiuk:
use paragraph per chart value instead of table to have less visualization issues in docs site and to have links to these values, which can be useful to reference in github issues

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 HELM_IMAGE = alpine/helm:3.16.2
-HELM_DOCS_IMAGE = local/helm-docs:v1.14.144
+HELM_DOCS_IMAGE = local/helm-docs:v1.14.2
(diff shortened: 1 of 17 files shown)


Haleygo:
image
Should the format `key` be displayed here, it confuses me with the values, can we remove it?
And the new format doesn't have a good place to show the default value.

AndrewChubatiuk:
without `<key>` don't know how to show a default value, which accepts json string, like this one 
```
key: |
  foo: bar
```
removing key will make this information less informative
```
foo: bar
```
or
```
|
  foo: bar
```

AndrewChubatiuk:
also i would rather target a look in docs site, instead of one in a repo
image
image

AndrewChubatiuk:
removed `<key>` prefix

PR #2021, opened 2025-02-28
charts/victoria-metrics-k8s-stack: add example of using quay registry

zekker6:


AndrewChubatiuk:
why not just use `VM_CONTAINERREGISTRY` instead?

charts/victoria-metrics-k8s-stack/README.md
```diff
@@ -197,6 +197,41 @@ defaultDashboards:
...
+It is possible to override default image registry for all components deployed by operator and operator itself by using the following values:
+```yaml
+victoria-metrics-operator:
+  image:
+    registry: "quay.io"
+  env:
+    - name: "VM_USECUSTOMCONFIGRELOADER"
+      value: "true"
+    - name: VM_CUSTOMCONFIGRELOADERIMAGE
+      value: "quay.io/victoriametrics/operator:config-reloader-v0.53.0"
```
Haleygo, line 212: I just realize we put`config-reloader` into the tag instead of name, which looks inconsistent from others.
@f41gh7 should we rename the image to `victoriametrics/(operator-)config-reloader`? We can push both of them for some time.

Haleygo (approved):

charts/victoria-metrics-k8s-stack/README.md
f41gh7: Yes, I think we can do that. And it's better just stop pushing previous version and use new naming instead.

PR #2069, opened 2025-03-19
charts/vmanomaly: add support of sharding

zekker6:
- upgrade anomaly to v1.21.0
- add support of sharding and high availability (replication)


fred-navruzov (approved):
Thanks, @zekker6 , LGTM!

PR #2072, opened 2025-03-20
add ability to disable vmselect and vminsert.

### LINKED ISSUE #2059: feature: make `VMSelect` optional in `victoria-metrics-k8s-stack` helmchart
**Chart name and version**
chart: [victoria-metrics-k8s-stack](https://github.com/VictoriaMetrics/helm-charts/tree/master/charts/victoria-metrics-k8s-stack)
version: 0.38.3

**Describe the bug**
VMSelect is an optional component when using a CRD directly but helmchart doesnt allow us to skip VMSelect in the `values.yaml`

**Custom values**
Please provide only custom values (excluding default ones):
```yaml
vmcluster:
  spec:
    vmselect:
      enabled: false // Possible value to introduce
```

RIght now the only option is to set `vmcluster.vmselect.replicas: 0` which is again not a workaround as it fails the update flow since the operator does not proceed on updating the vminsert compnent as it does not find any VMselect replicas.
In such case we need to update the replicas for VMSelect to 1 and then finish the update and again shift the replicas to 0
More information in the slack thread - https://victoriametrics.slack.com/archives/CGZF1H6L9/p1741322230457589

AndrewChubatiuk:
add .Values.vmcluster.spec.vmselect.enabled and .Values.vmcluster.spec.vminsert.enabled to disable vmselect and vminsert. fixes #2059

### DIFF
--- a/charts/victoria-metrics-k8s-stack/CHANGELOG.md
+++ b/charts/victoria-metrics-k8s-stack/CHANGELOG.md
@@ -1,8 +1,10 @@
 ## Next release
 
-- updated dashboards and rules
-- updated common dependency 0.0.39 -> 0.0.42
-- fixed additional notifier config rendering. See [this issue](https://github.com/VictoriaMetrics/helm-charts/issues/2075).
+- update dashboards and rules
+- update common dependency 0.0.39 -> 0.0.42
+- set default vmalertmanager image to 0.28.1
+- fix additional notifier config rendering. See [this issue](https://github.com/VictoriaMetrics/helm-charts/issues/2075).
+- add .Values.vmcluster.spec.vmselect.enabled and .Values.vmcluster.spec.vminsert.enabled flags to disable vmselect or vminsert. See [this issue](https://github.com/VictoriaMetrics/helm-charts/issues/2059).
 
 ## 0.39.0
 
--- a/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
+++ b/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
@@ -5,7 +5,7 @@
   {{- if $Values.vmsingle.enabled -}}
     {{- $_ := set . "appKey" (list "vmsingle" "spec") -}}
     {{- $_ := set $endpoint "url" (include "vm.url" .) -}}
-  {{- else if $Values.vmcluster.enabled -}}
+  {{- else if and $Values.vmcluster.enabled $Values.vmcluster.spec.vmselect.enabled -}}
     {{- if $Values.vmauth.enabled -}}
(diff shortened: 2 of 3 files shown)


charts/victoria-metrics-k8s-stack/CHANGELOG.md
```diff
@@ -1,7 +1,9 @@
 ## Next release
 
-- Updated dashboards and rules
-- updated common dependency 0.0.39 -> 0.0.42
+- update dashboards and rules
+- update common dependency 0.0.39 -> 0.0.42
+- set default vmalertmanager image to 0.28.1
+- add ability to disable vmselect or vminsert by setting replicaCount to 0. See [this issue](https://github.com/VictoriaMetrics/helm-charts/issues/2059).
```
Haleygo, line 6: It's a breaking change for users who have already set `replicaCount=0` already, as the update will cause vm-operator to remove the related resources.

AndrewChubatiuk: added vmselect.enabled and vminsert.enabled parameters instead

kartik-moolya: IMO, enabled flag would be an overkill (although I suggested it in the beginning).
For users who have set VMSelect replicas to 0, it would just delete the vmselect sts right ?
Also setting the replicas to 0 puts the VMCluster in failed state which isnt ideal.

AndrewChubatiuk: neither me nor @Haleygo were able to reproduce a cluster failure. it should be fixed in operator if there is an issue. could you please share operator logs, chart parameters to identify an error?

kartik-moolya: Reproducing is easy.
* Deploy a VMCluster with VMSelect running 0 replicas
* Cluster should deploy fine with status either `operational` / `failed`
* Once deployed make any change to VMStorage, VMInsert configuration which will induce a rollout of both those components.

You will see the below behavior
* The change will rollout to VMStorage pods
* The change will never rollout to the VMInset pods as the operator will keep finding VMSelect pods before rolling out VMInsert
* Eventually the operation will timeout and cluster would end up in failed state
* Now if you update the VMSelect replicas to 1, the operator will find what it was looking for and eventually rollout the changes for VMInsert as well

AndrewChubatiuk: I wasn't able to reproduce, I've checked code, in a latest release operator skips vmselect if replicaCount is 0

Haleygo (approved):
LGTM

PR #2102, opened 2025-04-03
feat: Add posibility to template ingress annotations

sergeyshevch:
Annotations templating can be useful to template some variables like cluster name in some controller used annotation like externalDns


AndrewChubatiuk (approved):
LGTM. Thanks for contribution

PR #2107, opened 2025-04-03
Add erofs to ignored fs type

Nemric:
erofs should be considered as squashfs, a read only file system that should not be monitored


AndrewChubatiuk (approved):
LGTM. Thanks for contribution

Issue #2112, opened 2025-04-04
bug: victoria-metrics-operator-crds chart tries to create crds 2 times


slimes28:
**Chart name and version**
chart: victoria-metrics-operator-crds
version: v0.0.2

**Describe the bug**
victoria-metrics-operator-crds chart tries to create crds 2 times, here is an example of a warning that I see (have 16 of those)
![Image](https://github.com/user-attachments/assets/7fc65bab-78a0-405d-a4e4-eb8107cfe20e)
Before deploying this chart all victoria metrics operator crds were cleared manually

PR #2113, opened 2025-04-04
Fix doubled crd manifests vm-operator-crd

andrejshapal:
[2112](https://github.com/VictoriaMetrics/helm-charts/issues/2112)

Now crds are taken from 2 sources:
1. crd.yaml
2. from dependant chart
it creates conflict

This PR:
1. Fixes symlinks
2. Allows to disable dep chart and deploy from crd.yaml or use dep chart instead.


AndrewChubatiuk:
thank you for a fix
this issue was already addressed in [this commit](https://github.com/VictoriaMetrics/helm-charts/commit/96ced9099b75c2531a29520c4353c0cb63f35393)

Issue #2127, opened 2025-04-11
feature: add ability to run VictoriaLogs cluster


valyala:
**Chart name and version, where you feel a lack of requested feature**

The new chart must be created for [VictoriaLogs cluster](https://docs.victoriametrics.com/victorialogs/cluster/). The cluster mode is available starting from [v1.18.0-victorialogs](https://docs.victoriametrics.com/victorialogs/changelog/).

**Is your feature request related to a problem? Please describe.**

Cluster mode in VictoriaLogs isn't covered by helm charts.

**Describe the solution you'd like**

A helm chart for VictoriaLogs cluster, with the ability to configure the number of vlstorage, vlselect and vlinsert nodes.

**Describe alternatives you've considered**

Create a custom helm chart according to https://docs.victoriametrics.com/victorialogs/cluster/#architecture

valyala:
Please update https://docs.victoriametrics.com/victorialogs/cluster/#quick-start with the link to docs on how to configure and run the created helm chart for VictoriaLogs cluster when the helm chart will be published.

Haleygo:
We will create a new chart for VictoriaLogs cluster mode, with the following topology:

<img width="822" alt="Image" src="https://github.com/user-attachments/assets/6405a0fd-abe6-4990-acc6-20e3e2bf204d" />

The main features of this chart:
1. installs a single `vlcluster` by default(the replicas and resources of each component can be customized), providing write&read endpoints for data ingestion&querying.
2. provides an option to enable HA, which will:
   1. create two `vlcluster`;
   2. use vector as the global write endpoint to replicate data to both `vlcluster`;
   3. install vmauth-read-proxy as the global read endpint, which proxies query requests and switch from temporarily unavailable `vlcluster` to the other to provide query high availability.
3. allows to add AZ affinities to different `vlcluster`, so they can be scheduled to different AZs
4. provides a way for reading old data from an existing VictoriaLogs single instance, with the following topology:

<img width="864" alt="Image" src="https://github.com/user-attachments/assets/57fe96d9-2a9d-4e21-be61-5c9aca01b425" />

cc @valyala @hagen1778 @AndrewChubatiuk @tenmozes

f41gh7:
I just realized that using vector has one major disadvantage - it's not compatible with VictoriaLogs streaming fields setting.

 Vector doesn't have any knowledge of it, while it could an important feature for VLogs clients.

f41gh7:
[...]

PR #2130, opened 2025-04-12
victoria-logs-single: support cluster mode

AndrewChubatiuk:
added victorialogs cluster support to single chart

existing setups can be switched to cluster mode by setting
`.Values.servers.select.enabled` and `.Values.servers.insert.enabled` to `true`

Vector writes data either to `insert` instances if they are set or to `storage` if no `insert` instances found

Template rendering fails if neither no `storage` nodes found nor if no `extraArgs.storageNode` param is set. 

also it allows to use it for multilayered setup using each chart release as a separate layer (`-storageNode` flag lacks SRV discovery support for convenience and also prevents having HPA in a chart)

@zekker6 @f41gh7 @Haleygo @hagen1778 need your feedback

### DIFF
--- a/charts/victoria-logs-single/CHANGELOG.md
+++ b/charts/victoria-logs-single/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Next release
 
-- TODO
+- moved `.Values.podDisruptionBudget` to `.Values.servers.<groupName>.podDisruptionBudget` and `.Values.server.podDisruptionBudget`.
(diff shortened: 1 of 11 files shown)


makasim:
Fixes https://github.com/VictoriaMetrics/helm-charts/issues/2127

hagen1778:
After some internal discussion there was no consensus on whether we want to allow upgrades from single to cluster via the same helm chart. Potentially, it could add additional complexity and confusion to users.
We're now considering creating design for a separate helm chart for vlogs cluster with HA support. This work is to TBD by @Haleygo. Based on that work we will decide whether it should remain a separate chart or can be merged into existing chart.

PR #2132, opened 2025-04-14
distributed: support substituting VMAuth with k8s Service

AndrewChubatiuk:
may be useful for setups which have built-in load balancing, e.g: istio
In this setup k8s services are placed instead of global read and write load balancers, which target VMSelects and VMAgents respectively, cross zone balancer should be disabled


AndrewChubatiuk:
I haven't documented this use case yet, just prepared template updates

PR #2137, opened 2025-04-16
Allow overriding the load balancing policy for unauth access

lpetrazickisupgrade:
The motivation is that the hard-coded load balancing policy of `first_available` leaves half the vmselect instances unused. The change optionally overrides it when a different one is set in the spec, otherwise it continues to use the same default. It uses the same structure as in the [API ](https://docs.victoriametrics.com/operator/api/index.html#vmauthunauthorizeduseraccessspec)


AndrewChubatiuk:
hey @lpetrazickisupgrade 
thanks for PR. this was already implemented [here](https://github.com/VictoriaMetrics/helm-charts/commit/a0c61cc019815b8651352e82ced47986e7ffe78c)

Issue #2168, opened 2025-05-06
feature: Add plugin to VM datasource for Grafana operator


kobin23:
**Chart name and version, where you feel a lack of requested feature**
chart: victoria-metrics-k8s-stack-chart
version: v0.45.0

**Is your feature request related to a problem? Please describe.**
Yes. When deploying the VictoriaMetrics Kubernetes stack using the Helm chart, the `GrafanaDatasource` Custom Resource created for VictoriaMetrics does not include plugin metadata (such as marking `victoriametrics-metrics-datasource` as a required plugin)

**Describe the solution you'd like**
I’d like the Helm chart to optionally (or by default) add the appropriate `plugins` field to the generated `GrafanaDatasource` CR

**Describe alternatives you've considered**
- Manually editing the `GrafanaDatasource` after deployment.
- Creating a separate `GrafanaDatasource` manifest with the plugin requirement defined.
- Using a custom overlay or post-render patch in Helm to inject the plugin information.
All of these are more error-prone and reduce the value of an automated Helm deployment.

**Additional context**
Here’s an example of the missing field:
```
apiVersion: grafana.integreatly.org/v1beta1
kind: GrafanaDatasource
metadata:
  name: victoria-metrics-datasource
spec:
  datasources:
    - name: VictoriaMetrics (DS)
       type: victoriametrics-metrics-datasource
       access: proxy
       url: http://vmsingle:8428
       isDefault: true
  # MISSING:
  plugins:
    - victoriametrics-metrics-datasource
```

AndrewChubatiuk:
fixed in 0.45.1

kobin23:
Hello!
Still missing plugins section in GrafanaDatasource on latest version of chart (0.47.1)
May be I'm doing something wrong?

Part of my values.yaml
```
defaultDatasources:
  grafanaOperator:
    # -- Create datasources as CRDs (requires grafana-operator to be installed)
    enabled: true
    annotations: {}
    spec:
      instanceSelector:
        matchLabels:
          dashboards: grafana
      allowCrossNamespaceImport: false
  victoriametrics:
    # -- Create per replica prometheus compatible datasource
    perReplica: false
    # -- List of prometheus compatible datasource configurations.
    # VM `url` will be added to each of them in templates.
    datasources:
      - name: VictoriaMetrics
        type: prometheus
        access: proxy
        isDefault: true
      - name: VictoriaMetrics (DS)
        isDefault: false
        access: proxy
[...]

PR #2178, opened 2025-05-12
k8s-stack: fix plugins section in GrafanaDatasource

devsunb:
Fixed the following error:

> Error: UPGRADE FAILED: release vmks failed, and has been rolled back due to atomic being set: cannot patch "vmks-victoria-metrics-k8s-stack-victoriametrics-ds" with kind GrafanaDatasource: GrafanaDatasource.grafana.integreatly.org "vmks-victoria-metrics-k8s-stack-victoriametrics-ds" is invalid: [spec.plugins[0]: Invalid value: "string": spec.plugins[0] in body must be of type object: "string", <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation]

[Grafana Operator API Reference - GrafanaDashboard.spec.plugins[index]](https://grafana.github.io/grafana-operator/docs/api/#grafanadashboardspecpluginsindex)

Related Issue: #2168
Related Commit: https://github.com/VictoriaMetrics/helm-charts/commit/492341fcb66de9bef12ca75575283977e57821f5

### DIFF
--- a/charts/victoria-metrics-k8s-stack/CHANGELOG.md
+++ b/charts/victoria-metrics-k8s-stack/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next release
 
-- TODO
+- add plugin version while using GrafanaDatasource CRD
 
 ## 0.46.0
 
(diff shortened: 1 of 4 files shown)


AndrewChubatiuk:
Thanks for contribution
This PR was included into 0.47.1

PR #2201, opened 2025-05-28
victoria-logs-single: create sink per statefulset pod

AndrewChubatiuk:
elasticsearch sink sends data to a single endpoint among listed, which is not expected behaviour for HA single node HA setup, creating multiple sinks in a loop for each statefulset pod


f41gh7 (dismissed):
LGTM

PR #2217, opened 2025-06-03
k8s-stack: allow enabling `vmcluster` dashboard, even if vmcluster CR itself isn't enabled

rpardini:
The use-case here is, due to gitops repo/tree organization, we use k8s-stack only for dashboards/rules/scrapers, while most CRs are actually deployed separately.

### DIFF
--- a/charts/victoria-metrics-k8s-stack/files/dashboards/generated/victoriametrics-cluster.yaml
+++ b/charts/victoria-metrics-k8s-stack/files/dashboards/generated/victoriametrics-cluster.yaml
@@ -49,7 +49,7 @@ annotations:
     iconColor: dark-yellow
     name: restarts
     textFormat: '{{`{{`}}job{{`}}`}} restarted'
-condition: {{ ($Values.vmcluster).enabled }}
+condition: {{ or (($Values.vmcluster).enabled) (($Values.vmcluster).dashboardEnabled) }}
 description: Overview for cluster VictoriaMetrics v1.117.0 or higher
 editable: false
 fiscalYearStartMonth: 0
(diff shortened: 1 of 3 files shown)


AndrewChubatiuk:
hey @rpardini 
have you tried `defaultDashboards.dashboards.<dashboard-name>.enabled: true`
in your case setting 
```
defaultDashboards:
  dashboards:
    victoriametrics-cluster:
      enabled: true
```
should work

makasim:
At this point, there’s very little left of the original stack chart. Given the level of customization you're applying, I’d recommend cherry-picking only the specific dashboards, rules, and scrape configs you need.

You can do this by running:
```
helm template <release-name> > manifests.yaml
```

Then, extract the required manifests from the generated file.

AndrewChubatiuk:
it's already supported by chart as I've mentioned before

makasim:
> it's already supported by chart as I've mentioned before

@AndrewChubatiuk, as I understood it, only covers dashboards. What about rules and scrapers?

rpardini:
> have you tried `defaultDashboards.dashboards.<dashboard-name>.enabled: true`

That works -- at first glance it seemed to me the 'condition' could not be avoided. This solves it for me. 
Thanks!

> there’s very little left of the original stack chart

Agree, but what's left is of very high value; I do have a lot YAML-values disabling everything else which I'd love to avoid. 

Maybe in the future we could split off dashboards/rules/scrapes into a subchart (or three), so that k8s-stack ends up mostly empty. 

But I see how that is a lot of work for a very particular use case. 

Either way super thanks for the help.

makasim:
We're always happy to help! If there's anything else you need, just let us know.

PR #2245, opened 2025-06-14
VictoriaLogs: improve documents

func25:


charts/victoria-logs-cluster/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 appVersion: v1.24.0
-description: VictoriaLogs cluster version - high-performance, cost-effective and scalable logs storage
+description: The VictoriaLogs cluster chart packages everything required to run a horizontally scalable, highly available log-storage backend inside Kubernetes.
```
AndrewChubatiuk, line 4: hopefully it will not confuse people who expect setup to be highly available

charts/victoria-logs-cluster/_index.md.gotmpl
```diff
@@ -20,37 +20,113 @@ tags:
...
+    VLQ1 <--> Users["Users/Grafana/vlogscli"]
+    VLQ2 <--> Users
+```
+
+### Vector
+
+When Vector is enabled with `vector.enabled: true`, the default configuration works out of the box:
+
+* The default role is set to "Agent" (typically deployed as a DaemonSet with a data directory at `/vector-data-dir`).
+* Vector is configured with the `k8s` source, using the `kubernetes_logs` type to collect logs from all Kubernetes pods in the cluster.
```
zekker6, line 81: Could you add links to vector docs for these resources?
That would make it very easy for user to find an exact references when checking what this chart does and will help to speed up configuration troubleshooting.

zekker6 (commented):
Great work, see one small comment.

charts/victoria-logs-cluster/_index.md.gotmpl
```diff
@@ -20,37 +20,113 @@ tags:
 
 ## Prerequisites
 
-* Install the follow packages: ``git``, ``kubectl``, ``helm``, ``helm-docs``. See this [tutorial](https://docs.victoriametrics.com/helm/requirements/).
+Before installing this chart, ensure your environment meets the following requirements:
 
-* PV support on underlying infrastructure.
+* **Kubernetes cluster** - A running Kubernetes cluster with sufficient resources
```
AndrewChubatiuk, line 25: can we move prerequisites to a shared template to have it the same for all charts?
https://github.com/VictoriaMetrics/helm-charts/blob/22efa073d5e107ccc563db269096dab2c0f167b6/hack/docs/template.tmpl#L31

charts/victoria-logs-cluster/Chart.yaml
valyala: I think this is OK for the majority of users (even if they think they need HA, in reality they need just a cluster, since the most of the users think that cluster == HA). We'll add a link to the real HA cluster setup of VictoriaLogs similar to https://docs.victoriametrics.com/helm/victoriametrics-distributed/ when it will be ready, with a big warning explaining why the majority of users need just cluster setup (or even single-node setup) instead of the real HA cluster setup.

charts/victoria-logs-cluster/_index.md.gotmpl
valyala: I don't think this is a good idea in general, since every Helm chart may have different requirements. It is better to copy these requirements among Helm charts instead of trying to unify them in a single shared template.

charts/victoria-logs-cluster/_index.md.gotmpl
```diff
@@ -20,37 +20,113 @@ tags:
...
+* [Installing kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/)
 
 ## Chart Details
 
-This chart will do the following:
+### VictoriaLogs Cluster
 
-* Rollout [VictoriaLogs cluster](https://docs.victoriametrics.com/victorialogs/cluster/).
-* (optional) Rollout [vector](https://vector.dev/) to collect logs from pods.
+When you install the chart, it creates three core services and an optional log collector:
```
valyala, line 41: It is better to don't mention the number of components, since this number may change over time and this sentence may become inconsistent with the rest of the docs (this issue is repeated quite frequently in my practice). So it is better to replace the `three` with the `following`

```suggestion
When you install the chart, it creates the following core services and an optional log collector:
```

charts/victoria-logs-cluster/_index.md.gotmpl
```diff
@@ -20,37 +20,113 @@ tags:
...
+
+- `vlinsert` receives incoming log streams from various sources and load-balances them across remote nodes.
+- `vlstorage` forms a StatefulSet that stores raw data on persistent volumes, applies the configured retention policy, and replicates the data across the number of replicas you define.
```
valyala, line 44: The data isn't replicated at `vlstorage`. Also it isn't necessary mentioning about the retention policy here, since this is only a single configuration option out of many other configuration options available at `vlstorage` nodes.

```suggestion
- `vlstorage` forms a StatefulSet that stores raw data on persistent volumes.
```

valyala:
@func25 , please address my comments above in a separate pull request, since it looks like they weren't addressed in the merged pull request.

@AndrewChubatiuk , please do not merge pull requests next time until all the comments are addressed.

charts/victoria-logs-cluster/_index.md.gotmpl
func25: Most charts are the same, so I think it’s fine to use the same setup. For charts that need a different setup, we can simply avoid importing the install section.

charts/victoria-logs-cluster/_index.md.gotmpl
func25: Thanks for catching that. All the feedback has been addressed by Andrii.

PR #2246, opened 2025-06-15
charts/vmagent: Add trafficDistribution for service in vmagent chart

szibis:
Based on https://kubernetes.io/docs/concepts/services-networking/service/#traffic-distribution
Minimum Kubernetes needed is v1.31, but full functionality can be achieved in Kubernetes v1.33


AndrewChubatiuk:
LGTM, thanks for contribution

PR #2261, opened 2025-06-23
charts: run `make update` after 30e50c9

zekker6:


zekker6:
@valyala FYI: it is required to run `make update` if `dependencies` in `Chart.yaml` has changed as helm builds a digest based on dependencies hash.

Haleygo (approved):

PR #2286, opened 2025-07-04
k8s-stack: fail if storage nodes count is smaller than 2*replicationFactor - 1

### LINKED ISSUE #2285: replicaCount inconsistency in helm chart
### Describe the bug

The documentation on https://docs.victoriametrics.com/victoriametrics/cluster-victoriametrics/#replication-and-data-safety mentions
"The cluster must contain at least 2*N-1 vmstorage nodes, where N is replication factor, in order to maintain the given replication factor for newly ingested data when N-1 of storage nodes are unavailable."
k8s helm chart sets replication factor to "2" but leaves the vmstorage replicaCount to just "2"

### To Reproduce

Deploy cluster with helm chart setting replication factor 2, check that only 2 storage nodes are created

### Version

reproducible on victoria-metrics-cluster-0.25.0

### Logs

_No response_

### Screenshots

_No response_

### Used command-line flags

_No response_

### Additional information

_No response_

AndrewChubatiuk:
fixes #2285

### DIFF
--- a/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
+++ b/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
@@ -297,6 +297,12 @@
   {{- $Chart := (.helm).Chart | default .Chart }}
   {{- $selectSpec := include "vm.select.spec" . | fromYaml -}}
   {{- $clusterSpec := deepCopy $Values.vmcluster.spec -}}
+  {{- $replicationFactor := (int $clusterSpec.replicationFactor) | default 1 }}
+  {{- $storageNodes := (int ($clusterSpec.vmstorage).replicaCount) | default 1 }}
+  {{- $minStorageNodes := sub (mul 2 $replicationFactor) 1 }}
+  {{- if gt $minStorageNodes $storageNodes }}
+    {{ fail (printf "storage nodes count %d is smaller than expected %d" $storageNodes $minStorageNodes) }}
+  {{- end }}
   {{- $image := dict "image" (dict "tag" (printf "%s-cluster" (include "vm.image.tag" .))) }}
   {{- $clusterSpec = mergeOverwrite (dict "vminsert" (deepCopy $image)) $clusterSpec -}}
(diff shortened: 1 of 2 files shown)


charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
```diff
@@ -297,6 +297,12 @@
   {{- $Chart := (.helm).Chart | default .Chart }}
   {{- $selectSpec := include "vm.select.spec" . | fromYaml -}}
   {{- $clusterSpec := deepCopy $Values.vmcluster.spec -}}
+  {{- $replicationFactor := (int $clusterSpec.replicationFactor) | default 1 }}
+  {{- $storageNodes := (int ($clusterSpec.vmstorage).replicaCount) | default 1 }}
+  {{- $minStorageNodes := sub (mul 2 $replicationFactor) 1 }}
+  {{- if gt $minStorageNodes $storageNodes }}
+    {{ fail (printf "storage nodes count %d is smaller than expected %d" $storageNodes $minStorageNodes) }}
```
Haleygo, line 304: Not sure if chart should force this, it's the rule for maintain the given replication factor for newly ingested data when N-1 of storage nodes are unavailable. Maybe user don't want to follow that:)

But yeah, the change should be good.
Could you add a update note to the change log?

Haleygo (dismissed):

charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
AndrewChubatiuk: done

YuryHrytsuk:
@AndrewChubatiuk the default chart configuration still uses `replicationFactor: 2` and `vmstorage.replicaCount: 2` which does violate `2*N-1` rule. 

Why was https://github.com/VictoriaMetrics/helm-charts/issues/2285 closed? Am I missing something?

AndrewChubatiuk:
@YuryHrytsuk it was replaced by a warning
decided not to change chart defaults

PR #2290, opened 2025-07-07
charts/vm-anomaly: allow using VMPodScrape for monitoring configuration

zekker6:
This allows to avoid installing Prometheus operator CRDs if VictoriaMetrics operator is already present.

### DIFF
--- a/charts/victoria-metrics-anomaly/values.yaml
+++ b/charts/victoria-metrics-anomaly/values.yaml
@@ -145,6 +145,8 @@ podMonitor:
   extraLabels: {}
   # -- PodMonitor annotations
   annotations: {}
+  # -- Whether to use [VMPodScrape](https://docs.victoriametrics.com/operator/resources/vmpodscrape/) from VM operator instead of PodMonitor
+  vmPodScrape: false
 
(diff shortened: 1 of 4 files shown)


charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -145,6 +145,8 @@ podMonitor:
   extraLabels: {}
   # -- PodMonitor annotations
   annotations: {}
+  # -- Whether to use [VMPodScrape](https://docs.victoriametrics.com/operator/resources/vmpodscrape/) from VM operator instead of PodMonitor
+  vmPodScrape: false
```
AndrewChubatiuk: can we name it in the same manner like it's done in operator chart to keep in consistent?
https://github.com/VictoriaMetrics/helm-charts/blob/45a961bc8c1e01b7154af13062346fcbde524122/charts/victoria-metrics-operator/values.yaml#L306

zekker6: Updated in https://github.com/VictoriaMetrics/helm-charts/pull/2290/commits/b40e6d440088ac3bc1bb2b05ffca5add9f12f210

AndrewChubatiuk (approved):

PR #2294, opened 2025-07-07
remove hard variant

sfo-md:
We had issues with the deployment while not having set image tags via common option.

please be kind. this is only my second PR.


zekker6:
The fix was included in v0.0.7 release - https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-logs-cluster-0.0.7

Issue #2298, opened 2025-07-10
feature: provide victoria-logs-collector chart


vadimalekseev:
**Chart name and version, where you feel a lack of requested feature**
chart: victoria-logs-collector
version: v0.0.1

Provide a Helm chart for collecting logs in a Kubernetes cluster, which allows sending logs to VictoriaLogs. The chart should be deployed as a DaemonSet and parse incoming logs formatted as JSON. It is also necessary to enrich logs with pod metadata.

Additionally, the chart should abstract away the internal implementation details so that the underlying components can be replaced in the future with vlagent without impacting the users

valyala:
The `victoria-logs-collector` Helm chart has been added in the commit 9eb3620148a56bd78e796ae0ce44e249490390e0 . See https://github.com/VictoriaMetrics/helm-charts/tree/master/charts/victoria-logs-collector . Closing the issue as fixed.

@vadimalekseev , please add the `victoria-logs-collector` chart into `victoria-logs-single` and `victoria-logs-cluster` as described [here](https://github.com/VictoriaMetrics/helm-charts/pull/2300#issuecomment-3209273039).

PR #2300, opened 2025-07-11
charts/victoria-logs-collector: implement victoria-logs-collector chart

vadimalekseev:
Related #2298

The victoria-logs-collector chart is built on top of vector.dev and provides a minimal configuration, allowing users to start collecting logs into a VictoriaLogs cluster with minimal setup and no need for deep customization.

In the future, vector.dev is planned to be replaced with vlagent, so the chart should be designed to support this migration without breaking backward compatibility.

### DIFF
--- a/charts/victoria-logs-collector/Chart.yaml
+++ b/charts/victoria-logs-collector/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 type: application
 name: victoria-logs-collector
-description: VictoriaLogs Collector - collects logs from nodes and stores them to VictoriaLogs
+description: VictoriaLogs Collector - collects logs from Kubernetes containers and stores them to VictoriaLogs
 version: 0.0.1
 sources:
   - https://github.com/VictoriaMetrics/helm-charts
--- a/charts/victoria-logs-collector/_index.md.gotmpl
+++ b/charts/victoria-logs-collector/_index.md.gotmpl
@@ -24,7 +24,10 @@ tags:
 ## Chart Details
 
 This Helm chart deploys a log collection agent as a DaemonSet.
-It collects logs from all nodes in a Kubernetes cluster and forwards them to one or more VictoriaLogs instances.
+It collects logs from all containers in a Kubernetes cluster and forwards them to the configured VictoriaLogs destinations.
+If more than one destination is specified, then the collected logs are replicated among the configured destinations.
+
+This chart will expand its functionality as the corresponding features are added to [vlagent](https://docs.victoriametrics.com/victorialogs/vlagent/index.html).
 
 The simplest working configuration includes specifying the `remoteWrite` array and setting CPU and memory resources for the chart.
 
@@ -114,6 +117,44 @@ remoteWrite:
     tls:
       insecureSkipVerify: true
 ```
+
+### Ignore fields
+
+VictoriaLogs efficiently compresses repeated values, such as pod and node labels.
+However, if you prefer not to store certain fields, you can ignore them using the `ignoreFields` option.
+For example:
+
+```yaml
+remoteWrite:
+  - url: http://victoria-logs:9428
+    ignoreFields:
+      - file
+      - kubernetes.container_id
+      - kubernetes.pod_annotations*
+      - kubernetes.node_labels*
+      - kubernetes.namespace_labels*
+      - kubernetes.pod_labels*
+```
+
+This allows you to exclude unnecessary or sensitive fields from being ingested.
+If sensitive data has already been ingested, see how to
+[exclude logs from search result](https://docs.victoriametrics.com/victorialogs/security-and-lb/#access-control-inside-a-single-tenant).
+
+### Extra fields
+
+You can add custom fields to your logs by including the `extraFields` section in your configuration.
+For example:
+
+```yaml
+remoteWrite:
+  - url: http://victoria-logs:9428/
+    extraFields:
+      zone: us-east1-c
+      source: victoria-logs-collector
+```
+
+This feature lets you attach metadata to every log entry,
+making it easier to filter, group, or analyze logs based on these additional attributes.
 
 ## Parameters
 
--- a/charts/victoria-logs-collector/templates/configmap.yaml
+++ b/charts/victoria-logs-collector/templates/configmap.yaml
@@ -33,7 +33,7 @@ data:
         {{ fail (printf "'url' field must be set for remoteWrite[%d]" $i) }}
       {{- end }}
       {{- range $rwKey, $_ := $rw }}
-        {{- if and (ne $rwKey "url") (ne $rwKey "basicAuth") (ne $rwKey "accountID") (ne $rwKey "projectID") (ne $rwKey "tls") }}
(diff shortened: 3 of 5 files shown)


AndrewChubatiuk (dismissed):

AndrewChubatiuk:
@zekker6 @f41gh7 @Haleygo 
what do you think about start naming charts with a shorter prefix to make it harder to reach 63 symbols limit:
`vl-` instead of `victoria-logs-`
`vm-` instead of `victoria-metrics-`

Haleygo:
> @zekker6 @f41gh7 @Haleygo what do you think about start naming charts with a shorter prefix to make it harder to reach 63 symbols limit: `vl-` instead of `victoria-logs-` `vm-` instead of `victoria-metrics-`

What about setting `vl` as the default value of `.Values.nameOverride`, while keeping the full chart name?

zekker6:
> What about setting vl as the default value of .Values.nameOverride, while keeping the full chart name?

Additionally, we can make this a default value for vm charts as well but that needs to be explicitly documented with a way to keep previous behavior.
That should help to reduce amount of issues which users are facing with greenfield installation while keeping existing installations untouchted.

AndrewChubatiuk:
adding it as a default value to old charts will break existing installations as many users are reading update notes only if upgrade failed

zekker6:
Wdyt about bumping major version to highlight a breaking change?

AndrewChubatiuk:
i don't like idea with default fullnameOverride as it may lead to collision if multiple instances of chart are deployed. it replaces `<chart-name>-<release-name>` with a static value from fullnameOverride

AndrewChubatiuk:
but nameOverride as Haley mentioned may help. will check this.
by the way could you both give any reasons why are you against naming new charts in a different way?

AndrewChubatiuk:
there's also another issue i would like to start doing in new charts - change naming convention to the same as we have in operator: `<resource-kind>-<resource-name>` instead of `<resource-name>-<resource-kind>`

Haleygo:
>by the way could you both give any reasons why are you against naming new charts in a different way?

Because short name like `vm` is not very specific, I prefer to keep the clarity.

AndrewChubatiuk:
we already don't have this clarity as we have `vtselect`, `vminsert` and `vlstorage`, we already use it everywhere

charts/victoria-logs-collector/templates/configmap.yaml
```diff
@@ -0,0 +1,87 @@
...
+        inputs: [json_parser]
+        uri: {{ printf "%s/insert/jsonline" (trimSuffix "/" $rw.url) | quote }}
+        encoding:
+          codec: json
+        framing:
+          method: newline_delimited
+        healthcheck:
+          enabled: false
+        request:
+          headers:
```
vadimalekseev, line 51: I am not sure we need to expose the `headers` section, because it complicates setting default headers like `VL-Stream-Fields` and so on. What do you think?

charts/victoria-logs-collector/Chart.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v2
+type: application
+name: victoria-logs-collector
+description: VictoriaLogs Collector - collects logs from nodes and stores them to VictoriaLogs
```
valyala:
```suggestion
description: VictoriaLogs Collector - collects logs from Kubernetes containers and stores them to VictoriaLogs
```

charts/victoria-logs-collector/_index.md.gotmpl
```diff
@@ -0,0 +1,124 @@
...
+{{ template "chart.description" . }}
+
+{{ template "chart.prerequisites" . }}
+
+{{ include "chart.installSection" . }}
+
+## Chart Details
+
+This Helm chart deploys a log collection agent as a DaemonSet.
+It collects logs from all nodes in a Kubernetes cluster and forwards them to one or more VictoriaLogs instances.
```
valyala:
```suggestion
It collects logs from all containers in a Kubernetes cluster and forwards them to the configured VictoriaLogs destinations. If more than one destination is specified, then the collected logs are replicated among the configured destinations.
```

charts/victoria-logs-collector/templates/configmap.yaml
f41gh7: I think, current configuration is fine and there is no need to expose headers  at .Values.

charts/victoria-logs-collector/_index.md.gotmpl
f41gh7: We should mention why it's not possible to override or change collector configuration.

charts/victoria-logs-collector/values.yaml
```diff
@@ -0,0 +1,129 @@
...
+# Advanced helm-chart configuration below, it is recommended to pass default parameters
+
+# -- Override chart name
+nameOverride: ""
+
+# -- Override resources fullname
+fullnameOverride: ""
+
+global:
+  # -- Image pull secrets, that can be shared across multiple helm charts
```
f41gh7, line 60: For me, it'd be great to provide some global "labels" for all collected logs. It helps to distinguish logs from multiple clusters.

 As an example, there is 2 Kubernetes cluster with the same application sets. And helm charts forwards collected logs into central cluster. Due to possible podname collision, it may be hard to query exact cluster logs. Or perform logs grouping over clusters.

Utwo:
Isn't this already available here https://docs.victoriametrics.com/helm/victorialogs-single/#sending-logs-to-external-victorialogs?

charts/victoria-logs-collector/values.yaml
vadimalekseev: Added an `extraFields` section for each `remoteWrite`. See https://github.com/VictoriaMetrics/helm-charts/pull/2300/commits/f454de1d1530ca8ee8b235bca0e35e0ba867fdaf

This setting is not global, but it solves the problem described, what do you think?

charts/victoria-logs-collector/_index.md.gotmpl
vadimalekseev: I don't know how to do this without using the word 'vector', which we agreed not to mention. 

Added information that the chart's functionality will be expanded as soon as the corresponding functionality becomes available in vlagent.

f41gh7 (approved):
LGTM

vadimalekseev:
> Isn't this already available here https://docs.victoriametrics.com/helm/victorialogs-single/#sending-logs-to-external-victorialogs?

@Utwo, please see https://github.com/VictoriaMetrics/VictoriaLogs/issues/68#issuecomment-3158214868 for details.

valyala:
@vadimalekseev , thank you for adding victoria-logs-collector Helm chart to the official list of helm charts provided by VictoriaMetrics!

The next step is to replace `vector` with the `victoria-logs-collector` in [`victoria-logs-single`](https://github.com/VictoriaMetrics/helm-charts/tree/master/charts/victoria-logs-single) and [`victoria-logs-cluster`](https://github.com/VictoriaMetrics/helm-charts/tree/master/charts/victoria-logs-cluster) charts. The `vector` collector shouldn't be removed from these charts in order allow users upgrading to new charts without breaking backwards compatibility. The `vector` collector must be removed from all the documentation for these charts. Instead, it must be replaced with the docs on how to enable `victoria-logs-collector` (it must be disabled by default). The docs must be as simple as possible, such as "put `victoria-logs-collector: true`" in the chart config.

PR #2340, opened 2025-08-13
Add templating for node-exporter job label in rules

Nenq:
### Description
This PR introduces the ability to configure the `job` label used in the `node-exporter` rules  files. It replaces the previously hardcoded label with a templated variable.

### DIFF
--- a/hack/rules-and-dashboards/sync_rules.py
+++ b/hack/rules-and-dashboards/sync_rules.py
@@ -188,6 +188,15 @@ replacement_map = {
         "replacement": 'namespace="[[ include "vm.namespace" . ]]"',
         "limitGroup": ["alertmanager.rules"],
     },
+    'job="node-exporter"': {
+        "replacement": 'job="[[ $jobLabel ]]"',
+        "limitGroup": [
+            "node-exporter.rules",
+            "node-exporter",
+            "node-network",
+            "node.rules",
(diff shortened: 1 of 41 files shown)


hack/rules-and-dashboards/sync_rules.py
```diff
@@ -281,6 +290,7 @@ def write_group_to_file(group, url, charts):
             content += "{{- $additionalGroupByLabels := append $Values.defaultRules.additionalGroupByLabels $clusterLabel }}\n"
             content += '{{- $groupLabels := join "," $additionalGroupByLabels }}\n'
             content += "{{- $grafanaHost := ternary (index (($Values.grafana).ingress).hosts 0) (($Values.external).grafana).host ($Values.grafana).enabled }}\n"
+            content += '{{- $jobLabel := (index $Values "prometheus-node-exporter").service.labels.jobLabel | default "node-exporter" }}\n'
```
AndrewChubatiuk: this should not be the same for all rules since not all of them are for node-exporter

Nenq: adjusted the code to add the variable only to node-exporter rule files by adding a list with static file names.

AndrewChubatiuk:
LGTM! Thanks for contribution

PR #2354, opened 2025-08-18
set default nameOverride to avoid hitting resource name limit for new setups

AndrewChubatiuk:
updated default nameOverride for all charts to avoid hitting 63 chars limit in a future
makes sense to bump a major version during the next release

### DIFF
--- a/charts/victoria-logs-cluster/CHANGELOG.md
+++ b/charts/victoria-logs-cluster/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Next release
 
-- TODO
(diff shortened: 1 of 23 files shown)


Haleygo:
I'm okay with changing the default nameOverride for VLogs charts since they're relatively new and won't affect much of existing setups.
But I'd like to avoid changing the default name for VM charts. Even with upgrade notes, users may ignore them, and the upgrade could trigger many restarts and cause connection issues(e.g. some service is hard-coded to connect to VM chart-deployed resources). 
Maybe we can add notes to the VM charts documentation, advising users who fail installation due to the name length limit to set the nameOverride?

AndrewChubatiuk:
[added note](https://github.com/VictoriaMetrics/helm-charts/commit/e3adc77ea004502cc429e57cc759ed511a4c015e) to docs instead

Issue #2367, opened 2025-08-20
feature: vmalert: only use server.notifier.alertmanager.url if it is not empty


schustersv:
**Chart name and version, where you feel a lack of requested feature**
chart: victoria-metrics-cluster
version: v0.25.0

**Is your feature request related to a problem? Please describe.**
With v0.25.0 of victoria-metrics-alert Helm chart there was a change to how the notifier url list is generated with commit b9ba026: now, the value of server.notifier.alertmanager.url is always prepended to the list. While this has existed in a similar fashion before, when only using server.notifiers.alertmanager.url (mind the plural here) for consistency reasons this now leads to an superfluous colon at the beginning of the alertmanager url parameter to vmalert:
`--notifier.url=,http://alertmanager1:9093,http://alertmanager2:9093,http://alertmanager3:9093`

**Describe the solution you'd like**
Only use server.notifier.alertmanager.url if it is not empty

**Describe alternatives you've considered**
When using multiple alertmanagers, another approach is to set server.notifier.alertmanager.url to the first alertmanager url and server.notifiers.alertmanager.url for the remaining ones. For such setups, this neither intuitive nor consistent thoug imo.

**Additional context**
I'll provide a pr soon for the suggested change.

AndrewChubatiuk:
released in 0.25.1

PR #2368, opened 2025-08-20
fix: don't print server.notifier.alertmanager.url if empty

schustersv:
implements #2367

### DIFF
--- a/charts/victoria-metrics-alert/templates/_helpers.tpl
+++ b/charts/victoria-metrics-alert/templates/_helpers.tpl
@@ -30,7 +30,9 @@ Create base alertmanager url for notifers
       {{- $urls = append $urls (include "vm.url" $ctx) -}}
     {{- end }}
   {{- else -}}
-    {{- $urls = append $urls ($Values.server.notifier.alertmanager.url) -}}
+    {{- with $Values.server.notifier.alertmanager.url -}}
+      {{- $urls = append $urls . -}}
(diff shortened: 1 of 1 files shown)


charts/victoria-metrics-alert/templates/_helpers.tpl
```diff
@@ -30,7 +30,9 @@ Create base alertmanager url for notifers
       {{- $urls = append $urls (include "vm.url" $ctx) -}}
     {{- end }}
   {{- else -}}
-    {{- $urls = append $urls ($Values.server.notifier.alertmanager.url) -}}
+    {{- if not (empty ($Values.server.notifier.alertmanager.url)) -}}
+      {{- $urls = append $urls ($Values.server.notifier.alertmanager.url) -}}
```
AndrewChubatiuk:
```suggestion
    {{- with $Values.server.notifier.alertmanager.url -}}
      {{- $urls = append $urls . -}}
```

AndrewChubatiuk (dismissed):
LGTM, please check a minor suggestion

charts/victoria-metrics-alert/templates/_helpers.tpl
schustersv: good catch, thanks for the suggestion - changed

AndrewChubatiuk (approved):
Thanks for contribution

PR #2384, opened 2025-09-01
Host aliases

tbos-h:
We use vmagent to send metrics from cloud to local infrastructure. Unfoturnatly, we do not have option to configure DNS server in cloud infra. 
Adding hostAliases allows more flexible use of vmagent without requiring a DNS server.

### DIFF
--- a/charts/victoria-metrics-agent/Chart.yaml
+++ b/charts/victoria-metrics-agent/Chart.yaml
@@ -3,7 +3,7 @@ type: application
 name: victoria-metrics-agent
 description: VictoriaMetrics Agent - collects metrics from various sources and stores them to VictoriaMetrics
 version: 0.25.3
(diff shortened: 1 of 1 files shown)


charts/victoria-metrics-agent/Chart.yaml
```diff
@@ -3,7 +3,7 @@ type: application
 name: victoria-metrics-agent
 description: VictoriaMetrics Agent - collects metrics from various sources and stores them to VictoriaMetrics
 version: 0.25.3
-appVersion: v1.125.0
+appVersion: v1.125.1
```
AndrewChubatiuk:
```suggestion
appVersion: v1.125.0
```

AndrewChubatiuk:

Issue #2388, opened 2025-09-03
bug: victoria-logs-cluster vlstorage.existingClaim doesn't work


sahilarora535:
**Chart name and version**
chart: victoria-logs-cluster
version: v0.10.0

**Describe the bug**
I am using `victoria-logs-cluster` with two `vlstorage` pods. There is a key in `vlstorage.persistentVolume.existingClaim`. When this key is set, the expectation is to use the supplied persistent volume claim instead of creating a new claim. When this is unset, the `helm-chart` creates 2 `pvc`, `<pvc-name>-0` and `<pvc-name>-1`.

What I did was I created the persistent volume claim myself with the same name. When I set `vlstorage.persistentVolume.existingClaim: "victoria-logs-cluster-vlstorage-pvc"` then the `vlstorage` pods don't come up.

```
Events:                                                                                                                                                                                                          │
│   Type     Reason             Age                 From                Message                                                                                                                                    │
│   ----     ------             ----                ----                -------                                                                                                                                    │
│   Warning  FailedScheduling   35m                 default-scheduler   0/12 nodes are available: persistentvolumeclaim "victoria-logs-cluster-vlstorage-pvc" not found. preemption: 0/12 nodes are available: 12  │
│ Preemption is not helpful for scheduling.                                                                                                                                                                        │
│   Warning  FailedScheduling   20m (x3 over 30m)   default-scheduler   0/12 nodes are available: persistentvolumeclaim "victoria-logs-cluster-vlstorage-pvc" not found. preemption: 0/12 nodes are available: 12  │
│ Preemption is not helpful for scheduling.                                                                                        
```

This is happening because of the fact that I created two PVCs, which were:

```
victoria-logs-cluster-vlstorage-pvc-0
victoria-logs-cluster-vlstorage-pvc-1
```

And instead of following the `statefulset` indexing for the existing claim, it's using the claim name as is, which it's unable to find.

This looks like a gap in `helm` chart. It would be greatly helpful if this could be fixed.

**Custom values**
Please provide only custom values (excluding default ones):
```yaml
  persistentVolume:
    # -- Create/use Persistent Volume Claim for vlstorage component. Empty dir if false. If true,  vlstorage will create/use a Persistent Volume Claim
    enabled: true
    name: vlstorage-volume

    # -- Array of access modes. Must match those of existing PV or dynamic provisioner. Details are [here](https://kubernetes.io/docs/concepts/storage/persistent-volumes/)
    accessModes:
[...]

PR #2389, opened 2025-09-03
charts/vlogs-cluster/vlstorage: existingClaim support for StatefulSet PVC naming

func25:
Related: https://github.com/VictoriaMetrics/helm-charts/issues/2388

### DIFF
--- a/charts/victoria-logs-single/templates/server.yaml
+++ b/charts/victoria-logs-single/templates/server.yaml
@@ -160,21 +160,21 @@ spec:
         {{- with $app.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or (eq $mode "deployment") (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
         - name: server-volume
-          {{- if or (and (eq $mode "deployment") $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
           persistentVolumeClaim:
             claimName: {{ $app.persistentVolume.existingClaim | default $app.persistentVolume.name | default $fullname }}
           {{- else }}
           emptyDir: {{ toYaml $app.emptyDir | nindent 12 }}
           {{- end }}
         {{- end }}
-  {{- if and (eq $mode "statefulSet") $app.persistentVolume.enabled (not $app.persistentVolume.existingClaim) }}
+  {{- if and (eq $mode "statefulSet") $app.persistentVolume.enabled }}
   volumeClaimTemplates:
     - apiVersion: v1
       kind: PersistentVolumeClaim
       metadata:
-        name: server-volume
+        name: {{ $app.persistentVolume.existingClaim | default "server-volume" }}
         {{- with $app.persistentVolume.annotations }}
         annotations: {{ toYaml . | nindent 10 }}
         {{- end }}
--- a/charts/victoria-logs-single/templates/server.yaml
+++ b/charts/victoria-logs-single/templates/server.yaml
@@ -160,9 +160,9 @@ spec:
         {{- with $app.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
+        {{- if or (not $app.persistentVolume.enabled) (eq $mode "deployment") }}
         - name: server-volume
-          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled }}
           persistentVolumeClaim:
             claimName: {{ $app.persistentVolume.existingClaim | default $app.persistentVolume.name | default $fullname }}
           {{- else }}
--- a/charts/victoria-metrics-agent/templates/server.yaml
+++ b/charts/victoria-metrics-agent/templates/server.yaml
@@ -171,9 +171,9 @@ spec:
         {{- end }}
       {{- end }}
       volumes:
-        {{- if or (eq $mode "deployment") (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
         - name: tmpdata
-          {{- if or (and (eq $mode "deployment") $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
           persistentVolumeClaim:
             claimName: {{ $app.persistentVolume.existingClaim | default $app.persistentVolume.name | default $fullname }}
           {{- else }}
@@ -197,12 +197,12 @@ spec:
         {{- toYaml . | nindent 8 }}
         {{- end }}
         {{- include "vm.license.volume" . | nindent 8 }}
-  {{- if and (eq $mode "statefulSet") $app.persistentVolume.enabled (not $app.persistentVolume.existingClaim) }}
+  {{- if and (eq $mode "statefulSet") $app.persistentVolume.enabled }}
   volumeClaimTemplates:
   - apiVersion: v1
     kind: PersistentVolumeClaim
     metadata:
-      name: tmpdata
+      name: {{ $app.persistentVolume.existingClaim | default "tmpdata" }}
       {{- with $app.persistentVolume.extraLabels }}
       labels: {{ toYaml . | nindent 8 }}
       {{- end }}
--- a/charts/victoria-metrics-agent/templates/server.yaml
+++ b/charts/victoria-metrics-agent/templates/server.yaml
@@ -171,9 +171,9 @@ spec:
         {{- end }}
       {{- end }}
       volumes:
-        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
+        {{- if or (not $app.persistentVolume.enabled) (eq $mode "deployment") }}
         - name: tmpdata
-          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled }}
           persistentVolumeClaim:
             claimName: {{ $app.persistentVolume.existingClaim | default $app.persistentVolume.name | default $fullname }}
           {{- else }}
--- a/charts/victoria-metrics-alert/templates/alertmanager-server.yaml
+++ b/charts/victoria-metrics-alert/templates/alertmanager-server.yaml
@@ -143,9 +143,9 @@ spec:
         {{- with $app.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
+        {{- if or (not $app.persistentVolume.enabled) (eq $mode "deployment") }}
         - name: server-volume
-          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled }}
           persistentVolumeClaim:
             claimName: {{  $app.persistentVolume.existingClaim | default $app.persistentVolume.name | default $fullname }}
           {{- else }}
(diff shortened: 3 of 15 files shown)


AndrewChubatiuk:
@func25 
thanks for PR, could you please also propagate this change to other charts as well?

func25:
Hey @AndrewChubatiuk, please help review again.

charts/victoria-metrics-single/templates/server.yaml
```diff
@@ -279,22 +279,22 @@ spec:
         {{- with $app.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or (eq $mode "deployment") (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
         - name: server-volume
-          {{- if or (and (eq $mode "deployment") $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
```
AndrewChubatiuk:
```suggestion
          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled }}
```

charts/victoria-metrics-cluster/templates/vmselect-server.yaml
```diff
@@ -142,22 +142,22 @@ spec:
         {{- with $app.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or (eq $mode "deployment") (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
         - name: cache-volume
-          {{- if or (and (eq $mode "deployment") $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
```
AndrewChubatiuk:
```suggestion
          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled }}
```

charts/victoria-logs-single/templates/server.yaml
```diff
@@ -160,21 +160,21 @@ spec:
         {{- with $app.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or (eq $mode "deployment") (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
```
AndrewChubatiuk:
```suggestion
        {{- if or (not $app.persistentVolume.enabled) (eq $mode "deployment") }}
```

charts/victoria-logs-single/templates/server.yaml
```diff
@@ -160,21 +160,21 @@ spec:
...
         - name: server-volume
-          {{- if or (and (eq $mode "deployment") $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
```
AndrewChubatiuk:
```suggestion
          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled }}
```

charts/victoria-metrics-agent/templates/server.yaml
```diff
@@ -171,9 +171,9 @@ spec:
         {{- end }}
       {{- end }}
       volumes:
-        {{- if or (eq $mode "deployment") (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
```
AndrewChubatiuk:
```suggestion
        {{- if or (not $app.persistentVolume.enabled) (eq $mode "deployment") }}
```

charts/victoria-metrics-agent/templates/server.yaml
```diff
@@ -171,9 +171,9 @@ spec:
...
         - name: tmpdata
-          {{- if or (and (eq $mode "deployment") $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
```
AndrewChubatiuk:
```suggestion
          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled }}
```

charts/victoria-metrics-alert/templates/alertmanager-server.yaml
```diff
@@ -143,21 +143,21 @@ spec:
         {{- with $app.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or (eq $mode "deployment") (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
```
AndrewChubatiuk:
```suggestion
        {{- if or (not $app.persistentVolume.enabled) (eq $mode "deployment") }}
```

charts/victoria-metrics-alert/templates/alertmanager-server.yaml
```diff
@@ -143,21 +143,21 @@ spec:
...
         - name: server-volume
-          {{- if or (and (eq $mode "deployment") $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled $app.persistentVolume.existingClaim }}
```
AndrewChubatiuk:
```suggestion
          {{- if and (eq $mode "deployment") $app.persistentVolume.enabled }}
```

charts/victoria-metrics-cluster/templates/vmselect-server.yaml
```diff
@@ -142,22 +142,22 @@ spec:
         {{- with $app.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or (eq $mode "deployment") (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
```
AndrewChubatiuk:
```suggestion
        {{- if or (not $app.persistentVolume.enabled) (eq $mode "deployment") }}
```

charts/victoria-metrics-single/templates/server.yaml
```diff
@@ -279,22 +279,22 @@ spec:
         {{- with $app.extraVolumes }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-        {{- if or (eq $mode "deployment") (not $app.persistentVolume.enabled) $app.persistentVolume.existingClaim }}
+        {{- if or (not $app.persistentVolume.enabled) (and (eq $mode "deployment") $app.persistentVolume.existingClaim) }}
```
AndrewChubatiuk:
```suggestion
        {{- if or (not $app.persistentVolume.enabled) (eq $mode "deployment") }}
```

charts/victoria-metrics-anomaly/templates/server.yaml
```diff
@@ -120,9 +120,9 @@ spec:
...
             name: {{ $fullname }}
         {{- include "vm.license.volume" . | nindent 8 }}
-        {{- if or (not .Values.persistentVolume.enabled) .Values.persistentVolume.existingClaim }}
+        {{- if or (and (not $useSts) (not .Values.persistentVolume.enabled)) (and (not $useSts) .Values.persistentVolume.existingClaim) }}
         - name: models-dump
-          {{- if .Values.persistentVolume.enabled }}
+          {{- if and (not $useSts) .Values.persistentVolume.enabled .Values.persistentVolume.existingClaim }}
           persistentVolumeClaim:
             claimName: {{ .Values.persistentVolume.existingClaim }}
           {{- else }}
```
AndrewChubatiuk:
```suggestion
```

charts/victoria-metrics-anomaly/templates/server.yaml
```diff
@@ -120,9 +120,9 @@ spec:
           configMap:
             name: {{ $fullname }}
         {{- include "vm.license.volume" . | nindent 8 }}
-        {{- if or (not .Values.persistentVolume.enabled) .Values.persistentVolume.existingClaim }}
+        {{- if or (and (not $useSts) (not .Values.persistentVolume.enabled)) (and (not $useSts) .Values.persistentVolume.existingClaim) }}
```
AndrewChubatiuk:
```suggestion
        {{- if not .Values.persistentVolume.enabled }}
```

charts/victoria-metrics-anomaly/templates/server.yaml
AndrewChubatiuk: also need to remove `{{ end }}` after `emptyDir`

charts/victoria-metrics-anomaly/templates/server.yaml
AndrewChubatiuk: useSts is true, when persistentVolume.enabled is true, it means, that `persistentVolumeClaim` section is not needed here

AndrewChubatiuk (commented):
left some suggestions, most of them are about the same condition, which requires existingClaim to be set to use persistent volume for deployment

func25:
Applied requested changes across charts:
- Simplified conditions per suggestions: volumes now use `or (not enabled) (eq mode "deployment")` and PVC mount blocks use `and (eq mode "deployment") enabled`.
- Updated vmanomaly models-dump logic to avoid PVC section when using StatefulSet and only render PVC for Deployment with existingClaim or emptyDir otherwise.
- Linted all modified charts locally.

Please take another look.

sahilarora535:
Hi @AndrewChubatiuk @func25, thanks for the great work on this. Looks like this PR is complete. When do you plan to merge and release this bugfix?

AndrewChubatiuk:
hi @sahilarora535 
have you tried to use `name` property instead of `existingClaim`?

sahilarora535:
@AndrewChubatiuk I am trying to use both `name` and `existingClaim`. Initially, I was using only `name: vlstorage-volume` with `existingClaim: ""`, and this created new `pvc`. Then I created two new `pvc`: `my-existing-persitent-volume-claim-1` and `my-existing-persitent-volume-claim-2`, and used the following configuration:

```yaml
vlstorage:
  persistentVolume:
    # -- Create/use Persistent Volume Claim for vlstorage component. Empty dir if false. If true,  vlstorage will create/use a Persistent Volume Claim
    enabled: true
    name: vlstorage-volume

    # -- Array of access modes. Must match those of existing PV or dynamic provisioner. Details are [here](https://kubernetes.io/docs/concepts/storage/persistent-volumes/)
    accessModes:
      - ReadWriteOnce
    # -- Persistent volume annotations
    annotations: {}
    # -- Persistent volume labels
    labels: {}
    # -- Storage class name. Will be empty if not set
    storageClassName: ""
    # --  Existing Claim name. Requires vlstorage.persistentVolume.enabled: true. If defined, PVC must be created manually before volume will be bound
    existingClaim: my-existing-persitent-volume-claim

    # -- Data root path. vlstorage data Persistent Volume mount root path
    mountPath: /storage
    # -- Mount subpath
    subPath: ""
```

Do you suggest the following:

```yaml
vlstorage:
  persistentVolume:
    name: my-existing-persitent-volume-claim
    existingClaim: ""
```
If so, should I keep the `size` config or remove it? What's the correct config?

AndrewChubatiuk:
you need to set a size equal to a size of existing claim

sahilarora535:
@AndrewChubatiuk thats the problem. So currently, we have multiple environments and we want different disk size for each environment, hence the pvc was made before for each environment. The idea was to attach the already existing pvc with configured size to the chart. Due to technical limitations, it’s not possible to have different helm chart values for each environment. If I mention the size in the helm chart, it would pick the same size for all envs. How do I workaround this?

AndrewChubatiuk:
there's no workaround for it, you need to pass custom values for each chart. it's possible to do with a separate values file or values defined using `--set` cmd argument

sahilarora535:
That might be a problem. Why can’t we use the existing PVC claim? The claim would already have the size configuration. That’s what we do for other tools as well, where we need different sizes for dev/staging/production environments. 

Is it possible to pass it via config map or something? I was hoping this PR would fix it and get us moving. If it’s merged, wouldn’t it automatically pick up the existing claim instead of creating it’s own?  That way, we wouldn’t have to pass the size. There is no way for us to set env specific variables.

AndrewChubatiuk:
There's no workaround and also this is not a chart limitation since resources.requests.storage is required by volumeclaimtemplate.spec. statefulset spec requires storage to be explicitly defined

AndrewChubatiuk:
resources.requests.storage is required property for volumeClaimTemplate spec, there's nothing can be done on charts side, since it's a k8s limitation. regarding setup you're having eventually there will be a difference in values you're passing to different environments, since they may at least differ in load, so requests for pods will be different anyway. I would recommend to provide an ability to pass additional values for each environment setup instead trying to find a workaround for this case

Regarding this PR it's unlikely to be merged since in the end it makes `persistentVolume.existingClaim` property an alias to existing `persistentVolume.name` property, which makes no sense

PR #2399, opened 2025-09-10
add victoria-traces-single chart

AndrewChubatiuk:
added victoria traces single chart


charts/victoria-traces-single/values.yaml
```diff
@@ -0,0 +1,367 @@
...
+  # -- Openshift security context compatibility configuration
+  compatibility:
+    openshift:
+      adaptSecurityContext: "auto"
+  cluster:
+    # -- K8s cluster domain suffix, uses for building storage pods' FQDN. Details are [here](https://kubernetes.io/docs/tasks/administer-cluster/dns-custom-nameservers/)
+    dnsDomain: cluster.local.
+
+# -- Override chart name
+nameOverride: ""
```
Haleygo: What about having the default `nameOverride: "vt"` for the new charts as we discussed before to avoid overlong  resource name?

AndrewChubatiuk: set it it `vt-single` and in PR for cluster chart to `vt-cluster` instead

PR #2405, opened 2025-09-10
Fix/remote write object validation

sasha-sup:
In my previous request, I broke the variable validation.
I fixed it.


AndrewChubatiuk:
introduced a fix in [this commit](https://github.com/VictoriaMetrics/helm-charts/commit/ae37f9b1769fde06340970643b394210941560e2)

PR #2409, opened 2025-09-10
docs: reduce redirect in docs

makasim:
The PR only adds / `2F%` before `#` to shield links, nothing else


AndrewChubatiuk:
these links are crafted automatically in CI job, need to update job as well

makasim:
> these links are crafted automatically in CI job, need to update job as well

fixed, @AndrewChubatiuk please take a look

PR #2424, opened 2025-09-16
docs: fix links

makasim:


.github/workflows/release.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
                 if [ "$APP_VERSION" != "null" ]; then
                   export APP_VERSION_HASH="$(echo $APP_VERSION | tr -d '.' | cut -f1 -d '-')"
                   export APP_LABEL=$(echo $APP_VERSION | sed 's/-/--/g')
-                  export APP_CHANGELOG=$(yq '.annotations["artifacthub.io/links"] | from_yaml | .[] | select(.name == "Changelog").url + "/#" + env(APP_VERSION_HASH)| @uri' $CHART_FILE)
```
makasim, line 81: Not needed because the slash was added to artifacthub.io/links directly.

AndrewChubatiuk (approved):

PR #2439, opened 2025-09-26
Remove fixed http prefix from external Grafana url.

a-bali:
This patch removes the fixed http:// prefix from the external Grafana url, allowing the use of a prefix (http or https) specified as part of that url. This is also in line with how the external Grafana url is used in the dashboard link of the alerts (i.e. no prefix is prepended there).

### DIFF
--- a/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
+++ b/charts/victoria-metrics-k8s-stack/templates/_helpers.tpl
@@ -150,7 +150,7 @@
     {{- end -}}
     {{- if not (index $output.extraArgs "external.url") -}}
       {{- $grafanaHost := ternary (index (($Values.grafana).ingress).hosts 0) (($Values.external).grafana).host ($Values.grafana).enabled }}
-      {{- $_ := set $output.extraArgs "external.url" (printf "http://%s" $grafanaHost) -}}
+      {{- $_ := set $output.extraArgs "external.url" $grafanaHost -}}
     {{- end -}}
   {{- end -}}
   {{- tpl ($output | toYaml) . -}}


AndrewChubatiuk:
@a-bali have you tested this change?
i'm not sure it works
value from external.url flag is passed to [this function](https://github.com/VictoriaMetrics/VictoriaMetrics/blob/cea9505bab08bc930c86a7e7c96d6b6f7e5febc6/app/vmalert/main.go#L272), where:
- value without scheme will be treated as url path
- validation fails, since parsed url has neither `http` nor `https` scheme

a-bali:
But the value should already include the scheme, this is how it is used elsewhere (e.g. in the dashboards). Right now, if I include the scheme, it gets prepended with http:// and therefore obviously fails. I haven't tested the change but the function you linked should work fine if the scheme is included, and is right to fail if it isn't.

AndrewChubatiuk:
possible values for external.url is result of this expression `ternary (index (($Values.grafana).ingress).hosts 0) (($Values.external).grafana).host ($Values.grafana).enabled`, 
where `(($Values.external).grafana).host` can be set to anything you want, but `index (($Values.grafana).ingress).hosts 0` is dns name only
if you're setting `.Values.external.grafana).host` you should remove remove scheme

a-bali:
You are right, I was looking at the external.grafana.host scenario only, which should include the scheme. Then I guess the solution is to keep my change and add a scheme to the host coming from the ingress, no?

AndrewChubatiuk:
added ability to set optionally scheme at `.Values.external.grafana.host` in [this commit](https://github.com/VictoriaMetrics/helm-charts/commit/821b645f1eaf6fa4d1bb8e7400d4e1cb5e4e100e). will be available in the next release. thanks for initial implementation

AndrewChubatiuk:
this ability was released in k8s-stack 0.61.0

Issue #2517, opened 2025-10-30
feature: `victoria-metrics-k8s-stack` allow multiple vmsingle replicas (servers)


YuryHrytsuk:
**Chart name and version, where you feel a lack of requested feature**
chart: `victoria-metrics-k8s-stack`
version: v0.62.1

**Is your feature request related to a problem? Please describe.**
I want to deploy _highly available_ victoria-metrics cluster for Kubernetes (including pre-configured scrapers). I don't want to use `vm-distributed` chart as it looks an overkill for my setup (I want 1 vm-agent, 1 vm-auth and 2+ VM Single)

**Describe the solution you'd like**
Setting `replicaCount: 2` on `vmsingle` object would be ideal. Configuring 2 `vmsingle` (since `vmsingle` CR does not support >1 replicas) is also fine

**Describe alternatives you've considered**
1. I tried to configure all services on my own without operator CRs (using charts directly) but then I am missing pre-configured scrapers that I want
2. victoria-metrics-distributed is an overkill

**Additional context**
I am looking for an easy to configure simple highly available victoria metrics stack for Kubernetes. I don't want to configure scraper on my own (e.g. for kube-state-metrics, kubelet, ...). Moreover, documentation states that people should rather re-use k8s-stack or operator for scraper configuration

Haleygo:
Hi!
The HA you want is the reason `victoria-metrics-distributed` exists. Both `victoria-metrics-k8s-stack` and `victoria-metrics-distributed` charts provide HA ability, but in different ways:
1. `victoria-metrics-k8s-stack` achieves it by setting `-replicationFactor > 1`.
2. `victoria-metrics-distributed` achieves it by creating multiple independent vmsingle/vmcluster.

So if you want a single vmcluster, use `victoria-metrics-k8s-stack`; if you want multiple vmsingle/vmcluster, use `victoria-metrics-distributed`.

And in different charts, the configs are different. 
For example, with multiple independent vmsingle/vmcluster instances, there must be a vmagent to write data to all instances, a read proxy(like vmauth) to query data from all instances, and clients like vmalert or grafana must use the read proxy as their datasource. 
So for `victoria-metrics-k8s-stack`, it's not only about adding a new vmsingle, but also requires changing the configurations mentioned above, and all those things are already handled in `victoria-metrics-distributed`.

>I want 1 vm-agent, 1 vm-auth and 2+ VM Single

[...]

PR #2526, opened 2025-11-01
distributed: allow targeting extra vmagent to vmsingle/vmcluster directly

AndrewChubatiuk:
- extra vmagent can target VMSingle/VMCluster instances directly if `write` and `rw` balancers are disabled (related issue #2517)


YuryHrytsuk:
@AndrewChubatiuk could you explain the purpose of this PR? And why / how is it related to the issue linked?

AndrewChubatiuk:
@YuryHrytsuk updated description

YuryHrytsuk:
> @YuryHrytsuk updated description

Thanks

What confuses me is word _mixed balancer_. Based on what I read, to write data to VM Single / Cluster, one uses `vm-agent` and to read data from VM Single / Cluster one uses `vm-auth`.

So word mixed implies that there is only one entity doing both? But if `vm-agent` cannot read data and `vm-auth` cannot write data (probably my assumption is wrong), what is actually a _mixed balancer_?

AndrewChubatiuk:
this PR introduces ability to set remoteWrites for `extra` VMAgent (if it's enabled) to a list of VMSingle/VMCluster endpoints in global write balancer is disabled

Haleygo:
@AndrewChubatiuk please check fix for extra vmagent `-remoteWrite` arg in https://github.com/VictoriaMetrics/helm-charts/pull/2526/commits/a7fd168dad843835fc91b6f16e649998de7346e7.

charts/victoria-metrics-distributed/values.yaml
```diff
@@ -37,6 +37,39 @@ common:
...
+      retry_status_codes: [500, 502, 503]
+      src_paths:
+        - /select/.+
+        - /admin/tenants
+    vmsingle:
+      load_balancing_policy: first_available
+      retry_status_codes: [500, 502, 503]
+      drop_src_path_prefix_parts: 2
+      src_paths:
+        - /select/.+
```
Haleygo, line 71: The default src_paths for vmsingle read&write look a bit weird though.

Haleygo (approved):
Thanks, LGTM!
Let's merge it if you're ok with my commits.

YuryHrytsuk:
Why wouldn't it be like this by default (instead of vmauth-write-lb + 2 vm agents)? 

The new architecture (judging by image) still contains `vmauth-write-loadbalancer` and I don't understand why this is necessary

image

Haleygo:
> Why wouldn't it be like this by default (instead of vmauth-write-lb + 2 vm agents)?
> 
> The new architecture (judging by image) still contains `vmauth-write-loadbalancer` and I don't understand why this is necessary
> 
> image

See
>Note:
The default topology tolerates zone outages by deploying components in every availability zone and enabling minimum-downtime during outages. If not required, some components(including vmauth-global-write, vmagent(per-zone), vmauth-read-proxy(per-zone)) are optional and can be disabled based on your use case, please refer to [Parameters](#parameters) section for details.

and https://github.com/VictoriaMetrics/helm-charts/issues/2517#issuecomment-3479209371
>vmauth-global-write: global write entrypoint, proxies requests to one of the zone vmagent with least_loaded policy.
The main responsibility of vmagent(per-zone) is to [buffer data on disk](https://docs.victoriametrics.com/victoriametrics/vmagent/#calculating-disk-space-for-persistence-queue) when zone is unavailable to ingest.

---
`vmauth-global-write` and `vmagent(per-zone)` can be disabled as noted in https://github.com/VictoriaMetrics/helm-charts/issues/2517#issuecomment-3476739500. If disabled, the extra-vmagent (configured automatically in this PR) or other collectors/clients must be configured to write to both vmsingle/vmcluster. 
However, that won't be the default setup; by default, we aim to provide the maximum SLA and a default buffer layer. See also https://docs.victoriametrics.com/guides/vm-architectures/#multi-cluster-and-multi-az.

AndrewChubatiuk:
closing in a favour of https://github.com/VictoriaMetrics/helm-charts/pull/2732

PR #2530, opened 2025-11-03
charts/victoria-logs-collector: add the ability to switch to vlagent

vadimalekseev:
After discussing with Andrew, we decided to add a native flag in the victoria-logs-collector Helm chart. This flag allows switching from the Vector implementation to [vlagent](https://github.com/VictoriaMetrics/VictoriaLogs/issues/538).

These changes do not need to be documented, because vlagent still cannot collect logs from Kubernetes. The purpose of this update is solely to test the chart in our private cluster using the [draft implementation](https://github.com/VictoriaMetrics/VictoriaLogs/tree/k8s-collector) for Kubernetes log collection.

### DIFF
--- a/charts/victoria-logs-collector/templates/_helpers.tpl
+++ b/charts/victoria-logs-collector/templates/_helpers.tpl
@@ -11,10 +11,8 @@
 {{- define "victoria-logs-collector.image" }}
 {{- $registry := .Values.global.image.registry | default "docker.io" }}
 {{- if .Values.native }}
-  {{- if not .Values.image }}
-  {{- fail "'image' must be set when 'native' is true" }}
-  {{- end }}
-{{- printf "%s/%s:%s" $registry "victoriametrics/vlagent" .Values.image.tag }}
+{{- $tag := required "'image' must be set when 'native' is true" (.Values.image).tag }}
+{{- printf "%s/%s:%s" $registry "victoriametrics/vlagent" $tag }}
 {{- else }}
 {{- printf "%s/%s:%s" $registry "timberio/vector" "0.49.0-alpine" }}
 {{- end }}


charts/victoria-logs-collector/templates/_helpers.tpl
```diff
@@ -7,3 +7,66 @@
...
 {{- end }}
 {{- end }}
+
+{{- define "victoria-logs-collector.image" }}
+{{- $registry := .Values.global.image.registry | default "docker.io" }}
+{{- if .Values.native }}
+  {{- if not .Values.image }}
+  {{- fail "'image' must be set when 'native' is true" }}
+  {{- end }}
+{{- printf "%s/%s:%s" $registry "victoriametrics/vlagent" .Values.image.tag }}
```
AndrewChubatiuk:
```suggestion
{{- $tag := required "'image' must be set when 'native' is true" (.Values.image).tag }}
{{- printf "%s/%s:%s" $registry "victoriametrics/vlagent" $tag }}
```

zekker6:
> These changes do not need to be documented, because vlagent still cannot collect logs from Kubernetes. The purpose of this update is solely to test the chart in our private cluster using the [draft implementation](https://github.com/VictoriaMetrics/VictoriaLogs/tree/k8s-collector) for Kubernetes log collection.

Could you elaborate on benefits of merging this to upstream as opposed to simply using version from this branch?
As we're using ArgoCD it is possible to simply use ref pointing to this repo/branch for testing without releasing the chart.

AndrewChubatiuk:
> As we're using ArgoCD it is possible to simply use ref pointing to this repo/branch for testing without releasing the chart.

yes, it's true, but since this functionality is not documented or enabled by default it shouldn't confuse or affect anyone

PR #2562, opened 2025-11-11
fix(config-reloader): set default image to match the operator version

arteonprifti:
Currently the reloader version is the one here: https://github.com/VictoriaMetrics/operator/blob/2ee997984d1b6d8ec7c3f07eceb2aea40fc993b1/internal/config/config.go#L115 and it doesn't match the operator version.
This is to set it to match the operator version.

By the way before v0.62.0 setting 
```
    - name: VM_USECUSTOMCONFIGRELOADER
      value: "false"
```
was enough to get the same version for the reloader, but I didn't dig into that much. So maybe there is a better way, in the operator code.


AndrewChubatiuk:
LGTM, could you please also add a changelog entry `charts/victoria-metrics-operator/CHANGELOG.md`?

PR #2570, opened 2025-11-18
feat: add unit and end-to-end tests via terratest

### LINKED ISSUE #2569: feature: add unit/end-to-end tests
So far only "common" chart is being tested (`test-charts/` dir) and each chart is attempted to be installed with one configuration and basic check afterwards.

Helm chart testing should be extended to include the following:
* multiple configurations for each chart
* namespaces / release name setting being respected (needs to be randomized)
* improved verifications - i.e. necessary resources created, pods starting etc.
* negative scenarios - error messages, resource name limitations etc.

vrutkovs:
This adds tests for helm charts. These install charts and verify necessary resources are created.
Instead of using yaml configuration the test logic can be written in golang using terratest, testify
and other familiar tools.

Fixes #2569

TODO:
* [x] Fix Junit reporting
* [x] Make sure that all tests are passing
* [x] Fix chart linting

### DIFF
--- a/test/utils.go
+++ b/test/utils.go
@@ -1,0 +1,47 @@
+package test
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	apiextensionsclientset "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	pollingInterval     = 5 * time.Second
+	pollingTimeout      = 10 * time.Minute
+	resourceWaitTimeout = 1 * time.Minute
+)
+
+var (
+	retries = int(resourceWaitTimeout.Seconds() / pollingInterval.Seconds())
+)
+
+func helmCleanup(ctx context.Context, t *testing.T, k8sOpts *k8s.KubectlOptions, helmOpts *helm.Options, releaseName string) {
+	helm.Delete(t, helmOpts, releaseName, true)
+
+	kubeConfigPath, err := k8sOpts.GetConfigPath(t)
+	require.NoError(t, err)
+	clientConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: kubeConfigPath}, &clientcmd.ConfigOverrides{})
+	restConfig, err := clientConfig.ClientConfig()
+	require.NoError(t, err)
+	extendedClient, err := apiextensionsclientset.NewForConfig(restConfig)
+	require.NoError(t, err)
+
+	crds, err := extendedClient.ApiextensionsV1().CustomResourceDefinitions().List(ctx, metav1.ListOptions{})
+	require.NoError(t, err)
+	for _, crd := range crds.Items {
+		if crd.Spec.Group == "operator.victoriametrics.com" {
+			err = extendedClient.ApiextensionsV1().CustomResourceDefinitions().Delete(ctx, crd.Name, metav1.DeleteOptions{})
+			require.NoError(t, err)
+		}
+	}
+}
--- a/test/victoria-logs-cluster_test.go
+++ b/test/victoria-logs-cluster_test.go
@@ -1,0 +1,66 @@
+package test
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/gruntwork-io/terratest/modules/helm"
+	"github.com/gruntwork-io/terratest/modules/k8s"
+	"github.com/gruntwork-io/terratest/modules/random"
+	"github.com/stretchr/testify/require"
+	appsv1 "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+// TestVictoriaLogsClusterInstallDefault tests that the victoria-logs-cluster chart can be installed with default values.
+func TestVictoriaLogsClusterInstallDefault(t *testing.T) {
+	const helmChartPath = "../charts/victoria-logs-cluster"
+
+	namespaceName := fmt.Sprintf("vlogcluster-%s", strings.ToLower(random.UniqueId()))
+	k8sOpts := k8s.NewKubectlOptions("", "", namespaceName)
+
+	helmOpts := &helm.Options{
+		BuildDependencies: true,
+		KubectlOptions:    k8sOpts,
+		ExtraArgs: map[string][]string{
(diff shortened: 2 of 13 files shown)


AndrewChubatiuk:
@vrutkovs 
thanks for a PR
could you also remove `ct` execution from CI? it seems to be redundant

vrutkovs:
Sure, once these tests are setup and running I can remove the legacy tests

.github/workflows/run-testing.yaml
```diff
@@ -74,57 +47,47 @@ jobs:
...
-            helmv3
-            kubectl
-          helm: "${{ env.HELM_VERSION }}"
-          kubectl: "${{ matrix.k8s_version }}"
+          path: ./bin
+          key: binary
 
-      - uses: actions/setup-python@v5
+      - name: Setup Go
+        uses: actions/setup-go@v5
```
AndrewChubatiuk:
```suggestion
        uses: actions/setup-go@v6
```

.github/workflows/run-testing.yaml
```diff
@@ -74,57 +47,47 @@ jobs:
...
-              --config ../../.github/ci/ct.yaml \
-              --charts . \
-              --helm-extra-set-args "--values e2e/simple.yaml" --skip-clean-up
-          else
-            echo "skipping tests since to e2e values were found"
-          fi
+          cd test && go test -json ./... 2>&1 | go-junit-report -set-exit-code -parser gojson -iocopy -out ../report.xml
+
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v5
```
AndrewChubatiuk:
```suggestion
        uses: mikepenz/action-junit-report@v6
```

.github/workflows/run-testing.yaml
```diff
@@ -74,57 +47,47 @@ jobs:
...
-      - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.7.0
-        with:
-          version: 3.13.0
+      - name: Install go-junit-report
+        run: |
+          # This is a fork, so go install won't work
+          git clone https://github.com/joschi/go-junit-report
+          cd go-junit-report
+          go install ./cmd/...
```
AndrewChubatiuk:
```suggestion
          go install github.com/joschi/go-junit-report/v2/cmd/go-junit-report@latest
```

vrutkovs: golang is just soimething else, thanks!

.github/workflows/run-testing.yaml
```diff
@@ -74,57 +47,45 @@ jobs:
...
+          cache: true
+        id: go
 
-      - name: Set up chart-testing
-        uses: helm/chart-testing-action@v2.7.0
-        with:
-          version: 3.13.0
+      - name: Install go-junit-report
+        run: |
+          # This is a fork, so go install won't work
```
AndrewChubatiuk: suppose this comment could be removed)
```suggestion   
```

AndrewChubatiuk:
image

vrutkovs:
Yup, this workflow adds checks, so it throws an error with current settings.

It [passes cleanly](https://github.com/vrutkovs/helm-charts/actions/runs/19566820163) in my fork

test/go.mod
```diff
@@ -0,0 +1,130 @@
+module github.com/gruntwork-io/terratest-helm-testing-example/test
```
AndrewChubatiuk:
```suggestion
module github.com/VictoriaMetrics/helm-charts/test
```

.github/workflows/run-testing.yaml
```diff
@@ -26,105 +28,66 @@ env:
...
```
AndrewChubatiuk: why have you reverted this change?

.github/workflows/run-testing.yaml
```diff
@@ -26,105 +28,66 @@ env:
...
     strategy:
       matrix:
-        chart: ${{ fromJSON(needs.get-changed-charts.outputs.charts) }}
         k8s_version: ["v1.31.12", "v1.32.8", "v1.33.4"]
       fail-fast: false
     steps:
-      - name: Checkout
-        uses: actions/checkout@v5
+      - name: Checkout code
+        uses: actions/checkout@v4
```
AndrewChubatiuk: why action version was dowgraded?

.github/workflows/run-testing.yaml
```diff
@@ -26,105 +28,66 @@ env:
...
-            helmv3
-            kubectl
-          helm: "${{ env.HELM_VERSION }}"
-          kubectl: "${{ matrix.k8s_version }}"
+          path: ./bin
+          key: binary
 
-      - uses: actions/setup-python@v5
+      - name: Setup Go
+        uses: actions/setup-go@v5
```
AndrewChubatiuk: v6 is available already

.github/workflows/run-testing.yaml
```diff
@@ -26,105 +28,66 @@ env:
...
-              --config ../../.github/ci/ct.yaml \
-              --charts . \
-              --helm-extra-set-args "--values e2e/simple.yaml" --skip-clean-up
-          else
-            echo "skipping tests since to e2e values were found"
-          fi
+          cd test && go test -json ./... 2>&1 | go-junit-report -set-exit-code -parser gojson -iocopy -out ../report.xml
+
+      - name: Publish Test Report
+        uses: mikepenz/action-junit-report@v5
```
AndrewChubatiuk:
```suggestion
        uses: mikepenz/action-junit-report@v6
```

.github/workflows/run-testing.yaml
```diff
@@ -26,105 +28,66 @@ env:
...
-            ct install \
-              --release-name test \
-              --namespace ${{ matrix.chart }}-testing \
-              --config ../../.github/ci/ct.yaml \
-              --charts . \
-              --helm-extra-set-args "--values e2e/simple.yaml" --skip-clean-up
-          else
-            echo "skipping tests since to e2e values were found"
-          fi
+          cd test && go test -json ./... 2>&1 | go-junit-report -set-exit-code -parser gojson -iocopy -out ../report.xml
```
AndrewChubatiuk: have you tried running several threads?

.github/workflows/run-testing.yaml
vrutkovs: Oops, this is due to two branches desyncing

.github/workflows/run-testing.yaml
vrutkovs: It would make a mess due to CRDs. Some tests don't need it, some install and manage it via Helm and some use operator to do that. This is why test don't set `t.Parallel()` after recent changes

The safest option here it to run each test separately. Added `-p 1` just to be sure

.github/workflows/run-testing.yaml
vrutkovs: Fixed

.github/workflows/run-testing.yaml
AndrewChubatiuk: what about running tests without CRDs separately with concurrency enabled?

vrutkovs: Yeah, we can experiment with this (although I'd prefer to keep them as uniform as possible), but at the moments its probably not necessary - tests don't take too long now

PR #2572, opened 2025-11-20
victoria-logs-collector: globally configurable `timeField` and `msgField` parameters

vadimalekseev:
These changes are required to prepare the migration of the underlying implementation to an application that uses the `/internal/insert` handler, which does not support the `VL-Time-Field`, `VL-Msg-Field` headers. Therefore, the client must handle this logic on its own side.

Since it is challenging and unnecessary to use different `_msg` and `_time` fields for each remoteWrite, the `rw.msgField` parameter is now deprecated - clients should use the global `msgField` instead. Existing configurations will not be affected, as the new logic takes the old parameter into account.

### DIFF
--- a/charts/victoria-logs-collector/templates/configmap.yaml
+++ b/charts/victoria-logs-collector/templates/configmap.yaml
@@ -24,6 +24,14 @@ data:
       {{ fail "specify at least one remoteWrite with valid url"}}
     {{- end }}
     sinks:
+    {{/* Before 0.1.0, each remoteWrite could specify its own msgField. This was removed in favor of global msgField configuration. */}}
+    {{- $msgField := list }}
+    {{- range $rw := .Values.remoteWrite }}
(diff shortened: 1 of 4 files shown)


charts/victoria-logs-collector/templates/configmap.yaml
```diff
@@ -24,6 +24,16 @@ data:
       {{ fail "specify at least one remoteWrite with valid url"}}
     {{- end }}
     sinks:
+    {{/* Before 0.1.0, each remoteWrite could specify its own msgField. This was removed in favor of global msgField configuration. */}}
+    {{- $msgField := list }}
+    {{- range $rw := .Values.remoteWrite }}
+      {{- with $rw.msgField }}
+        {{ $msgField = concat $msgField . }}
+      {{- end }}
```
AndrewChubatiuk:
```suggestion
      {{ $msgField = concat $msgField ($rw.msgField | default list) }}
```

PR #2573, opened 2025-11-20
victoria-logs-collector: support time and msg cli flags

vadimalekseev:


charts/victoria-logs-collector/e2e/simple.yaml
```diff
@@ -30,4 +30,4 @@ resources:
 
 native: true
 image:
-  tag: heads-k8s-collector-0-g6d7a87224
\ No newline at end of file
+  tag: heads-k8s-collector-0-gbedc0a7f9
```
vadimalekseev, line 33: Updated vlagent tag just in case. This tag contains the latest changes as of 21.11.25

PR #2587, opened 2025-12-01
charts/victoria-logs-collector: update native mode

vadimalekseev:
* Renamed `-kubernetes` to `-kubernetesCollector` CLI-flags.
* Exposed 9429 port to allow collecting vlagent metrics.

### DIFF
--- a/charts/victoria-logs-collector/CHANGELOG.md
+++ b/charts/victoria-logs-collector/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next release
 
-- explicitly define namespace for namespaced resources. See [#2578](https://github.com/VictoriaMetrics/helm-charts/issues/2578).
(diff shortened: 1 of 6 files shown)


AndrewChubatiuk:
@vadimalekseev 
could you please also add podscrape once you already expose metrics?

vadimalekseev:
@AndrewChubatiuk, thank you for the quick review! I have added VMPodScrape. The configuration is taken from [vm-anomaly helm chart](https://github.com/VictoriaMetrics/helm-charts/blob/4b1f7be35c22532e82a70b821ffe2f14d05323ae/charts/victoria-metrics-anomaly/values.yaml#L141-L149) (since it looked simple), but I removed the Basic, scheme and TLS fields, since victoria-logs-collector does not export these parameters.

AndrewChubatiuk (approved):

Issue #2579, opened 2025-11-23
feature: create victoria-logs-multilevel Helm chart for building multi-level cluster setups for VictoriaLogs


valyala:
## The issue

VictoriaLogs supports flexible ability to build multi-level cluster setups according to [these docs](https://docs.victoriametrics.com/victorialogs/cluster/#multi-level-cluster-setup). This feature allows configuring `vlselect` for querying multiple independent VictoriaLogs clusters or multiple independent single-node instances of VictoriaLogs, or a mixed set of clusters and single-node instances. This helps building VictoriaLogs scheme with multiple distinct retentions per distinct groups of logs according to [this comment](https://github.com/VictoriaMetrics/VictoriaLogs/issues/15#issuecomment-3043557052).

The existing Helm charts for VictoriaLogs do not support building multi-level cluster setups.

## The solution

To add a new Helm chart - `victoria-logs-multilevel`, which will deploy the given number of `vlselect` nodes, which query the given set of addresses, which speak `vlstorage` querying protocol (these can be bare `vlstorage` nodes, single-node VictoriaLogs nodes (deployed via [victoria-logs-single Helm chart](https://docs.victoriametrics.com/helm/victoria-logs-single/)) or VictoriaLogs clusters hidden behind `vmauth` (deployed via [victoria-logs-cluster Helm chart](https://docs.victoriametrics.com/helm/victoria-logs-cluster/)) - see [single-node and cluster mode duality docs](https://docs.victoriametrics.com/victorialogs/cluster/#single-node-and-cluster-mode-duality).

The Helm chart must provide the following essential configs:

- A list of addresses to query by `vlselect`. These addresses must be passed to `-storageNode` command-line flag at `vlselect`.
- The number of `vlselect` replicas to deploy
- Limits on memory and CPU per each `vlselect` replica

The Helm chart must deploy the given number of `vlselect` nodes hidden behind `vmauth`. The `vmauth` must accept incoming queries and spread them evenly among the available `vlselect` nodes. The `vmauth` address must be exposed by the chart.

This proposal has been extracted from the https://github.com/VictoriaMetrics/helm-charts/issues/2336#issuecomment-3523405812 .

AndrewChubatiuk:
@valyala basically this is the same as victorialogs single chart, but without storageNodes parameter. What do you think about just adding storageNodes parameter to single chart, that will switch it to multi-level select mode and should cover this functionality instead of introducing another copy paste chart?

[...]

Issue #2336, opened 2025-08-04
feature: create a Helm chart for VictoriaLogs cluster, which allows running dedicated groups of storage nodes with distinct configs per different groups of tenants


valyala:
**Is your feature request related to a problem? Please describe.**

Some users of VictoriaLogs need the ability to configure different retention for different subsets of logs. See https://github.com/VictoriaMetrics/VictoriaLogs/issues/15 . This functionality won't be implemented at the VictoriaLogs level because of the issues described at https://github.com/VictoriaMetrics/VictoriaLogs/issues/15#issuecomment-3043557052 . But this functionality can be implemented in a dedicated Helm chart (let's name it VictoriaLogs enterprise).

**Describe the solution you'd like**

The Helm chart should provide the ability to setup multiple independent groups of [vlstorage nodes](https://docs.victoriametrics.com/victorialogs/cluster/) with distinct configs on a per-tenant basis. Every such a group of vlstorage nodes must have the following configs:
- The number of vlstorage nodes
- per-node CPU and RAM limits
- per-node persistent volume size
- [retention](https://docs.victoriametrics.com/victorialogs/#retention)

Additional configs can be added in the future.

The number of vlstorage groups can be arbitrary. Every vlstorage group must be identified by a list of [`AccountID` header values](https://docs.victoriametrics.com/victorialogs/#multitenancy). The `ProjectID` part of the tenant ID must be set to `0` when ingesting data to all the vlstorage groups.

At least a single `vlinsert` node must be put in front of every `vlstorage` group. The Helm chart must provide an ability to configure per-group `vlinsert` nodes with the following options:
- The number of `vlinsert` nodes
- per-node CPU and RAM limits

The Helm chart must run an [vmauth](https://docs.victoriametrics.com/victoriametrics/vmauth/) in front of all the `vlstorage` groups. The `vmauth` must proxy the ingested data to the corresponding `vlstorage` group according to the `AccountID` HTTP request header. The `vmauth` must proxy SELECT queries to a global group of `vlselect` nodes, which query all the `vlstorage` nodes across all the groups. This provides global querying view - e.g. a single query can return all the data from all the tenants.

**Additional context**

This Helm chart is needed mostly for Enterprise customers, so it must be protected by an Enterprise agreement. Non-enterprise users can use the chart for evaluation purposes, but not in production.

makasim:
[...]

PR #2591, opened 2025-12-01
introduce victoria-logs-multiselect chart

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/2579
just requires `.Values.storageNodes` to be defined:
```
storageNodes:
 - vlselect-n-cluster-1
 - vlselect-n-cluster-2
```
the same setup can be achieved using `victoria-logs-cluster` chart with given below values
```
vlinsert:
  enabled: false
vlstorage:
  enabled: false
vmauth:
  enabled: true
vlselect:
  storageNodes:
    - vlselect-n-cluster-1
    - vlselect-n-cluster-2
```

### DIFF
--- a/test/victoria-logs-cluster_test.go
+++ b/test/victoria-logs-cluster_test.go
@@ -3,12 +3,9 @@ package test
 import (
 	"context"
 	"fmt"
-	"strings"
 	"testing"
 
-	"github.com/gruntwork-io/terratest/modules/helm"
 	"github.com/gruntwork-io/terratest/modules/k8s"
-	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -17,32 +14,20 @@ import (
 
 // TestVictoriaLogsClusterInstallDefault tests that the victoria-logs-cluster chart can be installed with default values.
 func TestVictoriaLogsClusterInstallDefault(t *testing.T) {
-	const helmChartPath = "../charts/victoria-logs-cluster"
-
-	namespaceName := fmt.Sprintf("vlogcluster-%s", strings.ToLower(random.UniqueId()))
-	k8sOpts := k8s.NewKubectlOptions("", "", namespaceName)
-
-	helmOpts := &helm.Options{
-		BuildDependencies: true,
-		KubectlOptions:    k8sOpts,
-		ExtraArgs: map[string][]string{
-			"upgrade": {"--create-namespace", "--wait"},
-		},
-	}
-
-	// Install the chart and verify no errors occurred.
-	releaseName := fmt.Sprintf("vlogcluster-%s", strings.ToLower(random.UniqueId()))
-	defer helmCleanup(context.Background(), t, k8sOpts, helmOpts, releaseName)
-	helm.Upgrade(t, helmOpts, helmChartPath, releaseName)
-
-	k8sClient, err := k8s.GetKubernetesClientFromOptionsE(t, k8sOpts)
-	require.NoError(t, err)
+	t.Parallel()
+	name := "victoria-logs-cluster"
+	ctx := context.Background()
+	cp := chartInstall(t, name, nil)
+	defer chartCleanup(t, ctx, cp)
 
 	// Verify log-storage StatefulSet was created and is ready using manual polling
-	logStorageStatefulSetName := fmt.Sprintf("%s-victoria-logs-cluster-vlstorage", releaseName)
+	releaseName := cp.releaseName
+	o := cp.opts
+	namespaceName := o.KubectlOptions.Namespace
+	logStorageStatefulSetName := fmt.Sprintf("%s-%s-vlstorage", releaseName, name)
 	var logStorageStatefulSet *appsv1.StatefulSet
-	err = wait.PollUntilContextTimeout(context.Background(), pollingInterval, pollingTimeout, true, func(ctx context.Context) (done bool, err error) {
-		logStorageStatefulSet, err = k8sClient.AppsV1().StatefulSets(namespaceName).Get(ctx, logStorageStatefulSetName, metav1.GetOptions{})
+	err := wait.PollUntilContextTimeout(ctx, pollingInterval, pollingTimeout, true, func(ctx context.Context) (done bool, err error) {
+		logStorageStatefulSet, err = cp.client.AppsV1().StatefulSets(namespaceName).Get(ctx, logStorageStatefulSetName, metav1.GetOptions{})
 		if err != nil {
 			return false, nil
 		}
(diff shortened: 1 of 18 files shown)


szibis:
I don't see in this PR any changes for `victoria-logs-cluster` adding `storageNodes` is there anywhere available now ?

szibis:
ok I see helper here https://github.com/VictoriaMetrics/helm-charts/blob/16e18de7a46f06985c012938a8650240ce924ff2/charts/victoria-logs-cluster/templates/_helpers.tpl#L10

AndrewChubatiuk:
@szibis 
victoria-logs-cluster chart has no `storageNodes` parameter, but has `vlselect.storageNodes`

AndrewChubatiuk:
@vrutkovs could you please take a look?

vrutkovs:
This looks good to me, lets add an end2end test (see https://github.com/VictoriaMetrics/helm-charts/tree/master/test)

charts/victoria-logs-multilevel/values.yaml
```diff
@@ -0,0 +1,579 @@
...
+  automountToken: true
+
+# -- Override chart name
+nameOverride: ""
+
+# -- Add extra specs dynamically to this chart
+extraObjects: []
+
+# -- Storage nodes for multilevel setuo
+storageNodes: [test]
```
Haleygo: Since `storageNodes` is mandatory for vlselect, how about setting the default value to null and failing the installation if it is undefined? 
I don't see a scenario where the user needs to install it prior to installing a Victorialog instance.

AndrewChubatiuk: it was mistakenly added

AndrewChubatiuk:
> I think the chart name victoria-logs-multilevel is both misleading and confuing, as it only contains vlselect deployment and relies on other charts or deployments.

it was done as it was requested in issue, chart contains not only vlselect, but vmauth.

> What's the advantage of maintaining this separate chart compared to adding an additional vlselect deployment to the existing vlogs single or cluster chart?

for me it's hard to defend idea of having this chart, as my initial suggestion was just to extend existing chart. it's better to ask in attached issue if you want

vrutkovs:
Not a big Helm chart fan myself, but it certainly useful to have one - as customers would prefer a supported solution instead of home-brewing configurations again and again.

I'd prefer an operator update (as it would allow us to do more complicated configuration verifications)

Haleygo:
>it was done as it was requested in issue, chart contains not only vlselect, but vmauth.

vmauth has nothing to do with "logs-multilevel", I think we can exclude it from our discussion.

>for me it's hard to defend idea of having this chart, as my initial suggestion was just to extend existing chart. it's better to ask in attached issue if you want

The latest comment from @valyala states: "Adding unrelated functionality to this chart complicates its' usage and configuration, and makes users unhappy. "
Which doesn't seem entirely valid to me, as the proposed addition would be a separate component that does not affect or bother the existing users or users that don't want it. So I thought there was some additional internal discussions and consensus on this matter.

Haleygo:
> Not a big Helm chart fan myself, but it certainly useful to have one - as customers would prefer a supported solution instead of home-brewing configurations again and again.
> 
> I'd prefer an operator update (as it would allow us to do more complicated configuration verifications)

My question is: do we need an independent chart (which requires additional maintenance) to deploy components that can be integrated into the existing setup?

Let's say users want to have two vlcluster with different retention config and a global read entry, what would we recommend as the best approach?
With this proposal, we recommend deploying vlogs-cluster chart twice and then victoria-logs-multilevel chart. With every version update, users need to perform 3 helm update, which doesn't sound a final solution to me and not more efficient than crafting some custom configurations.
Including this additional vlselect&vmauth into one of the vlogs-cluster chart also does not simplify the process much.

If the logs team **considers the use case valid and believes it should be natively supported by the community**, should we implement the overarching solution, a single chart that support both multiple vlogs setups and a global read entrypoint like https://github.com/VictoriaMetrics/helm-charts/issues/2336?

vrutkovs:
>With this proposal, we recommend deploying vlogs-cluster chart twice and then victoria-logs-multilevel chart. With every version update, users need to perform 3 helm update, which doesn't sound a final solution to me and not more efficient than crafting some custom configurations.

Yes, operational overhead with charts seems to be significant. imo operator could have solved it more efficiently (seems the concern here is the same as in VMDistributedCluster). That said, having a chart for a start may be a good idea, given that the user is aware of its limitations.

I don't have a preference over new chart vs updating existing one - both options seems insufficient for proper operations anyway

AndrewChubatiuk:
@vrutkovs 
> lets add an end2end test

added and updated a bit old ones

test/victoria-logs-cluster_test.go
```diff
@@ -3,46 +3,31 @@ package test
...
 	"github.com/gruntwork-io/terratest/modules/k8s"
-	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/stretchr/testify/require"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-// TestVictoriaLogsClusterInstallDefault tests that the victoria-logs-cluster chart can be installed with default values.
+// TestVictoriaLogsClusterInstallDefault tests that the %s chart can be installed with default values.
```
vrutkovs:
```suggestion
// TestVictoriaLogsClusterInstallDefault tests that the victoria-logs-cluster chart can be installed with default values.
```

test/victoria-logs-cluster_test.go
```diff
@@ -3,46 +3,31 @@ package test
...
-	}
-
-	// Install the chart and verify no errors occurred.
-	releaseName := fmt.Sprintf("vlogcluster-%s", strings.ToLower(random.UniqueId()))
-	defer helmCleanup(context.Background(), t, k8sOpts, helmOpts, releaseName)
-	helm.Upgrade(t, helmOpts, helmChartPath, releaseName)
-
-	k8sClient, err := k8s.GetKubernetesClientFromOptionsE(t, k8sOpts)
-	require.NoError(t, err)
+	t.Parallel()
```
vrutkovs, line 17: Does this chart use CRDs? If yes we can't use Parallel here as it may clash with other charts installing CRDs like vmoperator and k8s-stack

test/victoria-logs-multilevel_test.go
```diff
@@ -0,0 +1,45 @@
...
+	err := wait.PollUntilContextTimeout(ctx, pollingInterval, pollingTimeout, true, func(ctx context.Context) (done bool, err error) {
+		logSelectDeployment, err = cp.client.AppsV1().Deployments(namespaceName).Get(ctx, logSelectDeploymentName, metav1.GetOptions{})
+		if err != nil {
+			return false, nil
+		}
+		// Ensure all replicas are ready
+		return logSelectDeployment.Status.ReadyReplicas == *logSelectDeployment.Spec.Replicas && *logSelectDeployment.Spec.Replicas > 0, nil
+	})
+	require.NoError(t, err)
+	require.NotNil(t, logSelectDeployment)
```
vrutkovs: Why not `k8s.WaitUntilDeploymentAvailable(t, k8sOpts, logSelectDeploymentName, retries, pollingInterval)`?

test/victoria-logs-cluster_test.go
AndrewChubatiuk: no, it doesn't

vrutkovs (approved):

PR #2610, opened 2025-12-12
Feature: Added support for custom stream fields

Seshathri-saravanan:
- Added support for custom stream fields
- Currently the stream fields are hardcoded as `kubernetes.pod_name,kubernetes.container_name,kubernetes.pod_namespace`

### DIFF
--- a/charts/victoria-logs-collector/templates/configmap.yaml
+++ b/charts/victoria-logs-collector/templates/configmap.yaml
@@ -58,7 +58,7 @@ data:
           headers:
             AccountID: {{ or $rw.accountID 0 | quote }}
             ProjectID: {{ or $rw.projectID 0 | quote }}
-            VL-Stream-Fields: kubernetes.pod_name,kubernetes.container_name,kubernetes.pod_namespace
+            VL-Stream-Fields: {{ join "," ($.Values.streamField) }}
             VL-Msg-Field: {{ join "," ($msgField | default (list "message")) }}
             VL-Time-Field: {{ join "," ($.Values.timeField | default (list "timestamp")) }}
             {{- with $rw.extraFields }}
--- a/charts/victoria-logs-collector/values.yaml
+++ b/charts/victoria-logs-collector/values.yaml
@@ -45,6 +45,14 @@ msgField:
   - message
   - msg
 
+# -- List of fields to be used as `_stream` field.
+# See https://docs.victoriametrics.com/victorialogs/keyconcepts/#stream-fields for more details
+streamField:
+  - kubernetes.pod_name
+  - kubernetes.container_name
+  - kubernetes.pod_namespace
+
+  
 # -- Enable VLAgent logs collection
 native: false
 


vadimalekseev:
@Seshathri-saravanan @AndrewChubatiuk I am not sure that setting custom stream fields is a good idea, as it complicates the victoria-logs-collector chart configuration. The default streams `kubernetes.pod_name`, `kubernetes.container_name`, and `kubernetes.pod_namespace`  are well-suited for applications deployed in Kubernetes.

Also the `/internal/insert` endpoint used by vlagent does not support the `VL-Stream-Fields` header. This means we must implement support for configuring custom streams within the vlagent Kubernetes collector itself, and only after this we should support this feature in the victoria-logs-collector chart.

Seshathri-saravanan:
Hey @vadimalekseev `kubernetes.pod_name` field have very high cardinality for applications with frequent deployments and dynamic scaling policies. Exluding kubernetes.pod_name from streamId helps with better compression and disk usage.  `/insert/jsonline` supports `VL-Stream-Fields` header and thats being used in this chart.  I'm open for contribution for `/internal/insert` changes.

vadimalekseev:
Hi @Seshathri-saravanan,  VictoriaLogs should perform well with tens of thousands of unique `kubernetes.pod_name` values for a single `kubernetes.container_name` and `kubernetes.pod_namespace` pair. This is expected to improve the compression ratio, as there is a much higher probability that logs from the same pod will be stored within the same data block. This happens because each pod has its own unique set of fields such as `kubernetes.pod_node_name`, `kubernetes.pod_name`, `kubernetes.pod_ip` (and sometimes kubernetes.labels) - which are more likely to reside within the same data block.

Have you observed any improvement in the compression ratio on your graphs after removing `kubernetes.pod_name` from the stream fields?

AndrewChubatiuk:
> I am not sure that setting custom stream fields is a good idea, as it complicates the victoria-logs-collector chart configuration.

how it complicates configuration?

Seshathri-saravanan:
@vadimalekseev We started storing raw_msg(stringified version of the message with all fields) in message field for full text search over all fields. That had decreased the compression ratio from 50 to 20 (Expected). Because of this big change, we didnt see any visible difference with streamId change. But streamIds were growing day by day before(In order of 10 thousands) ? 

We recently started with victorialogs. Isn't the growing streamIds a bad thing ? Wont it affect the performance ?

vadimalekseev:
> how it complicates configuration?

Hi @AndrewChubatiuk, to effectively configure stream fields, users must understand what they are and be able to correctly select fields so as not to negatively impact the compression ratio and user query execution speed. I believe our default values are suitable for all (or at least most) scenarios.

If we add this setting, users will try to customize it, assuming it is a crucial parameter that must be changed for optimal log storage efficiency. But they are more likely to degrade performance. Therefore this setting significantly increases the barrier to entry for VictoriaLogs, IMO.

I am open to these changes, but we need to understand if this functionality is necessary for the Kubernetes collector. Also, as I mentioned previously, we must first implement this functionality in vlagent to maintain backward compatibility when migrating from Vector to vlagent.

vadimalekseev:
Perhaps we should add the ability to append stream fields (without changing the default values)? In this way, if availability zone information is stored in the labels, it would be beneficial to include it in the streams to accelerate query execution.

func25:
I think excluding `pod_name` is reasonable in specific environments, but I may be biased since I rarely use `pod_name` (of deployment) values like `service-7c9f6b4c9d-k2m7p` in queries.

> Isn't the growing streamIds a bad thing ? Wont it affect the performance ?

A growing number of streams isn't automatically bad, very fast growth is the typical red flag. Also `pod_name` is often stable enough while also improving locality (logs from the same pod tend to share many fields).

> to effectively configure stream fields, users must understand what they are and be able to...

Both the chart and the binary should have sensible default values (the current ones), and they should only be changed when required (I guess this need will arise sooner or later). + VictoriaLogs users should understand streams, it is very unlikely that someone would change them blindly :)

vadimalekseev:
Hello @Seshathri-saravanan, we've decided to revert this PR as we want to keep it as simple to use as possible. 
However, we will support overriding the `_stream` fields in vlagent, and in the next release of the victoria-logs-collector helm chart you will be able to override them using extra command-line flags section.

PR #2612, opened 2025-12-12
Feature: Added support for custom vector transformer source

Seshathri-saravanan:
- Added  a new key `vectorTransformsJsonparserSource` which defaults to `".message = parse_json(.message) ?? .message"`
- Usecase: We want to set the whole message as raw_msg field. Currently Its impossible with this. 
Eg: 

> source: |
>           .raw_msg = .message 
>           . = merge!(., parse_json(.message) ?? .message)
> 
This enables full text search over all the fields in message

### DIFF
--- a/charts/victoria-logs-collector/templates/configmap.yaml
+++ b/charts/victoria-logs-collector/templates/configmap.yaml
@@ -20,7 +20,7 @@ data:
         type: remap
         inputs: [k8s]
         source: |
-          .message = parse_json(.message) ?? .message
+          {{ .Values.vectorTransformsJsonparserSource | nindent 10 }}
     {{- if empty .Values.remoteWrite }}
       {{ fail "specify at least one remoteWrite with valid url"}}
     {{- end }}
--- a/charts/victoria-logs-collector/values.yaml
+++ b/charts/victoria-logs-collector/values.yaml
@@ -45,6 +45,10 @@ msgField:
   - message
   - msg
 
+# -- Custom VRL source code for the json_parser transform in Vector.
+# Only used if `native` is false.
+vectorTransformsJsonparserSource: ".message = parse_json(.message) ?? .message"
+
 # -- Enable VLAgent logs collection
 native: false
 


AndrewChubatiuk:
hey @Seshathri-saravanan 
unfortunately this will not be supported
we're planning to switch collector chart from vector to vlagent soon, where this option makes no sense

Seshathri-saravanan:
> hey @Seshathri-saravanan unfortunately this will not be supported we're planning to switch collector chart from vector to vlagent soon, where this option makes no sense

Hi @AndrewChubatiuk 
Can't we make it dynamic for now ? Until the switch happens ?
We use this config which enables full text search over all the fields in message. This will be good workaround for this issue https://github.com/VictoriaMetrics/VictoriaLogs/issues/82

```
source: |
  .raw_msg = .message
  . = merge!(., parse_json(.message) ?? .message)
msgField:
  - raw_msg
remoteWrite: 
    ignoreFields:
      - raw_msg*
```

AndrewChubatiuk:
will discuss it internally and let you know

Seshathri-saravanan:
@AndrewChubatiuk  Cool thanks!

AndrewChubatiuk:
@vadimalekseev @func25 
need your opinion here

vadimalekseev:
Hey @Seshathri-saravanan, I have found another workaround that might help you: https://play-vmlogs.victoriametrics.com/select/vmui/?#/?query=*+%7C+pack_json+%7C+fields+_msg%2C+_time+%7C+filter+_msg%3A+play-otel&g0.range_input=5m&g0.end_input=2025-12-15T11%3A40%3A26&g0.relative_time=last_5_minutes&hide_chart=true

This query packs all fields into `_msg` and performs filtering on it. See the [pack_json](https://docs.victoriametrics.com/victorialogs/logsql/#pack_json-pipe) pipe.

It has one drawback: the query execution speed will be significantly lower, but the compression ratio will be much better compared to storing unstructured logs in the `_msg` field.

Is this approach working for you?

Seshathri-saravanan:
> Hey @Seshathri-saravanan, I have found another workaround that might help you: https://play-vmlogs.victoriametrics.com/select/vmui/?#/?query=*+%7C+pack_json+%7C+fields+_msg%2C+_time+%7C+filter+_msg%3A+play-otel&g0.range_input=5m&g0.end_input=2025-12-15T11%3A40%3A26&g0.relative_time=last_5_minutes&hide_chart=true
> 
> This query packs all fields into `_msg` and performs filtering on it. See the [pack_json](https://docs.victoriametrics.com/victorialogs/logsql/#pack_json-pipe) pipe.
> 
> It has one drawback: the query execution speed will be significantly lower, but the compression ratio will be much better compared to storing unstructured logs in the `_msg` field.
> 
> Is this approach working for you?

Thanks for the suggestion. Will try this out. This will definitely increase the compression ratio a lot.  How this internally works does this uses all the bloom filters (all fields + _msg) ? Is this same as `*:searchValue` as requested in this [issue](https://github.com/VictoriaMetrics/VictoriaLogs/issues/82) ?

vadimalekseev:
> Thanks for the suggestion. Will try this out. This will definitely increase the compression ratio a lot. How this internally works does this uses all the bloom filters (all fields + _msg) ? 

Bloom filters are not used here because we read all logs from disk due to the initial filter before pipes (`*` at the start of the query).  Then this data is transformed in memory into a JSON-formatted string using the [`pack_json`](https://docs.victoriametrics.com/victorialogs/logsql/#pack_json-pipe) pipe. VictoriaLogs applies further filters to the resulting string (in memory).

> Is this same as *:searchValue as requested in this https://github.com/VictoriaMetrics/VictoriaLogs/issues/82 ?

This allows to achieve a similar result, but the option using `*:value` should be significantly faster because I guess it will use bloom filters to skip blocks on disk before actually reading the data. This should reduce memory consumption and disk load compared to the option I suggested.

However, because there is currently no built-in support for searching across all fields, and if you do not frequently need to search all fields, I believe the `pack_json` workaround should be more preferable compared to storing all unstructured data in the `_msg` field.

Seshathri-saravanan:
Unfortunately in our org, the logs dont have structured fields, most of our search patterns are without any field filters. Using `pack_json` wont be helping here. We will go with the approach of storing all the fields inside _msg. We are looking forward for this feature `*:value`.

PR #2632, opened 2025-12-24
collector: migrate from vector to vlagent, align remoteWrite parameters with vlagent command line flags

AndrewChubatiuk:
it's complicated to make parameters from remoteWrite section compatible for both vector and vlagent, also command line flags generation is complicated. replaced remoteWrite args parameters with ones that are similar to vlagent command line argument names. this allows to generate per remoteWrite command line arguments properly without a need to introduce cases for each section independently

additionally replaced vector with vlagent and added new collector arguments support


charts/victoria-logs-collector/templates/_helpers.tpl
```diff
@@ -1,3 +1,62 @@
+{{- define "victoria-logs-collector.headers" -}}
+  {{- $Values := (.helm).Values | default .Values -}}
+  {{- $idx := .idx }}
+  {{- $rw := index $Values.remoteWrite $idx }}
+  {{- $headers := deepCopy ($rw.headers | default dict) }}
+  {{- $_ := set $headers "VL-Stream-Fields" "kubernetes.pod_name,kubernetes.container_name,kubernetes.pod_namespace" }}
+  {{- $_ := set $headers "VL-Msg-Field" ($Values.msgField | default (list "message")) }}
```
vadimalekseev: VictoriaLogs reports a warning after installing this chart:

```
/insert/native endpoint doesn't support setting time fields via _time_field query arg and via ...
```

`_time` fields should be passed via `kubernetesCollector.timeField` because vlagent implements custom logic for this flag. See:

https://github.com/VictoriaMetrics/VictoriaLogs/blob/faee008c9e348d27016141aa335b90c14d79fc5a/app/vlagent/kubernetescollector/processor.go#L223

vlagent does not have custom logic for the `_stream` and `_msg` fields, but corresponding headers [are not supported](https://github.com/VictoriaMetrics/VictoriaLogs/blob/4c3b8521300071d20d21f581c1d3a16fd1e068af/app/vlinsert/nativeinsert/nativeinsert.go#L59-L68) in `/insert/native` endpoint

AndrewChubatiuk: updated PR

PR #2642, opened 2026-01-03
 Add vlagent chart serviceMonitor

szibis:
This is for the same scrape compatibility as for other Victoira components to be monitored and then vlagent dashboard can be reused same way as other dashboards.


AndrewChubatiuk (approved):

PR #2654, opened 2026-01-12
charts/vmanomaly - appVersion 1.28.4

fred-navruzov:
Q: Should it be merged after (missing) v1.28.3 PR #2653 is live, changelogs are updated and charts are released (?)


zekker6:
I'd suggest to just close #2653 and merge this one to produce one release of the helm chart. There are no requirements or other constraints which would force to create a helm chart version for every single VMAnomaly release.

fred-navruzov:
closed in favor of #2653 where we bumped version to latest (v1.28.7)

PR #2657, opened 2026-01-15
victoria-metrics-k8s-stack: unpin datasource version

vrutkovs:
Make sure we don't set an old victoriametrics-metrics-datasource version in default values


AndrewChubatiuk:
@vrutkovs could you sync branch with master?

PR #2671, opened 2026-01-23
operator: Fix image URL for configReloader

MariusSchiffer:
In my previous PR, I accidentally mixed up repository and registry (which I also did in my test values).
It also did not correctly insert the / into the image URL

Sorry about that, should've fully tested it beforehand


AndrewChubatiuk:
thanks for a PR
closing it as added [this commit](https://github.com/VictoriaMetrics/helm-charts/commit/5d5d73256fa136322551fb224cdf5ec90851f672) instead

PR #2680, opened 2026-01-29
victoria-metrics-operator: Add shareProcessNamespace flag

JHeilCoveo:
Recently we had to instrument the operator with [opentelemetry-go-instrumentation](https://github.com/open-telemetry/opentelemetry-go-instrumentation/blob/main/docs/getting-started.md#instrument-an-application-in-kubernetes) due to a performance issue.

The `extraContainer` configuration was useful but we also add to enable `shareProcessNamespace` on the pod. We did it with a postrender but it would be nice to be able to do it directly in the chart.

### DIFF
--- a/charts/victoria-metrics-operator/CHANGELOG.md
+++ b/charts/victoria-metrics-operator/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next release
 
-- TODO
+- added .Values.shareProcessNamespace
 
 ## 0.58.1
 
--- a/charts/victoria-metrics-operator/templates/server.yaml
+++ b/charts/victoria-metrics-operator/templates/server.yaml
@@ -37,6 +37,9 @@ spec:
       {{- if .Values.hostNetwork }}
       hostNetwork: true
       {{- end }}
+      {{- if .Values.shareProcessNamespace }}
+      shareProcessNamespace: true
+      {{- end }}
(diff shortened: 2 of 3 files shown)


JHeilCoveo:
Sure @AndrewChubatiuk.

For context we are in the process of migrating our Prom/Thanos stack to VM and to that effect we still have a bit over 600 prometheusrules in a single cluster. To make life easier for our developers we are using the prometheusrule -> VMRules sync mechanism from the operator.

But we had huge delays between the moment the prometheusrules are created and the moment the vmrule exist. We are speaking tens of minutes or even more than one hour is some extreme cases.

After doing the usual increase resources dance with no success we decided to instrument the code as we already have a tracing stack at hand.
I added tracing to victoria-metrics-operator and then I saw multiple POST/PUT queries stuck at 10 seconds. Then after reading the Kubernetes kubeapi logs it all clicked in my head.

1. A slew of prometheusrule get created/deleted by CI/CD
2. vm-operator picks up all the changes
3. vm-operator pushes creation/deletion of vmrules to the cluster
4. Validatingwebhook gets called and poke vm-operator in return
5. vm-operator is already maxed out on threads/goroutines (this is speculation on my part) and thus the hook times out.

In the end the fix was to disable the admissionWebhooks for the duration of the migration.

Note : Somehow I should have noticed but didn't see the almost solid line at 10s in Grafana. That should have tipped me in the direction of kubeapi logs.
image

PR #2760, opened 2026-03-13
charts: document enableTCP6 flag

andriibeee:
## Summary by cubic
Documented the `enableTCP6` flag across VictoriaMetrics Helm charts to show how to enable IPv6 in IPv6-only clusters. Adds commented examples in values.yaml for agent, alert, auth, cluster (vmselect, vminsert, vmauth, vmstorage), gateway, and single; docs-only with no behavior change.

<sup>Written for commit c9dc8d2d487e4b3edd395465c5630021668a6766. Summary will update on new commits.</sup>

### DIFF
--- a/charts/victoria-metrics-agent/values.yaml
+++ b/charts/victoria-metrics-agent/values.yaml
@@ -139,7 +139,7 @@ extraArgs:
   envflag.prefix: VM_
   loggerFormat: json
   httpListenAddr: :8429
-  # -- Enable IPv6 support. Useful for running in IPv6-only Kubernetes clusters
+  # Enable IPv6 support. Useful for running in IPv6-only Kubernetes clusters
   # enableTCP6: true
   # promscrape.maxScrapeSize: "167772160"
 
--- a/charts/victoria-metrics-agent/values.yaml
+++ b/charts/victoria-metrics-agent/values.yaml
@@ -139,6 +139,8 @@ extraArgs:
   envflag.prefix: VM_
   loggerFormat: json
   httpListenAddr: :8429
+  # -- Enable IPv6 support. Useful for running in IPv6-only Kubernetes clusters
+  # enableTCP6: true
   # promscrape.maxScrapeSize: "167772160"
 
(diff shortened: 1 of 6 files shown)


charts/victoria-metrics-agent/values.yaml
```diff
@@ -139,7 +139,7 @@ extraArgs:
   envflag.prefix: VM_
   loggerFormat: json
   httpListenAddr: :8429
-  # -- Enable IPv6 support. Useful for running in IPv6-only Kubernetes clusters
```
vrutkovs, line 142: Why change that? Its stylistically consistent with other comments already

vrutkovs (commented):
Why would we want to document this particular flag? I'd rather have 

```
# -- VMAgent extra command line arguments
# -- See https://docs.victoriametrics.com/victoriametrics/vmagent/#common-flags
```

charts/victoria-metrics-agent/values.yaml
AndrewChubatiuk: `# --` is not about style, it's used by helm-docs to generate a doc section for chart value

andriibeee:
@vrutkovs See https://github.com/VictoriaMetrics/VictoriaMetrics/issues/6858 for context. In short, when a VM component is configured with an IPv6 listening address but the enableTCP6 flag isn't set, it silently falls back to IPv4 - which isn't what users on IPv6-only networks expect. This change should reduce friction for those users as this flag is buried deep in the docs rn. I've also opened a PR upstream to log a warning when an IPv6 address is specified without enableTCP6 set: https://github.com/VictoriaMetrics/VictoriaMetrics/pull/10640 so the values.yaml file change is optional if the mentioned pr gets merged but still i see value in it having extra attention in this file.

vrutkovs:
>it silently falls back to IPv4

This is a VictoriaMetrics bug, not operator's one. There is no reason for operator to make fallbacks for every VictoriaMetric bug

>this flag is buried deep in the docs rn

One could argue that every flag is buried deep in the docs. Why not add a link to the docs for general use case?

PR #2806, opened 2026-04-12
all: replace custom app label with app.kubernetes.io/component

AndrewChubatiuk:
fixes https://github.com/VictoriaMetrics/helm-charts/issues/2785

this change as it leads to pods restart

---
## Summary by cubic
Replaced the custom `app` label with `app.kubernetes.io/component` across all charts and selectors (including tests and the Grafana datasource) to follow Kubernetes conventions and avoid surprise rollouts. Also moved chart dependencies to OCI and bumped `victoria-metrics-common` to `0.1.*`.

- **Migration**
  - Expect a rolling restart on upgrade due to pod template label changes.
  - Update any custom selectors (Service/Ingress/HPA/NetworkPolicy/ServiceMonitor/PodMonitor/`VM*` CRs) to use `app.kubernetes.io/component`.
  - Helm 3.8+ is required for OCI dependencies.

- **Dependencies**
  - Switched chart repositories to `oci://ghcr.io/victoriametrics/helm-charts`.
  - Bumped `victoria-metrics-common` to `0.1.*`; `victoria-metrics-k8s-stack` to `0.72.6` (where used).

<sup>Written for commit f8fce351701d863ba752717acefc6a66519b26cd. Summary will update on new commits.</sup>

### DIFF
--- a/charts/victoria-logs-agent/CHANGELOG.md
+++ b/charts/victoria-logs-agent/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Next release
 
-- TODO
+**Update node 1**: due to change in label name pods will be restarted.
+
+- replace custom app label with app.kubernetes.io/component. See [#2785](https://github.com/VictoriaMetrics/helm-charts/issues/2785).
 
 ## 0.0.15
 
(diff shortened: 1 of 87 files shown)


vrutkovs:
Found two more places where we could replace these - @AndrewChubatiuk could you check it?

vrutkovs (approved):

AndrewChubatiuk:
additionally need to add update notes to all charts, which do not use operator

sashasimkin:
gonna crosspost this here too: https://github.com/VictoriaMetrics/helm-charts/issues/2785#issuecomment-4247215422

this change as it is not is breaking for statefulsets - blocking them from upgrading

PR #2877, opened 2026-05-03
fix(victoria-metrics-cluster): make VPA option name match the template

InfraArtists:
The `victoria-metrics-cluster` chart had a name mismatch between `values.yaml` and the VPA template:

- `values.yaml` documented the option as `verticalPodAutoscaling` (for `vmselect`, `vminsert`, `vmauth`, `vmstorage`).
- `templates/vpa.yaml` actually reads `verticalPodAutoscaler` (no `-ing`).

This means if a user followed the documented option and set `verticalPodAutoscaling.enabled: true`, **nothing happened**. The VPA was never created. Only users who guessed the undocumented name got it to work.

---
## Summary by cubic
Fixes the VPA values key mismatch in the `victoria-metrics-cluster` chart so enabling VPA actually creates resources. Renames `verticalPodAutoscaling` to `verticalPodAutoscaler` for `vmselect`, `vminsert`, `vmauth`, and `vmstorage` to match the template.

- **Migration**
  - Rename values: `<component>.verticalPodAutoscaling.*` -> `<component>.verticalPodAutoscaler.*` for the components above.

<sup>Written for commit c7d7a7d513ba152fb88a294fdc95ef3b240b8c8d. Summary will update on new commits.</sup>

### DIFF
--- a/charts/victoria-metrics-cluster/CHANGELOG.md
+++ b/charts/victoria-metrics-cluster/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next release
 
-- TODO
+- **BREAKING**: rename `verticalPodAutoscaling` to `verticalPodAutoscaler` in `vmselect`, `vminsert`, `vmauth` and `vmstorage` to match the template, which previously read `verticalPodAutoscaler` and silently ignored the documented field. Users who set `<component>.verticalPodAutoscaling.*` must rename it to `<component>.verticalPodAutoscaler.*`.
 
 ## 0.41.2
 
--- a/charts/victoria-metrics-cluster/values.yaml
+++ b/charts/victoria-metrics-cluster/values.yaml
@@ -136,7 +136,7 @@ vmselect:
   # -- Vertical Pod Autoscaling.
   # Requires VPA CRD (`autoscaling.k8s.io/v1`) to be installed in the cluster.
   # Note that VPA should not be used together with HPA on the same resource metrics (CPU/memory).
-  verticalPodAutoscaling:
+  verticalPodAutoscaler:
     # -- Use VPA for vmagent
     enabled: false
     # -- List of custom recommenders to use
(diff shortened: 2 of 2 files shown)


AndrewChubatiuk:
thanks for spotting this
fixed it in all charts in [this commit](https://github.com/VictoriaMetrics/helm-charts/commit/3e03c2ec9021f788beb813da9f570bd419d744f4)

InfraArtists:
@AndrewChubatiuk Thanks for the quick fix! However, since I spent time diagnosing this bug and submitted a PR, I would have appreciated the chance to update my PR to cover all charts or at least be credited via a co-authored commit. It’s always encouraging for contributors to see their PRs merged rather than closed and re-implemented.

@zekker6 @vrutkovs   I spent several hours diagnosing this bug and provided the initial fix in this PR. Closing a contributor's PR only to re-implement it under your own name is known as **Steamrolling** in the open-source community.
It's a discouraging practice that goes against the core philosophy of open-source collaboration.

While the bug is fixed, the way it was handled ignores the effort of outside contributors. I hope in the future, we can stick to standard maintainer etiquette by providing feedback on existing PRs rather than overriding them.

vrutkovs:
We are very sorry that it happened. The commit we've merged was slightly larger in scope and avoided the breaking change, so we've preferred this version.

Unfortunately, we cannot modify the existing commit to list you as a co-author, as the master branch is protected. Once again, sorry about this, and I hope it won't discourage you from contributing here further.

InfraArtists:
@vrutkovs I appreciate the apology and the explanation. However, let’s remember that Open Source is about Community, not just Commits.
Tools and charts come and go; they all have a lifecycle and will eventually be replaced. What remains is the culture we build and how we treat those who dedicate their time to help.

Technical skills are common, but understanding the philosophy of collaboration is what defines a true Maintainer.

@AndrewChubatiuk, I suggest looking up to mentors like @vrutkovs  as role models to understand that Open Source is a philosophy of collaboration, not just a race to commit code. There is a profound difference between being a professional maintainer and being just a 'typist' who ignores the human effort behind a contribution.

I hope this experience serves as a lesson in respecting the alphabet of Open Source etiquette for future interactions.

InfraArtists:
On a final note, my company was evaluating VictoriaMetrics for an Enterprise support contract. However, after witnessing this level of technical gatekeeping and a dismissive attitude towards critical security fixes, culture, and collaboration, we have decided to halt those plans. (Because of #2879 and this PR)

We value stability and professional partnership, so we’ve decided to begin our migration back to Prometheus/Cortex. We’d rather invest in an ecosystem that values its contributors and users alike.

Decisions have consequences.

FYI: @valyala

PR #2879, opened 2026-05-04
fix(victoria-metrics-distributed): close anonymous vmauth access by default

InfraArtists:
## Summary

The `victoria-metrics-distributed` chart used to expose its vmauth proxies to **anyone with no password** by default.

When you installed the chart out-of-the-box, these paths were open to anonymous traffic:

- **Write paths** (any tenant): `/api/v1/write`, `/insert/<tenant>/...`, `/opentelemetry/...`, etc.
- **Read paths** (any tenant): `/select/...`, `/admin/tenants`

This means anyone who could reach the vmauth Service could read or write metrics for **any tenant**. That broke the whole point of multi-tenancy.

## What this PR does

Adds a new option called `allowUnauthenticated` to three places:

- `write.global.vmauth.allowUnauthenticated`
- `read.global.vmauth.allowUnauthenticated`
- `zoneTpl.read.vmauth.allowUnauthenticated`

The default value is `false`. This means:

1. **By default**, the chart will **not** open the vmauth to anonymous traffic.
2. If you don't configure any users, the chart will **stop with a clear error message** telling you what to do.
3. If you really want the old behavior (anonymous access), you can set `allowUnauthenticated: true`. The chart will then add the same routing rules as before.
4. If you want proper auth, you set `spec.users` (or `spec.unauthorizedAccessConfig`) and the chart will use that.

## Breaking change

Yes. Existing installs that relied on the silent anonymous proxy will fail to render after upgrade. The fix is one line in your `values.yaml`:

```yaml
# Option A: keep the old (insecure) behavior
write:
  global:
    vmauth:
      allowUnauthenticated: true
read:
  global:
    vmauth:
      allowUnauthenticated: true
zoneTpl:
  read:
    vmauth:
      allowUnauthenticated: true

# Option B (recommended): add real users
read:
  global:
    vmauth:
      spec:
        users:
          - username: reader
            password: <secret>

<!-- This is an auto-generated description by cubic. -->
---
## Summary by cubic
Locks down `vmauth` in `victoria-metrics-distributed` by default. Adds an `allowUnauthenticated` toggle and expands the template-time guard so charts fail fast when no auth is configured. This is breaking for clusters that relied on anonymous access.

- **New Features**
  - New `allowUnauthenticated` (default `false`) in `write.global.vmauth`, `read.global.vmauth`, and `zoneTpl.read.vmauth`.
  - When `false`, templating now fails unless one of these is set: `spec.users`, `spec.userSelector`/`spec.userNamespaceSelector`, `spec.selectAllByDefault`, `spec.configSecret`, `spec.unauthorizedUserAccessSpec`, or `spec.unauthorizedAccessConfig`.
  - Removed implicit `unauthorizedUserAccessSpec.url_map` defaults from `read.global.vmauth.spec` and `zoneTpl.read.vmauth.spec` to avoid open proxies.

- **Migration**
  - Keep old behavior (insecure): set `allowUnauthenticated: true` in the three vmauth blocks.
  - Recommended: configure auth via `spec.users`, or use selectors/secrets (`spec.userSelector`, `spec.userNamespaceSelector`, `spec.selectAllByDefault`, `spec.configSecret`).

<sup>Written for commit 1e016d1b266ccfaf418edcf927ff2b1c61acd211. Summary will update on new commits.</sup>

<!-- End of auto-generated description by cubic. -->

### DIFF
--- a/charts/victoria-metrics-cluster/CHANGELOG.md
+++ b/charts/victoria-metrics-cluster/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next release
 
-- TODO
+- **BREAKING**: rename `verticalPodAutoscaling` to `verticalPodAutoscaler` in `vmselect`, `vminsert`, `vmauth` and `vmstorage` to match the template, which previously read `verticalPodAutoscaler` and silently ignored the documented field. Users who set `<component>.verticalPodAutoscaling.*` must rename it to `<component>.verticalPodAutoscaler.*`.
 
 ## 0.41.2
 
--- a/charts/victoria-metrics-cluster/values.yaml
+++ b/charts/victoria-metrics-cluster/values.yaml
@@ -136,7 +136,7 @@ vmselect:
   # -- Vertical Pod Autoscaling.
   # Requires VPA CRD (`autoscaling.k8s.io/v1`) to be installed in the cluster.
   # Note that VPA should not be used together with HPA on the same resource metrics (CPU/memory).
-  verticalPodAutoscaling:
+  verticalPodAutoscaler:
     # -- Use VPA for vmagent
     enabled: false
     # -- List of custom recommenders to use
@@ -507,7 +507,7 @@ vminsert:
   # -- Vertical Pod Autoscaling.
   # Requires VPA CRD (`autoscaling.k8s.io/v1`) to be installed in the cluster.
   # Note that VPA should not be used together with HPA on the same resource metrics (CPU/memory).
-  verticalPodAutoscaling:
+  verticalPodAutoscaler:
     # -- Use VPA for vmagent
     enabled: false
     # -- List of custom recommenders to use
@@ -826,7 +826,7 @@ vmauth:
   # -- Vertical Pod Autoscaling.
   # Requires VPA CRD (`autoscaling.k8s.io/v1`) to be installed in the cluster.
   # Note that VPA should not be used together with HPA on the same resource metrics (CPU/memory).
-  verticalPodAutoscaling:
+  verticalPodAutoscaler:
     # -- Use VPA for vmagent
     enabled: false
     # -- List of custom recommenders to use
@@ -1285,7 +1285,7 @@ vmstorage:
   # -- Vertical Pod Autoscaling.
   # Requires VPA CRD (`autoscaling.k8s.io/v1`) to be installed in the cluster.
   # Note that VPA should not be used together with HPA on the same resource metrics (CPU/memory).
-  verticalPodAutoscaling:
+  verticalPodAutoscaler:
     # -- Use VPA for vmagent
     enabled: false
     # -- List of custom recommenders to use
--- a/charts/victoria-metrics-distributed/CHANGELOG.md
+++ b/charts/victoria-metrics-distributed/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next release
 
-- TODO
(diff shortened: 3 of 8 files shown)


AndrewChubatiuk:
distributed chart most likely will become deprecated soon. k8s-stack includes vmdistributed resource as a replacement. distributed resource starting operator version 0.69.0 has no default auth config, it's user's responsibility to configure it. check [these docs](https://docs.victoriametrics.com/operator/resources/vmdistributed/#authorization)

InfraArtists:
I’m submitting this fix, but honestly, after what happened with #2877, I was hesitant to spend more time on it.

I've implemented the fix here, but if the plan is to once again close my PR and re-implement it under a maintainer's name to avoid giving contribution credit, please let me know now.
I'd rather not waste time on the PR process if the project's policy is to 'Steamroll' outside contributors instead of following standard OSS etiquette.

CC: @vrutkovs @zekker6 (for visibility on the process)

AndrewChubatiuk:
only bugfixes are expected for victoria-metrics-distributed chart. created #2880 which adds a deprecation note to this chart. consider using k8s-stack with vmdistributed.enabled if you're interested in distributed cluster setup. closing this PR

InfraArtists:
@AndrewChubatiuk 

A security vulnerability that exposes private data by default is still a critical issue, whether the chart is nearing deprecation or not. Closing a functional security fix (especially after the 'Steamrolling' incident in #2877) suggests that this project is more interested in gatekeeping than in the security and contributions of its community.

If the goal is to discourage outside contributors and ignore security flaws, you are doing a great job. I’ll be taking my future security findings and contributions to projects that actually value their community and follow basic OSS ethics.

Good luck with the manual deprecation :)

vrutkovs:
>That broke the whole point of multi-tenancy.

This is orthogonal to the distributed chart goals.  Unlike the operator implementation, helm chart is left loose enough for users to modify and adapt to their needs. As a result, adding more features to the chart increases the maintenance load on us to carry those changes forward. Moreover, the distributed helm chart will never match the operator solution in terms of operational stability and convenience (which is why we're discussing deprecating the chart).

I'd prefer to create a feature request first before rushing with the implementation. For instance, this particular problem is already solved in the operator in an entirely different fashion. Why would we want to offer two different solutions to our users? Moreover, we need to make sure users can migrate from helm chart to the operator - so we need to make sure the similar solution is also implemented in the operator.

>A security vulnerability that exposes private data by default is still a critical issue

Please don't report security problems publicly - the correct way to do so is coordinating disclosure with us over at security@victoriametrics.com.

InfraArtists:
@vrutkovs, I appreciate the technical feedback, but let's be clear.

Regarding your point on disclosure: I didn't just 'report' a vulnerability; I provided a fix through the standard PR process.

It’s ironic to talk about 'operational stability' and 'maintenance load' as excuses to leave a critical security flaw unpatched. Closing security fix only 4 hours after it was submitted shows a lack of urgency for user safety.

Rejecting a solution for a known vulnerability based on 'future deprecation' is a poor trade-off for current users.

If the project's priority was truly security, the fix would have been reviewed or refined, not dismissed with a 'laugh' emoji (and after that, changing to the 'heart' emoji)

Whether the chart is nearing deprecation or not is irrelevant to its current security status. As long as the code is recommended for use, it should be secure. Telling users to migrate to the Operator is a long-term solution, but it doesn't fix the immediate risk for existing Helm users.

From now on, whenever I find more bugs or security flaws in your projects, I’ll continue to document them and provide the solutions publicly in the Issues section.

Since some members of your team seem so desperate for contribution credit that they resort to closing external PRs just to re-submit them as their own, I’m happy to provide the 'inspiration' for free. Consider it my contribution to the professional development of those who prioritize their GitHub stats over open-source ethics.

Good luck.

PR #2887, opened 2026-05-10
k8s-stack: use sync-job for dashboards and rules to decrease a chart size

AndrewChubatiuk:
One of the obstacles to adding VLSingle/VLCluster support to k8s-stack is the chart size. It's already quite large due to dashboards and rules templates. This PR introduces a sync-job that converts rules and dashboards from defined sources and creates VMRule, ConfigMap (or GrafanaDashboard). Additionally, this change should allow us not to limit users to a predefined list of dashboards and convert additional ones in the same manner. This approach also has some disadvantages:
- dashboards and rules cannot be installed on air-gapped envs. If it's needed rules and dashboards can be added to a docker image
- probably more cases should be covered later for other dashboards: limit grafana version, using which dashboards were generated, limit supported datasource types, etc.


AndrewChubatiuk:
@cubic-dev-ai

AndrewChubatiuk:
@vrutkovs added a workflow to publish sync-job, already published one and chart is ready to be checked without a need to build anything

charts/victoria-metrics-k8s-stack/values.yaml
```diff
@@ -42,41 +42,139 @@ victoria-metrics-operator:
...
+    node-exporter-full:
+      url: https://raw.githubusercontent.com/rfmoz/grafana-dashboards/master/prometheus/node-exporter-full.json
+      enabled: '{{ index .Values "prometheus-node-exporter" "enabled" }}'
+    etcd:
+      url: https://raw.githubusercontent.com/monitoring-mixins/website/master/assets/etcd/dashboards/etcd.json
+      enabled: '{{ .Values.kubeEtcd.enabled }}'
+    kube-prometheus:
+      url: https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/main/manifests/grafana-dashboardDefinitions.yaml
+
+syncJob:
```
vrutkovs, line 132: Let's nest this under `defaultDashboard.dashboards`?

AndrewChubatiuk: this is used for both rules and dashboards

hack/sync-job/dashboards.go
```diff
@@ -0,0 +1,497 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/VictoriaMetrics/metricsql"
+	"gopkg.in/yaml.v3"
+)
+
+var dashboardNameRe = regexp.MustCompile(`[ /_-]+`)
+
+// --- JSON types (unknown fields preserved via GOEXPERIMENT=jsonv2) ---
```
vrutkovs, line 18: I'm a little hesitant to rely on experimental features. I wonder if we have a contingency plan in case it is dropped - revert this PR, stick to particular golang etc?

AndrewChubatiuk: json v2 is expected in 1.27 at least when i checked it last time

AndrewChubatiuk: there's a repo which has this dependency as stable, it can be installed without experimental flag

vrutkovs (approved):

rpardini:
Seems this did away with `defaultDashboards.defaultTimezone`? (It's fine; handling that through Grafana's own mechanisms is okay).

AndrewChubatiuk:
right, it's mentioned in a changelog

PR #2892, opened 2026-05-12
fix(victoria-metrics-k8s-stack): add uid to default VictoriaMetrics datasource

eleboucher:
Set an explicit `uid: VictoriaMetrics` on the default Prometheus-type datasource in `defaultDatasources.victoriametrics.datasources`. Some dashboards reference the datasource by UID rather than name, and without a stable UID Grafana auto-generates one, breaking those dashboards.

---
## Summary by cubic
Set explicit UIDs on default Grafana datasources for stable identifiers and to prevent UID-based dashboards from breaking.

Adds `uid: VictoriaMetrics` to the default Prometheus datasource in both `victoria-metrics-k8s-stack` and `victoria-metrics-distributed`, plus `uid: VictoriaMetricsDS` for the `victoriametrics-metrics-datasource` and `uid: Alertmanager` for the Alertmanager datasource.

<sup>Written for commit a4c36ffa990e842345f2ab17a2e516616221b148. Summary will update on new commits.</sup>

### DIFF
--- a/charts/victoria-metrics-distributed/templates/grafana-datasource.yaml
+++ b/charts/victoria-metrics-distributed/templates/grafana-datasource.yaml
@@ -23,6 +23,7 @@ data:
       type: prometheus
       url: {{ $url }}
       access: proxy
+      uid: VictoriaMetrics
       isDefault: true
       jsonData: {}
 {{- end }}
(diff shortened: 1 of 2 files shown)


vrutkovs:
Thank you, this is a useful idea to implement. Could you add it to other datasources as well:
* "VictoriaMetrics (DS)" and "Alertmanager" in k8s stack
* "VictoriaMetrics" in distributed chart?

vrutkovs (approved):

vrutkovs:
Thank you for your contribution!

PR #2893, opened 2026-05-12
chore: update tested k8s releases

vrutkovs:
Update tested k8s versions to match the latest supported


AndrewChubatiuk (approved):

vrutkovs:
nevermind, we use kind images, which don't updated too often. Also, Renovate will create PRs once these are available

PR #2903, opened 2026-05-18
fix: use fileMatch in customManagers for custom.regex

vrutkovs:
Use fileMatch instead of managerFilePatterns to match all charts. Also cleans up missing Dockerfiles

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -46,21 +46,18 @@
     {
       "description": "Disable updates for victoriametrics images tracked via appVersion",
       "matchManagers": ["helm-values"],
-      "matchFileNames": ["*/victoria-metrics-*/Dockerfile"],
       "matchPackagePatterns": ["^victoriametrics/vm"],
       "enabled": false
     },
     {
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -19,15 +19,15 @@
   "customManagers": [
     {
       "customType": "regex",
-      "managerFilePatterns": ["charts/[^/]+/Chart\\.yaml$"],
+      "fileMatch": ["charts/[^/]+/Chart\\.yaml$"],
```
AndrewChubatiuk:
```
managerFilePatterns formerly known as fileMatch, which supported regex-only.

managerFilePatterns supports both regex and glob patterns. Any existing config containing fileMatch patterns will be automatically migrated.
```

vrutkovs: Oh, I misunderstood what https://github.com/renovatebot/renovate/discussions/36287#discussioncomment-13354071 says.

AndrewChubatiuk (approved):

PR #2927, opened 2026-05-23
victorialogs: allow overriding the container command

mattsre:
VictoriaLogs charts should allow overriding the command for the server so that they can be used with container images which may not have the VictoriaLogs binary as the default entrypoint

---
## Summary by cubic
Adds support to override the container command across VictoriaLogs Helm charts. This lets you run images where the VictoriaLogs binary isn’t the default entrypoint.

- **New Features**
  - New optional `command` values:
    - `victoria-logs-single`: `server.command`
    - `victoria-logs-cluster`: `vlinsert.command`, `vlselect.command`, `vlstorage.command`
    - `victoria-logs-multilevel`: `vlselect.command`
  - Templates render `command` only when set; defaults remain unchanged.

- **Migration**
  - To use a custom binary path, set the relevant value, e.g. `server.command: ["/custom/bin/victorialogs"]`.

<sup>Written for commit d4c02dc640ed993755e8019eb13371dee98a469a. Summary will update on new commits. <a href="https://cubic.dev/pr/VictoriaMetrics/helm-charts/pull/2927?utm_source=github">Review in cubic</a></sup>

PR #2928, opened 2026-05-25
feat: add ability to override container command

vrutkovs:
Users may want to run custom images or run different commands, the charts should allow overriding the entrypoint.

This is a followup for https://github.com/VictoriaMetrics/helm-charts/pull/2927, covering remaining charts

### DIFF
--- a/charts/victoria-metrics-agent/CHANGELOG.md
+++ b/charts/victoria-metrics-agent/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next release
 
-- TODO
+- add ability to override container command.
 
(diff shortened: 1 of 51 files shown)


AndrewChubatiuk:
Afraid with this ability people may also extend their custom entry points to support subcommands and other things, that are not covered by our templates

vrutkovs:
Certainly, it opens a way for users to prevent the cluster from starting, but the fix is simple - remove the custom command - and it won't corrupt the data doing so

AndrewChubatiuk (approved):

PR #2934, opened 2026-05-27
.github: enforce least priviledge permissions for action jobs

rudransh-shrivastava:
Scope permissions to jobs and set `permissions: {}` for workflow level.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -16,11 +16,12 @@ on:
       - '**/_index.md.gotmpl'
       - '.github/workflows/docs.yaml'
 
+permissions: {}
+
 jobs:
   publish-readme:
     permissions:
       contents: write
-      id-token: write
 
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -13,10 +13,12 @@ env:
   HELM_VERSION: 4.1.3
   YQ_VERSION: 4.44.2
 
+permissions: {}
+
 jobs:
   release:
     permissions:
-      contents: write
+      contents: read
       packages: write
       id-token: write
     runs-on: ubuntu-latest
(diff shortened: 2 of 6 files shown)


.github/workflows/docs.yaml
```diff
@@ -16,11 +16,12 @@ on:
       - '**/_index.md.gotmpl'
       - '.github/workflows/docs.yaml'
 
+permissions: {}
+
 jobs:
   publish-readme:
     permissions:
-      contents: write
-      id-token: write
```
vrutkovs, line 23: I think we need this one for the push? cc @AndrewChubatiuk

AndrewChubatiuk: suppose it's fine, to clone other repos in a pipeline we're using user's credentials, which are stored in secrets

rudransh-shrivastava: `Push to vmdocs` step should work as checkout step uses `secrets.VM_BOT_GH_TOKEN` to authenticate.

I had assumed the `env` variables would be enough but it seems only `gh` cli supports those:
```yaml
        env:
          GH_TOKEN: "${{ secrets.VM_BOT_GH_TOKEN }}"
          GITHUB_TOKEN: "${{ secrets.VM_BOT_GH_TOKEN }}"
```

`Synchronize docs` might fail, I'll just add back write permissions to be safe.

rudransh-shrivastava: fixed.

AndrewChubatiuk (approved):

.github/workflows/release.yaml
```diff
@@ -13,10 +13,12 @@ env:
   HELM_VERSION: 4.1.3
   YQ_VERSION: 4.44.2
 
+permissions: {}
+
 jobs:
   release:
     permissions:
-      contents: write
+      contents: read
```
AndrewChubatiuk: this job's token is used to publish chart to ghcr, write permissions are needed

AndrewChubatiuk:
```suggestion
      contents: write
```

PR #2999, opened 2026-06-15
Updates to vmsingle and vmcluster charts related to vmbackupmanager

brentteal:
during onboarding found that it would be useful to be able to mount an extra volume in tje vmbackupmanager sidecar for local filesystem backups.  This was easily done and straight forward.

This chnange would allow a customer to be able to setup a shared pvc in k8's that all vmstorage nodes could use as a backup location.  This shared pvc could use nfs to seperate the backups from the storage environment.

Once i looked at the vmsingle chart, i found that the extraVolumeMount existed in the vmbackupmanager sidecar container however it was not setup in the vmbackupmanager restore container that is created when restore on start is enabled.  

During testing found that the init restore container failed to start due to a misconfigured security context and misnamed storage volume. 

This is in draft so i can have a few folks have a look first.

### DIFF
--- a/charts/victoria-metrics-cluster/values.yaml
+++ b/charts/victoria-metrics-cluster/values.yaml
@@ -1454,9 +1454,8 @@ vmstorage:
       #  subPath: ""
       #  readOnly: true
 
-    #  Extra Volume Mounts for the container
-    extraVolumeMounts:
-      []
+    # --  Extra Volume Mounts for the container
+    extraVolumeMounts: []
       # - name: example
       #   mountPath: /example
   serviceMonitor:
--- a/charts/victoria-metrics-cluster/values.yaml
+++ b/charts/victoria-metrics-cluster/values.yaml
@@ -1453,6 +1453,12 @@ vmstorage:
       #  mountPath: /etc/credentials
       #  subPath: ""
       #  readOnly: true
+
+    #  Extra Volume Mounts for the container
+    extraVolumeMounts:
+      []
+      # - name: example
+      #   mountPath: /example
   serviceMonitor:
     # -- Enable deployment of Service Monitor for vmstorage component. This is Prometheus operator object
     enabled: false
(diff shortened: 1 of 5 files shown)


charts/victoria-metrics-cluster/values.yaml
```diff
@@ -1453,6 +1453,12 @@ vmstorage:
       #  mountPath: /etc/credentials
       #  subPath: ""
       #  readOnly: true
+
+    #  Extra Volume Mounts for the container
+    extraVolumeMounts:
+      []
```
vrutkovs:
```suggestion
    # --  Extra Volume Mounts for the container
    extraVolumeMounts: []
```

`--` is important here to make `helm-docs` autogenerate docs from these `values.yaml`. Moved empty array to the same line for consistency

charts/victoria-metrics-single/templates/server.yaml
```diff
@@ -65,8 +65,8 @@ spec:
         {{- $_ := set $ctx "appKey" (list "server" "vmbackupmanager") }}
         image: {{ include "vm.image" $ctx }}
         imagePullPolicy: {{ $app.image.pullPolicy }}
-        {{- with $app.podSecurityContext }}
-        securityContext:  {{ toYaml . | nindent 10 }}
+        {{- if $app.securityContext.enabled }}
+        securityContext: {{ include "vm.securityContext" (dict "securityContext" $app.securityContext "helm" .) | nindent 10 }}
```
vrutkovs, line 69: Good catch!

charts/victoria-metrics-single/templates/server.yaml
```diff
@@ -84,14 +84,17 @@ spec:
         - name: manager-http
           containerPort: 8300
         volumeMounts:
-        - name: vmstorage-volume
+        - name: server-volume
```
vrutkovs:
```suggestion
        - name: {{ $storageName }}
```
in case this volume name has an override

vrutkovs (commented):
Very good, I think this almost good to go

vrutkovs:
Looks good, needs CHANGELOG update

vrutkovs (approved):

AndrewChubatiuk (approved):

AndrewChubatiuk:
@brentteal could you please sign commits?

PR #3010, opened 2026-06-18
update chart to remove targetPort from values and set container port name to always be http

brentteal:
vmanomaly no longer needs the "monitoring" or "server" setting in the config to enable listening on port 8490.   Originally one would need to set either one or the other for the containerPort to be created  in server.yaml.  As the container listens by default it is no longer necessary to create this config, and also harmonize the expected targetport name in the service definition of the values file. 

If a user were to add a monitoring config, instead of a servier / port config, and keep the default targetport name of http the container port name would be set to "metrics" and the service would be targetting a port named "http"

This update resolves this by setting the service to always use targetport http and the pod's container port is always named http.  The targetport value in the values file has also been remvoved.  The service port and node port values still remain.

### DIFF
--- a/charts/victoria-metrics-anomaly/templates/server.yaml
+++ b/charts/victoria-metrics-anomaly/templates/server.yaml
@@ -56,17 +56,9 @@ spec:
         - name: model
           image: {{ include "vm.image" $ctx }}
           workingDir: {{ .Values.containerWorkingDir }}
-          {{- if or (((.Values.config).monitoring).pull).enabled ((.Values.config).server).port }}
           ports:
-            {{- with (((.Values.config).monitoring).pull).port }}
-            - containerPort: {{ . }}
-              name: metrics
-            {{- end }}
-            {{- with ((.Values.config).server).port }}
-            - containerPort: {{ . }}
+            - containerPort: 8490
               name: http
-            {{- end }}
-          {{- end }}
           {{- with .Values.command }}
           command: {{ toYaml . | nindent 12 }}
           {{- end }}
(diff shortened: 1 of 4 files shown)


AndrewChubatiuk:
let's add extraContainers support instead, which can be used for oauth proxy for example. in this case targetPort makes sense.

charts/victoria-metrics-anomaly/values.yaml
```diff
@@ -301,7 +301,6 @@ service:
   # -- Service port
   servicePort: 8490
   # -- Target port
```
vrutkovs: Remove this comment also so that helm-docs would not be confused

charts/victoria-metrics-anomaly/templates/server.yaml
```diff
@@ -56,17 +56,9 @@ spec:
...
           workingDir: {{ .Values.containerWorkingDir }}
-          {{- if or (((.Values.config).monitoring).pull).enabled ((.Values.config).server).port }}
           ports:
-            {{- with (((.Values.config).monitoring).pull).port }}
-            - containerPort: {{ . }}
-              name: metrics
-            {{- end }}
-            {{- with ((.Values.config).server).port }}
-            - containerPort: {{ . }}
+            - containerPort: 8490
```
vrutkovs: imo we should still use `((.Values.config).server).port | default 8490` here, in case this is a custom image

brentteal: OKay done.

charts/victoria-metrics-anomaly/values.yaml
brentteal: oops removed.

vrutkovs (approved):

PR #3020, opened 2026-06-23
fix: make sync job a helm pre hook

wittdennis:
The current implementation is a simple batch job with the config map hash in the name. This causes GitOps systems like argo-cd to always show an OutOfSync error. By making the Job a helm hook this is avoided

### DIFF
--- a/charts/victoria-metrics-k8s-stack/templates/sync-job/job.yaml
+++ b/charts/victoria-metrics-k8s-stack/templates/sync-job/job.yaml
@@ -3,15 +3,16 @@
 {{- $fullname := include "vm.fullname" $ctx }}
 {{- $ns := include "vm.namespace" $ctx }}
 {{- $job := .Values.syncJob }}
-
-{{- $configHash := include (print $.Template.BasePath "/sync-job/config.yaml") . | sha256sum | trunc 8 }}
 ---
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ $fullname }}-sync-job-{{ $configHash }}
+  name: {{ $fullname }}-sync-job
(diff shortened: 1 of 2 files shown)


charts/victoria-metrics-k8s-stack/templates/sync-job/job.yaml
```diff
@@ -3,15 +3,17 @@
...
 kind: Job
 metadata:
-  name: {{ $fullname }}-sync-job-{{ $configHash }}
+  name: {{ $fullname }}-sync-job
   namespace: {{ $ns }}
   labels: {{ include "vm.labels" $ctx | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
+    "helm.sh/hook-weight": "-5"
+    "helm.sh/hook-delete-policy": before-hook-creation,hook-succeeded
```
AndrewChubatiuk: it's better to use post-install,post-upgrade hook as operator chart (which is k8s-stack dependency) with `crds.plain: false` installs CRDs as templates, which are applied after pre-install hook and hook will fail since VMRule CRD is not yet present.

i think hash removal and setting only job annotations should be enough, annotations for other resources are not needed

```suggestion
  annotations:
    helm.sh/hook: post-install,post-upgrade
    helm.sh/hook-delete-policy: before-hook-creation
```

wittdennis: Thanks for the clarification! I've updated my PR.

AndrewChubatiuk (approved):

AndrewChubatiuk:

PR #3023, opened 2026-06-23
test: remove bloated terratest

AndrewChubatiuk:
remove bloated terratest dependency from tests


vrutkovs:
I'm not excited to have our own small terratest implementation, what exactly are the benefits - smaller OCI image?

AndrewChubatiuk:
we do not build any OCI. i don't want to see a huge diff after packages update or items in github security page, which are completely irrelevant. i have the same question about terratest, what it actually brings besides some tiny wrappers, which we can implement ourselves and bunch of unneeded dependencies?

vrutkovs:
Oh, okay, secutiy footprint is a good point. I extensively use terratest in e2e tests, so I wonder if there are slimmer alternatives (perhaps its worth maintaining our own without the pesky aws S3 dependency?)

Agreed on the direction, I'll have a closer look on the code bit later

vrutkovs (approved):

PR #3030, opened 2026-06-24
add useLegacyNaming flag to switch between legacy (current chart naming) and operator naming

AndrewChubatiuk:
added `useLegacyNaming` flag (default: true), which allows to define naming convention for plain k8s resources, that are managed by helm charts that do not contain operator. It allows for new users to have resources with naming convention, which is compatible with operator (`useLegacyNaming: false`)

### DIFF
--- a/charts/victoria-logs-agent/values.yaml
+++ b/charts/victoria-logs-agent/values.yaml
@@ -2,6 +2,10 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
+# -- Use legacy naming convention (release-name based). When false, resource names
+# -- use the same convention as the operator (type-prefixed, e.g. vlagent-<release>).
+useLegacyNaming: true
+
 # Important helm-chart settings below, modify them with your own settings
 
 # -- Maximum disk usage per remoteWrite URL. When exceeded, old logs will be deleted.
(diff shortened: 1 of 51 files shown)


vrutkovs:
`"vm.plain.fullname"` seems to conflict with `fullnameOverride` with `useLegacyNaming=false`? We also have `vm.fullname` and `vm.managed.fullname` - do we need to update those too?

AndrewChubatiuk:
> "vm.plain.fullname" seems to conflict with fullnameOverride with useLegacyNaming=false? We also have vm.fullname and vm.managed.fullname - do we need to update those too?

good point! updated `vm.plain.fullname` and `vm.managed.fullname`, for `vm.fullname` it's not needed

vrutkovs:
We still use `vm.fullname` in vlagent and vmanomaly charts - and these won't be affected by `useLegacyNaming` at all

AndrewChubatiuk:
> We still use vm.fullname in vlagent and vmanomaly charts - and these won't be affected by useLegacyNaming at all

now i'm thinking about keeping this useLegacyNaming for single and cluster instances only and migrate the rest to new convention

vrutkovs:
Wouldn't that affect service names too, so users would have to reconfigure vmagent addresses? Perhaps we should take objects one by one and carefully check that upgrade passes as expected

AndrewChubatiuk:
agree, probably also makes sense to add respective CR into each plain chart, it should cover one of existing issues

PR #3036, opened 2026-06-24
fix: use stable labels on volumeClaimTemplates on victoria-logs-agent

### LINKED ISSUE #3035: bug: victoria-logs-agent chart cannot be upgraded when using persistent volumes
**Chart name and version**
chart: victoria-logs-agent
version: 0.2.7

**Describe the bug**

When `persistentVolume.enabled` is `true`, the StatefulSet's `volumeClaimTemplates` metadata is rendered with the full `vm.labels` helper, which includes version-dependent labels such as `app.kubernetes.io/version` (derived from `Chart.AppVersion` / image tag) and `helm.sh/chart` (includes the chart version).

Kubernetes treats `volumeClaimTemplates` metadata as immutable on an existing StatefulSet. On chart upgrade (e.g. bumping `appVersion` to ship a new vlagent image), Helm attempts to update these labels and the upgrade fails with a forbidden-field error, even when no other StatefulSet spec changes are required.

**Steps to reproduce**

1. Install the chart with dynamic PVCs enabled:

   ```bash
   helm install vlagent victoria-logs-agent \
     --set persistentVolume.enabled=true \
     --set 'remoteWrite[0].url=http://victoria-logs:9428'
   ```

2. Verify the StatefulSet is running and note the PVC template labels, e.g.:

   ```bash
   kubectl get statefulset -l app.kubernetes.io/name=victoria-logs-agent -o yaml | grep -A 20 volumeClaimTemplates
   ```

   Observe labels such as `app.kubernetes.io/version: v1.51.0` and `helm.sh/chart: victoria-logs-agent-0.2.7`.

3. Upgrade to a newer chart version where `appVersion` (and/or chart `version`) has changed:

   ```bash
   helm upgrade vlagent victoria-logs-agent \
     --set persistentVolume.enabled=true \
     --set 'remoteWrite[0].url=http://victoria-logs:9428'
   ```

4. The upgrade fails because Kubernetes rejects changes to `volumeClaimTemplates` metadata.

**Expected behavior**

Chart upgrades should succeed when only the application image version (or chart version) changes. Version labels may change on the StatefulSet and pod template, but should not be applied to `volumeClaimTemplates` metadata, since those fields are immutable after creation.

**Actual behavior**

Helm upgrade fails with an error similar to:

```text
StatefulSet.apps "vlagent-victoria-logs-agent" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'ordinals', 'template', 'updateStrategy', 'revisionHistoryLimit', 'persistentVolumeClaimRetentionPolicy' and 'minReadySeconds' are forbidden
```

**Root cause**

In `templates/server.yaml`, the PVC template uses `include "vm.labels"`:

```yaml
volumeClaimTemplates:
  - metadata:
      labels: {{ include "vm.labels" $ctx | nindent 10 }}
```

The `vm.labels` helper (from `victoria-metrics-common`) includes mutable version metadata:

- `app.kubernetes.io/version` — set from the image tag, defaulting to `Chart.AppVersion`
- `helm.sh/chart` — includes the chart version string

Other VictoriaMetrics charts (e.g. `victoria-metrics-cluster`) avoid this by applying only `persistentVolume.extraLabels` on `volumeClaimTemplates`, not the full `vm.labels` set.

**Suggested fix**

On `volumeClaimTemplates` metadata, use stable labels only — for example `vm.selectorLabels` plus `persistentVolume.extraLabels` — and omit version/chart labels such as `app.kubernetes.io/version` and `helm.sh/chart`.

**Custom values**

Please provide only custom values (excluding default ones):

```yaml
persistentVolume:
  enabled: true

remoteWrite:
  - url: http://victoria-logs:9428
```

**Environment**

- Kubernetes: >= 1.25
- Helm: >= 3.14.0
- Chart: victoria-logs-agent 0.2.7
- Condition: existing StatefulSet with `volumeClaimTemplates` from a prior release

samuel-form3:
volumeClaimTemplates metadata is immutable after StatefulSet creation. Applying vm.labels there included app.kubernetes.io/version and helm.sh/chart, which change on every chart/app version bump and caused helm upgrade to fail when persistentVolume.enabled is true.

Fixes: https://github.com/VictoriaMetrics/helm-charts/issues/3035

### DIFF
--- a/charts/victoria-logs-agent/templates/server.yaml
+++ b/charts/victoria-logs-agent/templates/server.yaml
@@ -117,11 +117,11 @@ spec:
       kind: PersistentVolumeClaim
       metadata:
         name: vl-agent-data
-        {{- $_ := set $ctx "extraLabels" $pvc.extraLabels }}
-        labels: {{ include "vm.labels" $ctx | nindent 10 }}
-        {{- $_ := unset $ctx "extraLabels" }}
         {{- with $pvc.annotations  }}
         annotations: {{ toYaml . | nindent 10 }}
+        {{- end }}
+        {{- with $pvc.extraLabels }}
+        labels: {{ toYaml . | nindent 10 }}
         {{- end }}
       spec:
         {{- with $pvc.accessModes }}
--- a/charts/victoria-logs-agent/CHANGELOG.md
+++ b/charts/victoria-logs-agent/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## Next release
 
 - added `runtimeClassName` option to the pod spec
+- do not merge .Values.persistentVolume.extraLabels with default chart labels. See [#2460](https://github.com/VictoriaMetrics/helm-charts/issues/2460).
 
 ## 0.2.7
 


charts/victoria-logs-agent/templates/server.yaml
```diff
@@ -117,12 +117,12 @@ spec:
       kind: PersistentVolumeClaim
       metadata:
         name: vl-agent-data
-        {{- $_ := set $ctx "extraLabels" $pvc.extraLabels }}
-        labels: {{ include "vm.labels" $ctx | nindent 10 }}
-        {{- $_ := unset $ctx "extraLabels" }}
         {{- with $pvc.annotations  }}
         annotations: {{ toYaml . | nindent 10 }}
         {{- end }}
+        {{- with $pvc.extraLabels }}
```
vrutkovs, line 123: Wouldn't that ignore `commonLabels` and global `extraLabels`?

vrutkovs:
>Applying vm.labels there included app.kubernetes.io/version and helm.sh/chart, which change on every chart/app version bump and caused helm upgrade

I'm not following - why would a label change cause the helm upgrade to fail?

UPD: nevermind, `The upgrade fails because Kubernetes rejects changes to volumeClaimTemplates metadata.`

AndrewChubatiuk:
hey @samuel-form3 
thanks for a PR
could you please also add a changelog entry for a change you did?

samuel-form3:
Just added it @AndrewChubatiuk 🙏

AndrewChubatiuk (approved):

Issue #3040, opened 2026-06-25
bug: Allow old method of defining rules and dashboard in helm chart/templates


nebula-it:
> [!WARNING]
> **Before creating an issue please make sure it's reproducible on a latest chart version!**

**Chart name and version**
chart: [victoria-metrics-k8s-stack-0.85.0](https://github.com/VictoriaMetrics/helm-charts/releases/tag/victoria-metrics-k8s-stack-0.85.0)
version: 0.85.0

**Describe the bug**
With the new [syncjob](https://github.com/VictoriaMetrics/helm-charts/releases#:~:text=Update%20note%203) to create rules and dashboard, it is hard to maintain the deployment in air-gapped and/or limited network access setups.
Additionally, this goes against the ethos of GitOps, we are using rendered template deployment on argoCD and could clearly see what was going to change with an update. But with new setup its a blackbox where you have to manually track a remote source.

For these reason, can we please keep the original way of how rules and dashboards were populated.

AndrewChubatiuk:
Thanks for the detailed feedback — both concerns are valid.

The old approach was removed because embedding all dashboard and rule files directly in the chart was pushing it close to Kubernetes resource size limits (ConfigMaps/Secrets have a 1MB cap), requiring us to track chart size after every upstream dashboard update. Additionally the sync-job allows users to apply the same datasource and cluster label transformations to their own additional dashboards without having to worry about chart size at all. For these reasons, the old template-based approach will not be brought back.

To address the air-gapped and GitOps use cases we suggest the following solutions:

- Manifest generation mode (upcoming release): set OUTPUT=- (or a file path) to write all rendered ConfigMap/GrafanaDashboard/VMRule manifests as YAML to stdout instead of applying them to the cluster. Run this in CI where network access is available, commit the output to your repository, and deploy it with ArgoCD or any other GitOps tool — including in air-gapped clusters:
```
docker run \
  -e OUTPUT=- \
  -e CONFIG=/path/to/config.yaml \
  -e NAMESPACE=monitoring \
  -e RELEASE=my-release \
  victoriametrics/sync-job > manifests.yaml
```

- Private source URLs: mirror the dashboard and rule sources to a private registry or object storage and point the sync-job sources URLs to your internal endpoints.

nebula-it:
That seems like a reasonable solution. Will give it a shot once released. Thank you for quick fix.

AndrewChubatiuk:
[...]

PR #3041, opened 2026-06-26
sync-job: add manifest generation mode

### LINKED ISSUE #3042: bug: Grafana Variables Broken or Missing (syncJob)
> [!WARNING]
> **Before creating an issue please make sure it's reproducible on a latest chart version!**

**Chart name and version**
chart: victoria-metrics-k8s-stack
version: v0.85.5

**Describe the bug**
With the new syncJob injecting variables in Grafana dashboards and rules, the dashboards no longer work. Before syncJob was implemented, I had set `clusterLabel` to my clusters name in my values file. If I leave that variable set, it will change the variables on Grafana to clusterLabel={$cluster}. I believe the intention here is to be able to filter on the Label `cluster` present in most metrics. I know if I have that variable set in my helm chart, there's a constant that sets my `clusterLabel` to .\*, but having my `clusterLabel` value in a metric query as a label set to equal {$cluster} (`<my_cluster_name>={$cluster})` seems backwards. It seems like it should be cluster={$cluster} in the metric query, and the dashboard variables should be clusterLabel=.* and cluster={$clusterLabel}, that way the labels in the metric query work properly.

I removed the clusterLabel from my chart and made sure `defaultDashboards.syncJob.common.multicluster` is set to `false`. Doing this sets the metric label in the queries back to cluster={$cluster}, but the issue now is $cluster is not set in the variables for the dashboard.

Image

If there is a setting I have missed, please update the documentation to note this. I have compared the chart version I was using before (0.80.0) with the latest value chart (0.85.5), and aside from the new syncJob, the dashboards, and VL and VT sub charts, I can't find anything that suggests I need to either set or unset a variable to make the Grafana dashboards work.

AndrewChubatiuk:
related issue https://github.com/VictoriaMetrics/helm-charts/issues/3040
fixes https://github.com/VictoriaMetrics/helm-charts/issues/3042

### DIFF
--- a/hack/sync-job/main.go
+++ b/hack/sync-job/main.go
@@ -11,38 +11,64 @@ import (
 	"time"
 )
 
+func openOutput(path string) (io.WriteCloser, error) {
+	if path == "-" {
+		return os.Stdout, nil
+	}
+	return os.Create(path)
+}
+
 func main() {
 	configPath := envOrDefault("CONFIG", "/etc/config/config.yaml")
 	namespace := envOrDefault("NAMESPACE", "default")
-	instance := os.Getenv("RELEASE")
+	instance := envOrDefault("RELEASE", "vm-k8s-stack")
 	resourcePrefix := envOrDefault("RESOURCE_PREFIX", instance)
 	prune := envOrDefault("PRUNE", "true") != "false"
 	useOwnerRef := envOrDefault("OWNER_REFERENCES", "true") != "false"
+	outputPath := os.Getenv("OUTPUT")
 
 	cfg, err := loadConfig(configPath)
 	if err != nil {
 		log.Fatalf("load config: %v", err)
 	}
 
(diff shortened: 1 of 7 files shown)


hack/sync-job/main.go
```diff
@@ -11,38 +11,67 @@ import (
...
+	}
+	return os.Create(path)
+}
+
 func main() {
 	configPath := envOrDefault("CONFIG", "/etc/config/config.yaml")
 	namespace := envOrDefault("NAMESPACE", "default")
 	instance := os.Getenv("RELEASE")
 	resourcePrefix := envOrDefault("RESOURCE_PREFIX", instance)
+	if resourcePrefix == "" {
```
vrutkovs: Can we default `instance` to `vm-k8s-stack` instead?

AndrewChubatiuk: done

PR #3054, opened 2026-06-29
Fix annotations for victoria-metrics-operator-crds chart

cosandr:
The values file is misleading, the CRDs are actually a subchart so annotations need to specified as `crds.annotations`.

I have not run `helm-docs` as it results in loads of changes, it doesn't seem like it's been run for a while in this repo.


AndrewChubatiuk (approved):

AndrewChubatiuk:
Thanks for contribution!

AndrewChubatiuk:
we're using helm-docs fork, which is wrapped into Makefile `gen-docs` target

PR #3056, opened 2026-07-01
common: support global labels and annotations

AndrewChubatiuk:


charts/victoria-metrics-common/templates/_helpers.tpl
```diff
@@ -237,15 +237,35 @@ Returns "true", "false", or "" (not set at any level).
...
+  {{- $globalAnnotations := deepCopy (($Values.global).extraAnnotations | default dict) -}}
+  {{- $annotations := mergeOverwrite $globalAnnotations (.extraAnnotations | default dict) -}}
+  {{- if $annotations -}}
+    {{- toYaml $annotations -}}
+  {{- end -}}
+{{- end -}}
+
+{{- define "vm.keyValue" -}}
+  {{- $pairs := list -}}
+  {{- range $k, $v := . -}}
```
vrutkovs, line 263: This needs to be sorted - otherwise you may end up with `foo=1,bar=2` on one run and `bar=2,foo=1` in another

PR #3073, opened 2026-07-04
feat(victoria-metrics-single): add serviceMonitor path parameter

chrxmvtik:
## What changed
Added serviceMonitor path parameter, if .server.extraArgs.http.pathPrefix is set - metrics patch match, otherwise fallback to default /metrics.

## Motivation
Currently if you changed pathPrefix, metrics path also change and scraping will fail - to change metrics path you had to do some relabelings magic to make it work.

This pull request adds path parameter to serviceMonitor and if it is not set at all, it handles all the cases, if user use different http.pathPrefix or not.


vrutkovs (approved):

YuryHrytsuk:
@chrxmvtik thanks a lot for a fix. The same is missing for `vmServiceScrape` :worm:

PR #3090, opened 2026-07-13
victoria-metrics-auth: support passing raw configs

vrutkovs:
Instead of generating config indirectly via values we should support passing the config directly. This would allow advanced users to reuse repeatable blocks and template configuration

### DIFF
--- a/charts/victoria-metrics-auth/templates/secret.yaml
+++ b/charts/victoria-metrics-auth/templates/secret.yaml
@@ -13,5 +13,5 @@ metadata:
   labels: {{ include "vm.labels" $ctx | nindent 4 }}
 type: Opaque
 data:
-  auth.yml: |{{ toYaml .Values.config | b64enc | nindent 4 }}
+  auth.yml: |{{- if .Values.configRaw }}{{ tpl .Values.configRaw . | b64enc | nindent 4 }}{{- else }}{{ toYaml .Values.config | b64enc | nindent 4 }}{{- end }}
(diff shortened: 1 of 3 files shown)


charts/victoria-metrics-auth/templates/secret.yaml
```diff
@@ -13,5 +13,5 @@ metadata:
   labels: {{ include "vm.labels" $ctx | nindent 4 }}
 type: Opaque
 data:
-  auth.yml: |{{ toYaml .Values.config | b64enc | nindent 4 }}
+  auth.yml: |{{ tpl (.Values.configRaw | default (toYaml .Values.config)) . | b64enc | nindent 4 }}
```
AndrewChubatiuk: don't want to add tpl due to [jwt templating](https://docs.victoriametrics.com/vmauth/#jwt-claim-based-request-templating)

vrutkovs: Now `tpl` is used only when configRaw is set

AndrewChubatiuk: do you expect users to escape jwt templates?

vrutkovs: Yes, if they want to have both jwt and config raw they should be able to contruct a configuration themselves. Imo it can easily be tested with `helm template` anyway

AndrewChubatiuk (approved):

PR #3108, opened 2026-07-20
victoria-metrics-mcp: fix namespaceSelector to actually target the namespace

sapslaj:
Currently it is set to target the release name, which is incorrect and not what victoria-logs-mcp does. This fixes it to actually use the `$ns` variable defined.


AndrewChubatiuk (approved):

AndrewChubatiuk:
Thanks a lot for contribution!

PR #3130, opened 2026-07-31
fix(victoria-metrics-operator): grant networkpolicies in ClusterRole

### LINKED ISSUE #3129: bug: victoria-metrics-operator chart 0.67.0 ClusterRole lacks networkpolicies RBAC — operator silently stops reconciling everything
**Chart name and version**

```
chart: victoria-metrics-operator
version: 0.67.0        # appVersion v0.74.0 — latest published, and master is affected too
last good: 0.66.3      # appVersion v0.73.1
```

Kubernetes: K3s v1.36.2+k3s1.

**Custom values**

None relevant — reproduces on defaults. For completeness, mine are unrelated to RBAC:

```yaml
operator:
  disable_prometheus_converter: true
  enable_converter_ownership: false
```

`rbac.create` is left at its default (`true`).

**Describe the bug**

**Regression in chart 0.67.0.** The `victoria-metrics-operator` chart's ClusterRole (`templates/role.yaml`) is missing the `networkpolicies` RBAC rule that operator v0.74.0 requires. Chart 0.66.3 (operator v0.73.1) was unaffected.

The chart grants:

```yaml
- apiGroups:
  - networking.k8s.io
  resources:
  - ingresses
  - ingresses/finalizers
  verbs:
  - "*"
```

The operator repo's own `config/rbac/role.yaml` at `v0.74.0` grants:

```yaml
- apiGroups:
  - networking.k8s.io
  resources:
  - ingresses
  - ingresses/finalizers
  - networkpolicies      # <-- absent from the chart
  verbs:
  - "*"
```

The v0.74.0 release notes ship that grant as an explicit BUGFIX:

> BUGFIX: vmoperator: grant the operator's `ClusterRole` permissions on `networking.k8s.io/networkpolicies`. The `networkPolicy` feature above was shipped without this grant, so the operator would fail with a `Forbidden` error creating/updating/deleting `NetworkPolicy` objects on a real cluster.

That fix landed in the operator repo but was never mirrored into the chart.

**The impact is much wider than the release note implies**

The release note frames this as breaking `NetworkPolicy` create/update/delete, i.e. affecting only users who adopt the new field. It is worse than that: **a chart-installed operator stops reconciling everything, whether or not any CR sets `.spec.networkPolicy`.**

The `NetworkPolicy` read is not gated on the feature. Every factory's cleanup path takes the `== nil` branch unconditionally — `vmsingle.go:730`, `vmagent.go:1291`, `vmalert.go:844`, `vmauth.go:687`, `vmalertmanager/alertmanager.go:146`, `vmanomaly/statefulset.go:223`, `vlagent.go:782`, `vlsingle.go:394`, `vtsingle.go:380`:

```go
if cr.Spec.NetworkPolicy == nil {
    objsToRemove = append(objsToRemove, &networkingv1.NetworkPolicy{ObjectMeta: objMeta})
}
...
return finalize.SafeDeleteWithFinalizer(ctx, rclient, objsToRemove, cr)
```

`SafeDeleteWithFinalizer` starts with `rclient.Get`. That is a cached read, so controller-runtime lazily starts a `NetworkPolicy` informer — which can never sync without `list`/`watch`. The reconcile parks there.

**What that looks like in practice:** the controllers all start normally, then quietly stop making progress. On my K3s v1.36.2 cluster, all 25 controllers logged `Starting workers` at 10:31:35Z, and then:

| Signal | Value during the outage |
|---|---|
| Pod status | `Running`, `Ready 1/1`, `restartCount: 0` |
| `up{job="victoria-metrics-operator"}` | `1` — the metrics endpoint is fine |
| `:8081` liveness/readiness | green |
| Leader lease | renewing normally |
| `controller_runtime_reconcile_total` | **flat at 3 for 124 minutes** |
| `controller_runtime_reconcile_errors_total` | `0` |
| Operator logs | 393 × `Failed to watch` for `*v1.NetworkPolicy`, and nothing else |

Every VM CR change was silently ignored for those 124 minutes after an automated chart bump. Nothing in the pod's health surface reflected it. The only thing that caught it was an alert on `sum(rate(controller_runtime_reconcile_total{job="victoria-metrics-operator"}[15m])) == 0`.

Note that `reconcile_errors_total` stays at `0` — the reconciles never fail, they never *return*. Any alerting keyed on error rate will not see this.

**To Reproduce**

1. Install chart `victoria-metrics-operator` 0.67.0 with the default `rbac.create: true`.
2. Apply any VM CR (e.g. a `VMSingle`) — no `.spec.networkPolicy` needed.
3. Operator logs, repeating every few seconds:

```json
{"level":"error","logger":"controller-runtime.cache.UnhandledError","msg":"Failed to watch",
 "reflector":"pkg/mod/k8s.io/client-go@v0.36.3/tools/cache/reflector.go:343",
 "type":"*v1.NetworkPolicy",
 "error":"failed to list *v1.NetworkPolicy: networkpolicies.networking.k8s.io is forbidden: User \"system:serviceaccount:monitoring:victoria-metrics-operator\" cannot list resource \"networkpolicies\" in API group \"networking.k8s.io\" at the cluster scope"}
```

4. Confirm the missing grant:

```console
$ kubectl get clusterrole victoria-metrics-operator -o json | jq -c '.rules[] | select(.apiGroups[]? == "networking.k8s.io")'
{"apiGroups":["networking.k8s.io"],"resources":["ingresses","ingresses/finalizers"],"verbs":["*"]}
```

5. Watch `controller_runtime_reconcile_total` go flat while the pod stays `Ready 1/1`.

Chart 0.67.0 (released 30 Jul 2026) is the newest published version, and `master` still lacks the rule as of 2026-07-31 — so this reproduces on latest.

**Suggested fix**

Add `networkpolicies` to the existing `networking.k8s.io` block in `charts/victoria-metrics-operator/templates/role.yaml`, matching the operator repo's `config/rbac/role.yaml`. PR: #3130.

It should be **unconditional**, not gated behind a values flag like `gateway_support`/`vpa_support`. The informer starts on every reconcile regardless of whether the feature is used, so a flag-gated grant would leave the default install broken in exactly the way described above.

This is the same class as [#3102](https://github.com/VictoriaMetrics/helm-charts/issues/3102) (chart ClusterRole missing the `verticalpodautoscalers` grant the operator's own role.yaml carries), fixed in chart 0.66.3. It may be worth a CI check that diffs the chart's `role.yaml` against the operator repo's `config/rbac/role.yaml` for the pinned appVersion, so the next one is caught before release rather than in production.

**Workaround for anyone hitting this now**

A supplementary ClusterRole + ClusterRoleBinding is enough to restore reconciliation. No pod restart is needed — RBAC is picked up live, and the parked informer syncs on its next retry:

```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: victoria-metrics-operator-networkpolicy-reader
rules:
  - apiGroups: ["networking.k8s.io"]
    resources: ["networkpolicies"]
    verbs: ["get", "list", "watch"]
---
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRoleBinding
metadata:
  name: victoria-metrics-operator-networkpolicy-reader
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: victoria-metrics-operator-networkpolicy-reader
subjects:
  - kind: ServiceAccount
    name: victoria-metrics-operator
    namespace: monitoring   # adjust to your release namespace
```

Read-only verbs suffice **only** if no CR sets `.spec.networkPolicy` — `SafeDeleteWithFinalizer` returns early on `NotFound`, so the cleanup path never attempts a write. Anyone actually using the feature needs the full `["*"]`; there the operator surfaces the `Forbidden` as a reconcile error rather than skipping silently.

AKhozya:
Fixes #3129.

Adds `networkpolicies` to the operator ClusterRole's existing `networking.k8s.io` rule, matching the operator repo's own `config/rbac/role.yaml` at v0.74.0.

Operator v0.74.0 grants itself this permission and the release notes ship it as a BUGFIX, but the chart was never updated, so a chart-installed 0.67.0 operator cannot list or watch NetworkPolicy.

**The breakage is not limited to users of the new `.spec.networkPolicy` field.** Every factory's cleanup path takes the `== nil` branch unconditionally (`vmsingle.go:730`, `vmagent.go:1291`, `vmalert.go:844`, `vmauth.go:687`, `vmalertmanager/alertmanager.go:146`, `vmanomaly/statefulset.go:223`, `vlagent.go:782`, `vlsingle.go:394`, `vtsingle.go:380`) and hands a NetworkPolicy to `finalize.SafeDeleteWithFinalizer`, whose opening `rclient.Get` makes controller-runtime start a NetworkPolicy informer. Without `list`/`watch` that informer never syncs and reconciles park on it — the operator stops reconciling everything while its pod stays `Ready 1/1` with `up=1`, `restartCount 0` and a renewing leader lease. `controller_runtime_reconcile_total` goes flat and `reconcile_errors_total` stays at `0`. See #3129 for the full trace.

Because the informer starts regardless of whether the feature is used, the grant is deliberately **unconditional** rather than gated behind a values flag like `gateway_support`/`vpa_support` — a flag-gated grant would leave the default install broken.

**Verification**

- `helm lint` passes.
- Default values render the rule as `[ingresses, ingresses/finalizers, networkpolicies]` with verbs `["*"]`.
- `--set operator.gateway_support=true --set operator.vpa_support=true` renders unchanged apart from the new resource — the `gateway.networking.k8s.io` and `autoscaling.k8s.io` blocks are untouched.
- `rbac.create=false` behaviour is unchanged.
- Live-confirmed on K3s v1.36.2: granting the operator's ServiceAccount `networkpolicies` restores reconciliation immediately, with no pod restart.

Only `templates/role.yaml` and the chart `CHANGELOG.md` change; `values.yaml` is untouched, so `helm-docs` produces no README diff.

### DIFF
--- a/charts/victoria-metrics-operator/CHANGELOG.md
+++ b/charts/victoria-metrics-operator/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next release
 
-- TODO
+- added `networkpolicies` permission to the operator ClusterRole. Without it the operator cannot sync its NetworkPolicy informer and stops reconciling every CR, even when no CR sets `.spec.networkPolicy`. See [#3129](https://github.com/VictoriaMetrics/helm-charts/issues/3129).
 
 ## v0.67.0
 
--- a/charts/victoria-metrics-operator/templates/role.yaml
+++ b/charts/victoria-metrics-operator/templates/role.yaml
@@ -158,6 +158,7 @@ rules:
   resources:
   - ingresses
   - ingresses/finalizers
+  - networkpolicies
   verbs:
   - "*"
 {{- if .Values.operator.gateway_support }}


AndrewChubatiuk (approved):

AndrewChubatiuk:
thanks for contribution!

PR #3139, opened 2026-08-04
feat: configure renovate to allow minor/patch updates

vrutkovs:
Create PRs to bump chart dependencies (excluding major updates for now)

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -44,6 +44,16 @@
       "enabled": false
     },
     {
+      "description": "Disable updates for local CRD chart",
+      "matchManagers": ["helmv3"],
+      "matchPackageNames": ["crds"],
+      "enabled": false
+    },
+    {
+      "matchManagers": ["helmv3"],
+      "rangeStrategy": "replace"
+    },
+    {
       "description": "Disable updates for victoriametrics images tracked via appVersion",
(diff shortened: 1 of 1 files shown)


renovate.json
```diff
@@ -43,6 +43,18 @@
...
+    {
+      "description": "Disable updates for local CRD chart",
+      "matchManagers": ["helmv3"],
+      "matchPackageNames": ["crds"],
+      "enabled": false
+    },
+    {
+      "description": "Allow minor and patch Helm dependency updates",
+      "matchManagers": ["helmv3"],
+      "matchUpdateTypes": ["major"],
```
AndrewChubatiuk: why do you want to exclude major version?

vrutkovs: yes - I'm worried that kube-state-metrics or prometheus-node-exporter major bumps could break us

AndrewChubatiuk: but dependencies update is not automatic, we see in PR versions and changelog. if we just disable it we need to check manually the latest version

vrutkovs: okay, lets try that - I'm not sure if we can interpret breaking changes correctly - but hopefully e2e tests will help us catch these issues

AndrewChubatiuk (approved):

Issue #3102, opened 2026-07-17
victoria-metrics-operator: ClusterRole lacks verticalpodautoscalers RBAC required by the operator's embedded-VPA feature


voron:
### Describe the bug

The `victoria-metrics-operator` chart's ClusterRole (`templates/role.yaml`) is missing the RBAC rule for the operator's embedded-VPA feature (`spec.vpa` on VMCluster/VMAgent/VMSingle/etc., available since operator v0.69.0/v0.72.0). The operator repo's own `config/rbac/role.yaml` grants it unconditionally:

```yaml
- apiGroups:
  - autoscaling.k8s.io
  verbs:
  - "*"
  resources:
  - verticalpodautoscalers
```

but the chart's `role.yaml` only carries the `autoscaling / horizontalpodautoscalers` rule. As a result, a chart-installed operator (checked with chart 0.66.2 / operator v0.73.1) cannot create the `VerticalPodAutoscaler` objects that `spec.vpa` + `VM_VPA_API_ENABLED=true` ask it to manage — CR reconciliation fails on RBAC.

### To Reproduce

1. Install `victoria-metrics-operator` chart 0.66.2 with `env: [{name: VM_VPA_API_ENABLED, value: "true"}]`.
2. Apply a `VMCluster` with `spec.vmstorage.vpa` set (valid `updatePolicy` + `resourcePolicy`).
3. Operator reconcile fails: forbidden creating `verticalpodautoscalers.autoscaling.k8s.io`.

### Expected behavior

Chart RBAC matches the operator's declared requirements (`config/rbac/role.yaml`), as it already does for HPA. An unconditional rule is harmless on clusters without the VPA CRDs — RBAC referencing absent resource types is legal.

Nice-to-have on top: a first-class values toggle (e.g. `operator.vpa.enabled` or similar) that wires **both** the `VM_VPA_API_ENABLED=true` env var and the RBAC rule, so users of the documented `spec.vpa` CRD field don't have to discover two separate undocumented knobs (we found both only via reconcile error messages and a role diff).

### Workaround

```yaml
extraObjects:
  - apiVersion: rbac.authorization.k8s.io/v1
    kind: ClusterRole
    metadata:
      name: victoria-metrics-operator-vpa
    rules:
      - apiGroups: ["autoscaling.k8s.io"]
        resources: ["verticalpodautoscalers"]
        verbs: ["*"]
  - apiVersion: rbac.authorization.k8s.io/v1
[...]

PR #3207, opened 2026-09-11
feat: derive operator ClusterRole rules from operator-role.yaml

AquaOctet:
## Changes:
- pulls the operator's non-CRD ClusterRole rules from `operator-role.yaml` instead of hand-maintaining a copy of them in `templates/role.yaml`
- Adds `operator-role.yaml`, pulled from operator `v0.75.0-rc1`
- Adds `tests/role_test.yaml`: 6 snapshots across feature-flag, cleanup-hook, aggregated-role and watchNamespaces permutations, plus 5 capability invariants

## Description:

Follow-up to VictoriaMetrics/operator#2602, which syncs `config/rbac/role.yaml` into this chart on release. This is the PR that consumes that file.

The chart's copy drifts from the operator's  #3129 (`networkpolicies`) and
#3102 (`verticalpodautoscalers`). The chart already derives CRD permissions from the
bundled `crd.yaml`

Tested on `kind` with operator v0.74.1 and v0.75.0-rc1: VMSingle, VMAgent, VMAlert and
VMAuth all reconcile, no RBAC denials on either.

### DIFF
--- a/charts/victoria-metrics-operator/operator-role.yaml
+++ b/charts/victoria-metrics-operator/operator-role.yaml
@@ -1,0 +1,326 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: operator
+rules:
+- nonResourceURLs:
+  - /metrics
+  - /metrics/resources
+  - /metrics/slis
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  - configmaps/finalizers
+  - endpoints
+  - persistentvolumeclaims
+  - persistentvolumeclaims/finalizers
+  - secrets
+  - secrets/finalizers
+  - serviceaccounts/finalizers
+  - services
+  - services/finalizers
+  verbs:
+  - '*'
+- apiGroups:
+  - ""
+  resources:
+  - events
+  verbs:
+  - '*'
+- apiGroups:
+  - ""
+  resources:
+  - namespaces
+  - nodes
+  - nodes/metrics
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods
+  verbs:
+  - delete
+  - get
+  - list
+  - patch
+  - watch
+- apiGroups:
+  - ""
+  resources:
+  - pods/eviction
+  verbs:
+  - create
+- apiGroups:
+  - ""
+  resources:
+  - serviceaccounts
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - apps
+  resources:
+  - daemonsets
+  - daemonsets/finalizers
+  - deployments
+  - deployments/finalizers
+  - replicasets
+  - statefulsets
+  - statefulsets/finalizers
+  verbs:
+  - '*'
+- apiGroups:
+  - apps
+  resources:
+  - statefulsets/status
+  verbs:
+  - get
+  - patch
+  - update
+- apiGroups:
+  - autoscaling
+  resources:
+  - horizontalpodautoscalers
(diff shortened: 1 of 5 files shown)


AquaOctet:
Tested on `kind` with operator `v0.74.1` and `v0.75.0-rc1`: VMSingle, VMAgent, VMAlert and
VMAuth all reconcile, no RBAC denials on either.

I allowed for backwards compatibility since `escalate` is legacy, originally tested removing it but didn't want to break things for those who update the operator and not every component.

charts/victoria-metrics-operator/templates/role.yaml
```diff
@@ -64,135 +64,47 @@ rules:
...
-  - gateway.networking.k8s.io
-  resources:
-  - httproutes
-  - httproutes/finalizers
-  verbs:
-  - '*'
+{{- /* do not add permissions here, please add a +kubebuilder:rbac marker in the operator repo */ -}}
+{{- /* these rules come from /victoria-metrics-operator/operator-role.yaml, synced on operator release */}}
+{{- $gated := dict
+      "autoscaling.k8s.io/verticalpodautoscalers"        .Values.operator.vpa_support
```
AndrewChubatiuk: why only group name is not enough?

AquaOctet: Thank you for the sanity check. 

I wanted it to fail loudly, but I didn't consider the edge case.

charts/victoria-metrics-operator/operator-role.yaml
```diff
@@ -0,0 +1,326 @@
...
+  - registry/metrics
+  - routers/metrics
+  verbs:
+  - get
+- apiGroups:
+  - monitoring.coreos.com
+  resources:
+  - '*'
+  verbs:
+  - '*'
```
AndrewChubatiuk, line 139: it makes two items below useless

charts/victoria-metrics-operator/operator-role.yaml
```diff
@@ -0,0 +1,326 @@
...
+  resources:
+  - ingresses
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
AndrewChubatiuk, line 177: what is a reason of splitting networkpolicies and ingresses?

charts/victoria-metrics-operator/operator-role.yaml
AquaOctet: It is redundant, `vmagent_controller.go:78` and `vmsingle_controller.go:84` declare `resources=*`,`verbs=*` for monitoring.coreos.com . 

I would say we shouldn't edit this here (helm-charts) though, as this will be copied by the sync. Any change would be reverted on the next operator release. I can place an issue in the Operator repo to track this though, and add it to the backlog.

charts/victoria-metrics-operator/operator-role.yaml
AquaOctet: `controller-gen` groups rules by verb set, resources only share a rule when their verbs match.

- ingresses: get, list, watch, create, update, patch, delete  : `vmauth_controller.go:72` with `vmagent_controller.go:72` and `vmsingle_controller.go:78`
- networkpolicies : * : 13 controllers
- ingresses/finalizers : * : `finalize/common.go:19`

The grouping comes from markers, not from the chart.

charts/victoria-metrics-operator/operator-role.yaml
AquaOctet: Created :  https://github.com/VictoriaMetrics/operator/issues/2628

AquaOctet:
@AndrewChubatiuk, please let me know what you think  :) (not sure if it notifies unless new commit).

AndrewChubatiuk:
created PR in operator repo
https://github.com/VictoriaMetrics/operator/pull/2629

PR #3211, opened 2026-09-14
charts: update vm apps to v1.152.0

makasim:


AndrewChubatiuk:
@makasim 
could you please also update `.Values.global.versions.metrics` at charts/victoria-metrics-k8s-stack/values.yaml?

makasim:
> @makasim
could you please also update .Values.global.versions.metrics at charts/victoria-metrics-k8s-stack/values.yaml?

I'll do it manually, but I suppose it should be automated as well, right?

AndrewChubatiuk:
also would like to automate it, but for now have no ideas besides dependabot

makasim:
I'll take care of it

makasim:
The values file has been updated. There are unrelated changed made by `yq`. I think we should preserve them so next runs produce no diff. 

@AndrewChubatiuk ptal

AndrewChubatiuk (approved):

PR #3215, opened 2026-09-16
fix(chart: victoria-logs-single): Codeblock not escaped correctly

userbradley:
Currently there is an issue with the codeblock on the `victoria-logs-single` page where the backticks are imediatly followed by `.` which causes the codeblock not to close.

image

Running `make gen-docs` locally, this seems to have solved the issue

image


AndrewChubatiuk (approved):

AndrewChubatiuk:
LGTM! Thanks for a fix

PR #3218, opened 2026-09-18
feat(k8s-stack): allow imagePullSecrets on the sync-job Pod

obervinov:
## Problem

`templates/sync-job/job.yaml` renders a Pod spec with no `imagePullSecrets`, and it does not read `global.imagePullSecrets` either. `templates/sync-job/rbac.yaml` creates the job's ServiceAccount with no way to attach them, and the ServiceAccount name is not overridable.

The practical effect is that `syncJob.image.repository` cannot be pointed at a private registry or a pull-through mirror. Doing so leaves the Pod in `ImagePullBackOff`:

```
failed to authorize: failed to fetch oauth token: unexpected status from GET request to
https://<registry>/v2/token?scope=repository%3A<path>%2Fsync-job%3Apull: 403 Forbidden
```

Setting `syncJob.imagePullSecrets` in values is silently ignored, since no template consumes it. Because the Job is a `post-install,post-upgrade` hook and `ImagePullBackOff` never converts into a Job failure, the Job stays `active=1` and the hook never completes, which blocks the release from finishing.

## Change

Adds `syncJob.imagePullSecrets`, falling back to `global.imagePullSecrets`, using the same idiom already used elsewhere in this repository (for example `charts/victoria-metrics-single/templates/server.yaml`):

```
{{- with ($job.imagePullSecrets | default .Values.global.imagePullSecrets) }}
imagePullSecrets: {{ toYaml . | nindent 8 }}
{{- end }}
```

Default is `[]`, so rendering is unchanged for anyone not setting it.

## Verification

`helm template -s templates/sync-job/job.yaml`, three cases:

| values | rendered |
|---|---|
| `syncJob.imagePullSecrets[0].name=my-secret` | `imagePullSecrets: [{name: my-secret}]` |
| `global.imagePullSecrets[0].name=global-secret` | `imagePullSecrets: [{name: global-secret}]` |
| neither set | no `imagePullSecrets` key |

`helm lint` passes.


AndrewChubatiuk (approved):
LGTM! Thanks you!

